PR #1, opened 2022-09-11
use nodes; clean up code (.net-style)

gregsdennis:
Not sure if you have a build on this project or not.  The running instructions are in unix/mac, so I didn't run it locally, but I know it compiles.


gregsdennis:
looks like I need to add a .gitignore for this folder.

Julian:
Nice, thanks!

> Not sure if you have a build on this project or not.

(There's a build, though it's got a flaky test in CI which I haven't spent time fixing yet as it'll get better when I change how output is collected and it passes locally).

>  The running instructions are in unix/mac

Yeah, Windows docs also obviously are missing... What I get locally from running this with some (bowtie-less) test input is:

```
⊙  podman run --rm -i bowtie/dotnet-json-everything <<EOF                                                                                                                         julian@Airm ●
{"cmd": "start", "version": 1}
{"cmd": "run", "seq": 0, "case": {"description": "stuff", "schema": {"const": {"foo": "bar"}}, "tests": [{"description": "a test", "instance": {"foo": "bar"}, "valid": true}, {"description": "another test", "instance": {}, "valid": false}]}}
{"cmd": "stop"}
EOF
{
  "ready": true,
  "version": 1,
  "implementation": {
    "language": "dotnet",
    "name": "json-everything",
    "homepage": "https://json-everything.net/json-schema/",
    "issues": "https://github.com/gregsdennis/json-everything/issues"
  }
}
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Program.<Main>$(String[] args) in /source/Program.cs:line 51
```

Any chance you see what that comes from?

gregsdennis:
I'll have a look

Julian:
Found it, think it's all good now thanks for the PR!

implementations/dotnet-json-everything/Program.cs
```diff
@@ -58,7 +58,7 @@
 
 			var runResult = new JsonObject
 			{
-				["seq"] = root["seq"],
+				["seq"] = root["seq"].GetValue<int>(),
```
gregsdennis, line 61: This just gets cast back to a `JsonNode` anyway.  There's no benefit to extracting the `int` except for the side effect of checking that it is one.

implementations/dotnet-json-everything/Program.cs
```diff
@@ -41,7 +41,7 @@
 			}
 
 			var testCase = root["case"];
-			var schemaText = root["schema"];
+			var schemaText = testCase["schema"];
```
gregsdennis, line 44: Ha. oops.

implementations/dotnet-json-everything/Program.cs
Julian: Without it I got another error. One that said:

```
Unhandled exception. System.InvalidOperationException: The node already has a parent.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_NodeAlreadyHasParent()
   at System.Text.Json.Nodes.JsonNode.AssignParent(JsonNode parent)
   at System.Text.Json.Nodes.JsonObject.<>c__DisplayClass9_0.<SetItem>b__0()
   at System.Text.Json.JsonPropertyDictionary`1.SetValue(String propertyName, T value, Action assignParent)
   at System.Text.Json.Nodes.JsonObject.SetItem(String propertyName, JsonNode value)
   at Program.<Main>$(String[] args) in /source/Program.cs:line 59
```

Is there another way that should have been handled? (Google suggested a yet sillier one to serialize back to a string and deserialize again)

gregsdennis: Hmm yeah, I understand that, but I wouldn't have expected it.

Nodes hold a reference to the parent, which means they can only exist in a single tree.  Not sure how else to handle that.

Issue #57, opened 2022-11-09
registry is empty for some tests where it is expected


santhosh-tekuri:
in `draft2020-12/dynamicRef.json` for following cases, registry is passed as empty:
- `strict-tree schema, guards against misspelled properties`
- `tests for implementation dynamic anchor and reference link`
- `$ref and $dynamicAnchor are independent of order - $defs first`
- `$ref and $dynamicAnchor are independent of order - $ref first`

PR #59, opened 2022-11-09
Populate registry for dynamicRef.json

santhosh-tekuri:
fixes issue #57


Julian:
Thanks! You certainly seem correct, I hadn't realized we'd let some of these leak out into another file. That's a bit unfortunate (upstream for the suite), I'd tried to keep all the tests that needed this in one place, but seems we now indeed have two, oh well, not the worst thing.

Much appreciated! Along with any other feedback or help you can offer.

PR #63, opened 2022-11-09
Populate registry for vocabulary.json

santhosh-tekuri:
reqd http://localhost:1234/draft2020-12/metaschema-no-validation.json

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -471,7 +471,7 @@ def sequenced(cases, reporter):
 
 def suite_cases_from(files, remotes):
     for file in files:
-        if file.name == "refRemote.json" or file.name == "dynamicRef.json":
+        if file.name in {"refRemote.json", "dynamicRef.json", "vocabulary.json"}:
             registry = {
                 urljoin(
                     "http://localhost:1234",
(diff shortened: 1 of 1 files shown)


bowtie/_cli.py
```diff
@@ -471,7 +471,7 @@ def sequenced(cases, reporter):
 
 def suite_cases_from(files, remotes):
     for file in files:
-        if file.name == "refRemote.json" or file.name == "dynamicRef.json":
+        if file.name == "refRemote.json" or file.name == "dynamicRef.json" or file.name == "vocabulary.json":
```
Julian, line 474: :/ -- it now probably makes sense to do this as:

```suggestion
        if file.stem in {"refRemote", "dynamicRef", "vocabulary"}:
```

santhosh-tekuri: I guess it should be array literal not map literal

Julian: That's a set literal (with no colons).

Julian: (CI is failing because the line is now too long -- `.stem` will trim off the `.json`, which probably fits).

PR #65, opened 2022-11-09
Refactoring go-jsonschema impl

santhosh-tekuri:
- fix: Compiler.Draft is lost on next iteration
- use Compiler.LoadURL instead of jsonschema.LoadURL
- extract jsonschemaVersion, jsonReader, printJSON

### DIFF
--- a/implementations/go-jsonschema/bowtie_jsonschema.go
+++ b/implementations/go-jsonschema/bowtie_jsonschema.go
@@ -2,14 +2,16 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/santhosh-tekuri/jsonschema/v5"
 	"io"
 	"log"
 	"os"
 	"runtime/debug"
 	"strings"
+
+	"github.com/santhosh-tekuri/jsonschema/v5"
 )
 
 var drafts = map[string]*jsonschema.Draft{
@@ -21,19 +23,21 @@ var drafts = map[string]*jsonschema.Draft{
 }
 
 func main() {
-
 	var started = false
+	var draft *jsonschema.Draft = nil
 
 	scanner := bufio.NewScanner(os.Stdin)
 	for scanner.Scan() {
 		var request map[string]interface{}
-		err := json.Unmarshal([]byte(scanner.Text()), &request)
+
+		decoder := json.NewDecoder(strings.NewReader(scanner.Text()))
+		decoder.UseNumber()
+		err := decoder.Decode(&request)
 		if err != nil {
 			fmt.Printf("could not unmarshal json: %s\n", err)
 			return
 		}
-		encoder := json.NewEncoder(os.Stdout)
-		compiler := jsonschema.NewCompiler()
+
 		switch request["cmd"] {
 		case "start":
 			started = true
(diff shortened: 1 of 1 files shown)


implementations/go-jsonschema/bowtie_jsonschema.go
```diff
@@ -166,3 +136,34 @@ func main() {
...
+		if strings.Contains(dep.Path, "github.com/santhosh-tekuri/jsonschema") {
+			version = dep.Version
+			break
+		}
+	}
+	return version
+}
+
+func jsonReader(v interface{}) io.ReadCloser {
+	b, err := json.Marshal(v)
```
Julian, line 157: Interesting! Your implementation indeed doesn't have a way to pass in a thing that has already been deserialized? I thought I must have just been missing it.

implementations/go-jsonschema/bowtie_jsonschema.go
```diff
@@ -45,30 +46,15 @@ func main() {
...
-				if strings.Contains(dep.Path, "github.com/santhosh-tekuri/jsonschema") {
-					jsonschemaVersion = dep.Version
-					break
-				}
-			}
-
 			data := map[string]interface{}{
 				"ready":   true,
-				"version": 1,
+				"version": version,
```
Julian, line 51: I want this explicitly hardcoded, the point is that a harness shouldn't just blindly copy the version that comes in, it has to be sure it's indeed speaking the right protocol (because e.g. someone will come in here and change support to v2 if the protocol changes, and then they update this from 1 to 2). But if bowtie starts speaking v2 and no changes happen here, the harness shouldn't suddenly claim it's speaking v2, it probably isn't.

Julian:
Awesome. One minor change (to leave as-is), but the rest seems great.

implementations/go-jsonschema/bowtie_jsonschema.go
santhosh-tekuri: such cases are rare so I did not bother implementing it

implementations/go-jsonschema/bowtie_jsonschema.go
santhosh-tekuri: we already panic if version!=1. also when updating to version 2 in future we don't worry about changing version in `if condition` and also in `json response`

implementations/go-jsonschema/bowtie_jsonschema.go
santhosh-tekuri: also to support big numbers I unmarshal schema with `json.UseNumber()`

implementations/go-jsonschema/bowtie_jsonschema.go
Julian: Perhaps a middle ground then is to just have a global `VERSION = 1` which is defined in one place and then use that both for `panic`'ing and for the JSON response? That will save updating many places. This isn't major, but again, the issue I have is partially that it's not correct that the version here is the one from the request, which is what reusing the variable would indicate semantically -- the version here is the one that the harness is written for, which yes if the code is all correct and we paniced will always match the request, but it's asking for some subtle issue maybe not in this harness but in some future one that forgets to panic.

implementations/go-jsonschema/bowtie_jsonschema.go
Julian: Got it! Makes sense (certainly on the latter) -- which just to confirm, does that mean that the roundtrip here was lossy (because we already have deserialized the input without calling `.UseNumber()`?

santhosh-tekuri: might be lossy in case of exact float representation. seems some optional tests with multipleOf might fail if floating precision is lost. let me fix this also...

Julian: Cool (those aren't running yet, doing so is #36, which is why they may not appear yet as failures).

santhosh-tekuri: done. but I noticed in website there are 3 test errors but when locally ran it has 14 errors.

not so easy to find which are failing

Julian: Help welcome! As I say, it should be simple to improve the report for filtering and sorting, it's what's being worked on now.

Julian:
Thanks! Guessing this is ready to go now, looks good.

implementations/go-jsonschema/bowtie_jsonschema.go
santhosh-tekuri: can you run `Test Suite Report` action once for me

Julian: Done! Should be up to date, though I'm making more minor tweaks to the report on the way to the above so I'll be rebuilding it again in ~5 mins.

I'm also happy to add you as a contributor if you'd like (and then you can run it as often as you'd like yourself as well)

Julian: Hooray! Awesome, glad to hear.

PR #66, opened 2022-11-09
update lib name for json-everything

gregsdennis:
`json-everything` is the project name, but `JsonSchema.Net` is the library name.  There are other libraries for other JSON functionality in the project.


Julian:
Ah, got it! My mistake. Can you rename the directory as well then (to `dotnet-jsonschema-net` I guess)?

Also, I don't suppose you know how to get the version of a dependency (i.e. your lib) at runtime from .NET while you're here?

Julian:
It goes in the `implementation.version` field in the response to `start` commands -- here's an example from a [JS](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/js-ajv/bowtie_ajv.js#L39) [Golang](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/go-jsonschema/bowtie_jsonschema.go#L58) or [Python](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/python-jschon/bowtie_jschon.py#L62) implementation if it helps.

Julian:
The CI failure is because CI is running `clang-format`, rather than `dotnet format` to decide how to format `.cs` files -- I want / wanted to run `dotnet format`, but it's not runnable from `pre-commit` (because it requires installing .NET dependencies to run which yeah aren't available).

I obviously have no style preference for .NET though I would like for *some* autoformatter to be run -- do you know if it's possible for `clang-format` to be configured such that it emits what `dotnet format` emits basically?

gregsdennis:
> I want / wanted to run dotnet format, but it's not runnable from pre-commit (because it requires installing .NET dependencies to run which yeah aren't available).

I can make the same claim for `clang`: I don't have the runtime to support it.

I actually have `dotnet format` running as a pre-commit hook in my repo.

```sh
#!/bin/sh

LC_ALL=C
# Select files to format
FILES=$(git diff --cached --name-only --diff-filter=ACM "*.cs" | sed 's| |\\ |g')
[ -z "$FILES" ] && exit 0

# Format all selected files
echo "$FILES" | cat | xargs | sed -e 's/ /,/g' | xargs dotnet format json-everything.sln --include

# Add back the modified files to staging
echo "$FILES" | xargs git add

exit 0
```

It runs `dotnet format` when any _.cs_ file is changed.  Yeah, it does require that .Net is installed.

gregsdennis:
I'll revert that commit, but it's not typical C# formatting, regardless of what clang says.

Julian:
Yeah, I'm talking specifically about [pre-commit.ci](https://pre-commit.ci/) where you can't install any additional deps (other than ones pre-commit can install itself) -- if you don't have clang I'll run it here (or pre-commit.ci should be adding a commit fixing it but I guess I didn't actually enable the main reason to be using it...). Anyways don't worry about it (other than if you have suggestions for what to tweak), I'll merge.

Julian:
I think that last commit does it now, right?

gregsdennis:
yeah, that's pretty close.  thanks.

PR #69, opened 2022-11-11
Local debug support for JsonSchema.Net

gregsdennis:


gregsdennis:
I'm going to bump the version of my lib as well.  Just waiting until I publish.  Shouldn't be long.

gregsdennis:
```
-}
\ No newline at end of file
+}
```

My editor (VSCode) is doing this, I think.

Julian:
It wants a single newline at the end and is complaining there isn't one

Julian:
(I'll have to look tomorrow on why it isn't autofixing it :/)

gregsdennis:
It's there in my editor. I put it there explicitly. 🤷‍♂️

Julian:
OK I *think* the reason pre-commit.ci wasn't working (wasn't autofixing) was because I moved this repo between orgs -- hopefully now that's fixed. Sorry bout that.

Looks good though, will fix + merge. (I think if you have VSCode set not to save with newline at the end it will strip it on save, even if it looks like it's there!)

PR #76, opened 2022-11-15
Add skipped tests for Hyperjump JSV

jdesrosiers:
Add skips for Hyperjump JSV.

I didn't figure out how to run it locally, so I hope it works.


Julian:
Awesome thanks! Will merge, let's see, seems fine. Also added you as a collaborator, feel free to make changes that seem sane at will (with or without review if they apply to your implementation alone).

PR #104, opened 2022-12-14
bump json-everything version to v4.0.0-beta2

gregsdennis:
bumping json-everything to the latest version


Julian:
It looks like this works but is rendering the version as `4.0.0` on the [report](https://bowtie-json-schema.github.io/bowtie/) -- I think because [GetLibVersion is truncating the value to just the 3-digit portion](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/dotnet-jsonschema-net/Program.cs#L157) -- if you get a chance could you tweak that? Otherwise will try to get to it at some point.

PR #113, opened 2023-01-04
Partially update the Hyperjump harness for 1.0

Julian:
@jdesrosiers I've partially updated this though not fully yet as I can't tell if there's still a way to get all dialects imported at once, right now this produces "unknown dialect" messages -- lemme know if you get a second to have a look.

### DIFF
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -1,6 +1,10 @@
 import * as path from "path";
 import readline from "readline/promises";
-import { addSchema, validate } from "@hyperjump/json-schema";
+import { addSchema, validate } from "@hyperjump/json-schema/draft-2020-12";
+import "@hyperjump/json-schema/draft-2019-09";
+import "@hyperjump/json-schema/draft-07";
+import "@hyperjump/json-schema/draft-06";
+import "@hyperjump/json-schema/draft-04";
 
 const hyperjump_version = await import("./package.json", {
   assert: { type: "json" },
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -1,12 +1,10 @@
-const path = require("path");
-const readline = require("readline/promises");
+import * as path from "path";
+import readline from "readline/promises";
+import { addSchema, validate } from "@hyperjump/json-schema";
 
-const hyperjump_version = require(path.join(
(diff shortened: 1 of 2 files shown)


implementations/js-hyperjump/bowtie_hyperjump.js
```diff
@@ -1,12 +1,10 @@
-const path = require("path");
-const readline = require("readline/promises");
+import * as path from "path";
+import readline from "readline/promises";
+import { addSchema, validate } from "@hyperjump/json-schema";
```
jdesrosiers, line 3: > I can't tell if there's still a way to get all dialects imported at once

In this version you need to be explicit about which dialects you want support for.

```suggestion
import { addSchema, validate } from "@hyperjump/json-schema/draft-2020-12";
import "@hyperjump/json-schema/draft-2019-09";
import "@hyperjump/json-schema/draft-07";
import "@hyperjump/json-schema/draft-06";
import "@hyperjump/json-schema/draft-04";
```

jdesrosiers:
Not sure it's the right thing for dependabot to auto-merge for major version changes. That's almost sure to break things. I wonder if it's possible to have dependabot only auto-merge it's PRs if CI passes. Also, it would probably be nice if there was a custom message we can set for dependabot that tags the author of the library so they get a notification. (I got a notification because I watch this repo, but I doubt all implementers will.)

implementations/js-hyperjump/bowtie_hyperjump.js
Julian: Ah I see, thanks! I saw that in your docs (that one should import `addSchema` and friends from the relevant version) but didn't realize it meant the `imports` themselves enable the draft even *without* doing so. Thanks, this indeed seems to work.

Julian:
Yeah definitely a good point.

> I wonder if it's possible to have dependabot only auto-merge it's PRs if CI passes. 

This I think is the main thing to fix, I'm not sure why CI didn't fail, I *think* because the image build job is configured to run just on release, so probably it should also run for PRs which touch the directory, which I thought I'd added but guess not. Will have a look tomorrow probably.

> Also, it would probably be nice if there was a custom message we can set for dependabot that tags the author of the library so they get a notification. 

Another good idea. This is certainly possible I think -- probably via `CODEOWNERS`. I'm not 100% sure though all authors are that "proactive" though, so far only you, Greg and the golang implementation author have really engaged with the repo here, I dunno, maybe we make it opt-in to get pinged?

PR #114, opened 2023-01-05
Fix Hyperjump JSV test harness

jdesrosiers:
All the Hyperjump tests are failing with an "Invalid URL" error. I noticed that the `fakeURI` variable is not a full non-relative URI as is required. That's probably the reason for the error. I'm surprised it worked in the previous version. It shouldn't have.

### DIFF
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -112,7 +112,7 @@ const cmds = {
       });
     } else {
       try {
-        const fakeURI = "bowtie.sent.schema." + args.seq.toString() + ".json";
+        const fakeURI = "https://example.com/bowtie.sent.schema." + args.seq.toString() + ".json";
         addSchema(testCase.schema, fakeURI, dialect);
         const _validate = await validate(fakeURI);
         results = testCase.tests.map((test) => {


implementations/js-hyperjump/bowtie_hyperjump.js
```diff
@@ -112,7 +112,10 @@ const cmds = {
       });
     } else {
       try {
-        const fakeURI = "bowtie.sent.schema." + args.seq.toString() + ".json";
+        const fakeURI =
+          "https://example.com/bowtie.sent.schema." +
+          args.seq.toString() +
+          ".json";
```
jdesrosiers, line 118: This is why I hate prettier. It should be called uglier.

Julian:
Thanks for the fix!

> This is why I hate prettier. It should be called uglier.

I definitely have no specific preference for JS formatting, I just wanted *something* automatic, so definitely PRs welcome if you want to tweak it!

Julian:
Weird, this still seems to be failing, I *assume* for some related reason, but the error message is bad. Will have to diagnose what's happening.

jdesrosiers:
@Julian Are we sure the build ran? The website says it was last ran 4 days ago, but this was merged 3 days ago.

Julian:
They're running, yeah, they're just failing -- you can see that here: https://github.com/bowtie-json-schema/bowtie/actions/workflows/report.yml where runs are failing because they somehow have inconsistent numbers of tests executed across implementations.

There's a double bug I have to find basically -- one is why whichever implementation isn't running all the tests, and the other is making that not fail the whole run (so at least other implementations are reported on).

PR #116, opened 2023-01-13
Attempt to fix Hyperjump JSV version not showing

jdesrosiers:
This might fix the problem with the version not showing for `hyperjump-jsv` on the homepage.


Julian:
Cool, will re-run the report! I think another way we could do it that I saw is to run `npm list --json` as a subprocess.

Julian:
Seems that indeed [did it](https://bowtie-json-schema.github.io/bowtie/), thanks!

PR #141, opened 2023-02-21
Add rust-boon implementation

santhosh-tekuri:
https://crates.io/crates/boon


implementations/rust-boon/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN update-ca-certificates
+
+WORKDIR /usr/src/myapp
+COPY . .
+
+ENV RUSTFLAGS='-C linker=x86_64-linux-gnu-gcc'
+RUN cargo build --target x86_64-unknown-linux-musl --release
+
+FROM alpine
+COPY --from=builder /usr/src/myapp/target/x86_64-unknown-linux-musl/release/bowtie-rust-jsonschema /usr/local/bin
```
Julian, line 14:
```suggestion
COPY --from=builder /usr/src/myapp/target/x86_64-unknown-linux-musl/release/bowtie-rust-boon /usr/local/bin
```

implementations/rust-boon/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "bowtie-rust-boon"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
Julian, line 6: Can you add this directory to .github/dependabot so it's kept autoupdated? It should go in 2 places, both to keep the `Dockerfile` updated and also your lib.

And also, can you add the directory to `.pre-commit-config.yaml` which should just ensure that `cargo fmt` is run on the rust file? You'll see `rust-jsonschema` already in there.

Julian (changes requested):
Nice, thanks!

It seems CI isn't running the image building job on PRs for some reason, I'll have to look into why and fix that, but this won't build yet for a minor reason which I left a comment on, just a small copypaste typo.

Other than that and the dependencies stuff looks great though, can you address those and then happy to merge, appreciated!

implementations/rust-boon/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+                started = true;
+                if request["version"] != 1 {
+                    Err("not version 1")?;
+                }
+                let response = json!({
+                    "ready": true,
+                    "version": 1,
+                    "implementation": {
+                        "language": "rust",
+                        "name": "jsonschema",
```
Julian, line 40:
```suggestion
                        "name": "boon",
```

implementations/rust-boon/Dockerfile
santhosh-tekuri: done

implementations/rust-boon/Cargo.toml
santhosh-tekuri: done

implementations/rust-boon/src/main.rs
santhosh-tekuri: done

PR #143, opened 2023-02-22
rust-boon: fix validation failures

santhosh-tekuri:


Julian:
There's also a security warning from a dependency that looks related: https://github.com/bowtie-json-schema/bowtie/security/dependabot/14

implementations/rust-boon/Dockerfile
```diff
@@ -7,10 +7,11 @@ RUN update-ca-certificates
...
 COPY . .
 
-ENV RUSTFLAGS='-C linker=x86_64-linux-gnu-gcc'
-RUN cargo build --target x86_64-unknown-linux-musl --release
+#ENV RUSTFLAGS='-C linker=x86_64-linux-gnu-gcc'
+RUN cargo build --release
 
-FROM alpine
-COPY --from=builder /usr/src/myapp/target/x86_64-unknown-linux-musl/release/bowtie-rust-boon /usr/local/bin
+FROM debian:bullseye-slim
```
Julian, line 13: Using a debian base image also looks like it takes the image size from ~20MB up to 80MB, so it'd be slightly nice to see if we can get it working on alpine like the rust-jsonschema image. FWIW it's not (only) macOS that the build fails on, CI is running Linux -- it's ARM64 that fails (on either OS) I think.

Either way, if we do end up having to stick with Debian please remove the dead comment.

santhosh-tekuri: reverted to alpine.

the original Dockerfile was not building on my laptop. so I updated to Debian. anyway alpine is fine for me.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM rust:1.67-slim AS builder
```
Julian, line 1: You meant to put this in implementations/rust-boon

santhosh-tekuri: I just copied it from rust-jsonschema/Dockerfile which uses the same.
do you suggest a change?

Julian: Your PR is adding a file called `Dockerfile` to the root of this repository, rather than to your implementation directory.

santhosh-tekuri: it seems your forced push has messed with my git and I got confused in resolving conflict. let me recheck

santhosh-tekuri:
closing this. created fresh pull request

PR #149, opened 2023-02-24
rust-boon: show boon library version used

santhosh-tekuri:


santhosh-tekuri:
this may not be quite perfect but should work for us.

Julian:
Great!

PR #153, opened 2023-03-02
rust-boon: upgrade to v0.3

santhosh-tekuri:


santhosh-tekuri:
the site still shows rust-boon v0.2.1

Julian:
I haven't done a release (of Bowtie), it'll be done when I'm back to a computer.

Julian:
The image build seems to be failing again.

santhosh-tekuri:
Can we split image building for AMD and ARM into separate tasks, so that if ARM build fails it won't affect AMD build. and the website can be generated from AMD images

Issue #127, opened 2023-01-18
Teach bowtie summary to show validation results


Julian:
Right now it only shows errors + failures (meaning deviations from expected results), but we should also probably give it a way to show valid/invalid for particular instances & schemas, particularly so that `bowtie validate` has a way to become "show me the results".

Julian:
Great! So right now if you run something like:

```
bowtie suite -i js-hyperjump -i dotnet-jsonschema-net https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft7/type.json | bowtie summary
```

that will run the official JSON Schema tests from [this test suite file](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft7/type.json) against 2 implementations (a Javascript one and a .NET one), and then should show you some output like:

```
                          Bowtie                          
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━━━┓
┃ Implementation             ┃ Skips ┃ Errors ┃ Failures ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━╇━━━━━━━━━━┩
│ hyperjump-jsv (javascript) │ 0     │ 0      │ 0        │
│ JsonSchema.Net (dotnet)    │ 0     │ 0      │ 0        │
└────────────────────────────┴───────┴────────┴──────────┘
                       80 tests ran                       
```

which essentially means 80 tests were run, and those two implementations passed all the tests (they didn't say an instance was invalid when it was expected to be valid or vice versa, and they didn't crash).

`bowtie summary` also works if you want to directly check a specific schema against an instance, so e.g.:

```
bowtie validate --implementation js-hyperjump --implementation dotnet-jsonschema-net <(printf '{"type": "integer"}') <(printf '["foo", 1]') | bowtie summary
```

that will validate the instance `["foo", 1]` against the schema `{"type": "integer"` using those two implementations and again you should see this output:

```
                          Bowtie                          
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━━━┓
┃ Implementation             ┃ Skips ┃ Errors ┃ Failures ┃
[...]

PR #163, opened 2023-03-10
Display Validation using bowtie summary command #127

dante381:
Added a command line flag `--show -s` with choices `[failures,validation]`.
Created a table to display the validation for multiple instances and schemas for different implementations.


dante381:
@Julian Could you review my changes.

Julian:
Hi @dante381 -- I'll likely be taking a break for the weekend, but will review on Monday -- initially though without yet trying to run this I saw a thing or two which I'll leave a comment on in a second in case you want to work on this between now and when I get a chance to review, but as I say I haven't reviewed carefully yet.

bowtie/_cli.py
```diff
@@ -154,31 +163,63 @@ def summary(input: Iterable[str], format: str | None):
         key=lambda each: (sum(each[1].values()), each[0][0]),  # type: ignore[reportUnknownLambdaType]  # noqa: E501
         reverse=True,
     )
+    validity = {
+        str(summary._combined[schemas]["case"]["schema"]["type"]): [
```
Julian, line 167: This is accessing a `type` property in the schemas which were tested -- I'm not sure what you intend by doing so -- a schema may not have a `type` keyword in it, and if it did, it may not be unique across all schemas that were tested -- someone may have used `{"type": "integer", "minimum": 7}` and some other schema says `{"type": "integer", "title": "my cool schema"}` -- but the data structures can get complicated to think about here.

You should likely be trying to build up a list rather than a dictionary -- one ordered by the order we evaluated schemas here, so no need for a key. It probably also is worth thinking about splitting this function in two now -- `combined` here as a list is meant to generate the table used for `--show failures`, and it's that that you're building an alternative to here in order to collect validation results. As I say, after the weekend I can try to provide more specific hints.

dante381:
> Hi @dante381 -- I'll likely be taking a break for the weekend, but will review on Monday -- initially though without yet trying to run this I saw a thing or two which I'll leave a comment on in a second in case you want to work on this between now and when I get a chance to review, but as I say I haven't reviewed carefully yet.

Sorry to disturb you. I will go over the suggestions and change the code accordingly.
Have a nice weekend @Julian.

Julian:
Hi @dante381 -- no worries at all, thanks again for giving this a shot!

This looks like a good start! A few notes:

If I run `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4/type.json | bowtie summary --show validation` I definitely see output. But if I run `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4 | bowtie summary --show validation` (where I am now running all of the tests for draft 4 rather than just the type tests) I don't, instead I see:

```
  File "/Users/julian/Development/bowtie/venv/bin//bowtie", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 169, in summary
    validity = [
               ^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 172, in <listcomp>
    {
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 173, in <dictcomp>
    str(result[0]["instance"]): [
                                ^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 175, in <listcomp>
    if result[1][implementation["image"]][0].valid is True
       ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'ghcr.io/bowtie-json-schema/lua-jsonschema'
```

which is likely coming from the fact that if an implementation errors (as lua-jsonschema does on some tests) then you have to handle the possibility that an implementation doesn't have a result (and probably show "error" instead of valid/invalid).

It's probably a good time to try to start adding tests to the `test_integration` file which will test the above scenario (as well as the "happy" one where everything works).

Secondly, I think I see the output we see now is something like:

```
  Schema: {'type': 'integer'}  
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ invalid          │
│ 1.1      │ invalid          │
│ foo      │ invalid          │
│ {}       │ invalid          │
│ []       │ invalid          │
│ True     │ invalid          │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
   Schema: {'type': 'number'}  
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ invalid          │
│ 1.0      │ valid            │
│ 1.1      │ valid            │
│ foo      │ invalid          │
│ {}       │ invalid          │
│ []       │ invalid          │
│ True     │ invalid          │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
   Schema: {'type': 'string'}  
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ valid            │
│ 1.1      │ invalid          │
│ foo      │ valid            │
│          │ valid            │
│ {}       │ invalid          │
│ []       │ invalid          │
│ True     │ invalid          │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
   Schema: {'type': 'object'}  
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ invalid          │
│ 1.1      │ invalid          │
│ foo      │ invalid          │
│ {}       │ valid            │
│ []       │ valid            │
│ True     │ invalid          │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
   Schema: {'type': 'array'}   
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ invalid          │
│ 1.1      │ invalid          │
│ foo      │ invalid          │
│ {}       │ valid            │
│ []       │ valid            │
│ True     │ invalid          │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
  Schema: {'type': 'boolean'}  
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ invalid          │
│ 0        │ invalid          │
│ 1.1      │ invalid          │
│ foo      │ invalid          │
│          │ invalid          │
│ {}       │ invalid          │
│ []       │ invalid          │
│ True     │ valid            │
│ False    │ valid            │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
    Schema: {'type': 'null'}   
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ invalid          │
│ 1.1      │ invalid          │
│ 0        │ invalid          │
│ foo      │ invalid          │
│          │ invalid          │
│ {}       │ invalid          │
│ []       │ invalid          │
│ True     │ invalid          │
│ False    │ invalid          │
│ None     │ valid            │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
  Schema: {'type': ['integer', 
          'string']}           
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ 1        │ valid            │
│ foo      │ valid            │
│ 1.1      │ invalid          │
│ {}       │ invalid          │
│ []       │ invalid          │
│ True     │ invalid          │
│ None     │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
            Bowtie             
                               
  Schema: {'type': ['string']} 
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ foo      │ valid            │
│ 123      │ invalid          │
└──────────┴──────────────────┘
         79 tests ran          
              Bowtie               
                                   
     Schema: {'type': ['array',    
            'object']}             
┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance     ┃ jsonschema (lua) ┃
┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ [1, 2, 3]    │ valid            │
│ {'foo': 123} │ valid            │
│ 123          │ invalid          │
│ foo          │ invalid          │
│ None         │ invalid          │
└──────────────┴──────────────────┘
           79 tests ran            
              Bowtie               
                                   
     Schema: {'type': ['array',    
        'object', 'null']}         
┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Instance     ┃ jsonschema (lua) ┃
┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ [1, 2, 3]    │ valid            │
│ {'foo': 123} │ valid            │
│ None         │ valid            │
│ 123          │ invalid          │
│ foo          │ invalid          │
└──────────────┴──────────────────┘
           79 tests ran            
```

for the `type.json` case -- maybe rather than repeating the `Bowtie` and `79 tests ran` headings it'd be nicer to output one single table with all of the cases. In other words, maybe rather than having `Schema: ...` be part of the heading, maybe try making it a column, in other words:

```
                         Bowtie             
┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃       Schema            ┃ Instance ┃ jsonschema (lua) ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ {                       │ 1        │ invalid          │
│   "type": "integer"     │ 1.1      │ invalid          │
│ }                       │ foo      │ invalid          │
│                         │ {}       │ invalid          │
│                         │ []       │ invalid          │
│                         │ True     │ invalid          │
│                         │ None     │ invalid          │
└─────────────────────────┴──────────┴──────────────────┘
│ {                       │ 1        │ invalid          │
│   "type": "number"      │ 1.1      │ invalid          │
│ }                       │ foo      │ invalid          │
│                         │ {}       │ invalid          │
│                         │ []       │ invalid          │
│                         │ True     │ invalid          │
│                         │ None     │ invalid          │
└─────────────────────────┴──────────┴──────────────────┘
│ {                       │ 1        │ invalid          │
│   "type": "string"      │ 1.1      │ invalid          │
│ }                       │ foo      │ invalid          │
│                         │ {}       │ invalid          │
│                         │ []       │ invalid          │
│                         │ True     │ invalid          │
│                         │ None     │ invalid          │
└─────────────────────────┴──────────┴──────────────────┘
                      79 tests ran          
```

something like that, where we use one unified table. Perhaps see if that's easily doable?


I'll leave a few more minor notes on the code, but thanks again!

bowtie/_cli.py
```diff
@@ -117,27 +117,37 @@ def report(input: Iterable[str], output: TextIO):
     default=None,
     type=click.Choice(["json", "pretty"]),
 )
+@click.option(
+    "--show",
+    "-s",
+    help="Displays the summary according to Argument",
```
Julian, line 123: Perhaps instead:
```suggestion
    help="Configure whether to display validation results (whether instances are valid or not) or test failure results (whether the validation results match expected validation results)",
```

bowtie/_cli.py
```diff
@@ -117,27 +117,37 @@ def report(input: Iterable[str], output: TextIO):
...
-def summary(input: Iterable[str], format: str | None):
+def summary(input: Iterable[str], format: str | None, show: str | None):
     """
     Generate an (in-terminal) summary of a Bowtie run.
     """
     if format is None:
         format = "pretty" if sys.stdout.isatty() else "json"
 
+    if show is None:
+        show = "failures" if sys.stdout.isatty() else "validation"
```
Julian, line 140: I'm not sure deciding the default based on whether `sys.stdout` is a TTY makes sense for this one. That means deciding whether we show one or the other based on whether someone is piping `bowtie summary` into something else or not -- it makes sense for configuring output format because if stdout is not a TTY usually it means someone's storing output into a file or something, and there JSON makes more sense.

Compare e.g. what `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4/type.json | bowtie summary` shows you to what `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4/type.json | bowtie summary | jq` shows you, where the latter is useful if you want to query the output (say, pull out information on one implementation).

For `--show` though I think we should just pick a default that applies always -- probably `validation`.

dante381:
> Hi @dante381 -- no worries at all, thanks again for giving this a shot!
> 
> This looks like a good start! A few notes:
> 
> If I run `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4/type.json | bowtie summary --show validation` I definitely see output. But if I run `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4 | bowtie summary --show validation` (where I am now running all of the tests for draft 4 rather than just the type tests) I don't, instead I see:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/julian/Development/bowtie/venv/bin//bowtie", line 8, in <module>
>     sys.exit(main())
>              ^^^^^^
>   File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
>     return self.main(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1055, in main
>     rv = self.invoke(ctx)
>          ^^^^^^^^^^^^^^^^
>   File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1657, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/julian/Development/bowtie/venv/lib/python3.11/site-packages/click/core.py", line 760, in invoke
>     return __callback(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 169, in summary
>     validity = [
>                ^
>   File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 172, in <listcomp>
>     {
>   File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 173, in <dictcomp>
>     str(result[0]["instance"]): [
>                                 ^
>   File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 175, in <listcomp>
>     if result[1][implementation["image"]][0].valid is True
>        ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
> KeyError: 'ghcr.io/bowtie-json-schema/lua-jsonschema'
> ```
> 
> which is likely coming from the fact that if an implementation errors (as lua-jsonschema does on some tests) then you have to handle the possibility that an implementation doesn't have a result (and probably show "error" instead of valid/invalid).
> 
> It's probably a good time to try to start adding tests to the `test_integration` file which will test the above scenario (as well as the "happy" one where everything works).
> 
> Secondly, I think I see the output we see now is something like:
> 
> ```
>   Schema: {'type': 'integer'}  
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ invalid          │
> │ 1.1      │ invalid          │
> │ foo      │ invalid          │
> │ {}       │ invalid          │
> │ []       │ invalid          │
> │ True     │ invalid          │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>    Schema: {'type': 'number'}  
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ invalid          │
> │ 1.0      │ valid            │
> │ 1.1      │ valid            │
> │ foo      │ invalid          │
> │ {}       │ invalid          │
> │ []       │ invalid          │
> │ True     │ invalid          │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>    Schema: {'type': 'string'}  
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ valid            │
> │ 1.1      │ invalid          │
> │ foo      │ valid            │
> │          │ valid            │
> │ {}       │ invalid          │
> │ []       │ invalid          │
> │ True     │ invalid          │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>    Schema: {'type': 'object'}  
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ invalid          │
> │ 1.1      │ invalid          │
> │ foo      │ invalid          │
> │ {}       │ valid            │
> │ []       │ valid            │
> │ True     │ invalid          │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>    Schema: {'type': 'array'}   
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ invalid          │
> │ 1.1      │ invalid          │
> │ foo      │ invalid          │
> │ {}       │ valid            │
> │ []       │ valid            │
> │ True     │ invalid          │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>   Schema: {'type': 'boolean'}  
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ invalid          │
> │ 0        │ invalid          │
> │ 1.1      │ invalid          │
> │ foo      │ invalid          │
> │          │ invalid          │
> │ {}       │ invalid          │
> │ []       │ invalid          │
> │ True     │ valid            │
> │ False    │ valid            │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>     Schema: {'type': 'null'}   
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ invalid          │
> │ 1.1      │ invalid          │
> │ 0        │ invalid          │
> │ foo      │ invalid          │
> │          │ invalid          │
> │ {}       │ invalid          │
> │ []       │ invalid          │
> │ True     │ invalid          │
> │ False    │ invalid          │
> │ None     │ valid            │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>   Schema: {'type': ['integer', 
>           'string']}           
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ 1        │ valid            │
> │ foo      │ valid            │
> │ 1.1      │ invalid          │
> │ {}       │ invalid          │
> │ []       │ invalid          │
> │ True     │ invalid          │
> │ None     │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>             Bowtie             
>                                
>   Schema: {'type': ['string']} 
> ┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ foo      │ valid            │
> │ 123      │ invalid          │
> └──────────┴──────────────────┘
>          79 tests ran          
>               Bowtie               
>                                    
>      Schema: {'type': ['array',    
>             'object']}             
> ┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance     ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ [1, 2, 3]    │ valid            │
> │ {'foo': 123} │ valid            │
> │ 123          │ invalid          │
> │ foo          │ invalid          │
> │ None         │ invalid          │
> └──────────────┴──────────────────┘
>            79 tests ran            
>               Bowtie               
>                                    
>      Schema: {'type': ['array',    
>         'object', 'null']}         
> ┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃ Instance     ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ [1, 2, 3]    │ valid            │
> │ {'foo': 123} │ valid            │
> │ None         │ valid            │
> │ 123          │ invalid          │
> │ foo          │ invalid          │
> └──────────────┴──────────────────┘
>            79 tests ran            
> ```
> 
> for the `type.json` case -- maybe rather than repeating the `Bowtie` and `79 tests ran` headings it'd be nicer to output one single table with all of the cases. In other words, maybe rather than having `Schema: ...` be part of the heading, maybe try making it a column, in other words:
> 
> ```
>                          Bowtie             
> ┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
> ┃       Schema            ┃ Instance ┃ jsonschema (lua) ┃
> ┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
> │ {                       │ 1        │ invalid          │
> │   "type": "integer"     │ 1.1      │ invalid          │
> │ }                       │ foo      │ invalid          │
> │                         │ {}       │ invalid          │
> │                         │ []       │ invalid          │
> │                         │ True     │ invalid          │
> │                         │ None     │ invalid          │
> └─────────────────────────┴──────────┴──────────────────┘
> │ {                       │ 1        │ invalid          │
> │   "type": "number"      │ 1.1      │ invalid          │
> │ }                       │ foo      │ invalid          │
> │                         │ {}       │ invalid          │
> │                         │ []       │ invalid          │
> │                         │ True     │ invalid          │
> │                         │ None     │ invalid          │
> └─────────────────────────┴──────────┴──────────────────┘
> │ {                       │ 1        │ invalid          │
> │   "type": "string"      │ 1.1      │ invalid          │
> │ }                       │ foo      │ invalid          │
> │                         │ {}       │ invalid          │
> │                         │ []       │ invalid          │
> │                         │ True     │ invalid          │
> │                         │ None     │ invalid          │
> └─────────────────────────┴──────────┴──────────────────┘
>                       79 tests ran          
> ```
> 
> something like that, where we use one unified table. Perhaps see if that's easily doable?
> 
> I'll leave a few more minor notes on the code, but thanks again!

Thank you for the review. I will do the aforementioned tasks ASAP. There is another issue while printing valid/invalid. If you check the first 3 schemas, the first instance of those schemas are displaying the wrong result. Could you help me with that?

I will start working on the tests as well and change the table format. I will also add the error output.

If i have any queries can i message you on slack as to avoid filling the PR with unnecessary comments?

Julian:
> If i have any queries can i message you on slack as to avoid filling the PR with unnecessary comments?

Sure, either way is fine with me (here or on Slack).

I'll try to have a look at what you mentioned about wrong results -- I'd recommend rather than the complicated list comprehension you might want to write that as a simple for loop at least to start, and maybe step through with a debugger.

But I can try to see whether I see the mistake too, will probably give this a shot tomorrow.

dante381:
Hi @Julian , 
I tried to make a unified table but couldn't find any methods to merge cell or parameters like rowspan. As for the tests I am still working on the test case that will cover all the results(valid,invalid,error).

I have a workaround to make a unified table but i will need to import pandas. Can I try this method out?

Was hoping you could help me with the previously mentioned problem. I simplified the list comprehension and printed the result.
I printed once inside the for loop and once in the else condition.

          for implementation in summary.implementations:
                valid=result[1].get(implementation["image"],"error")
                print(valid[0].valid) #output true
                if valid == "error":
                    z.append(valid)
                elif valid[0].valid is True:
                    z.append("valid")
                else:
                    print(valid[0].valid) #output false
                    z.append("invalid")

Julian:
OK, will have a look in a bit (perhaps during office hours) -- can you share a full command I can run to reproduce?

dante381:
`bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4/type.json | bowtie summary`

To check the results:

[check.txt](https://github.com/bowtie-json-schema/bowtie/files/10993049/check.txt)

Julian:
Your issue is because you're calling `str` on both instances and schemas -- it's not correct to do that, and the `str` of two different JSON objects which are equal may be different, or vice versa, the str of two different JSON objects which are unequal may be the same -- e.g. in Python `str(1) == str("1")`.

Also, not all instances are going to be hashable, which maybe is the reason you were trying to call `str` on them in the first place -- but rather than putting instances in a dictionary, if you simply take them as-is and append them to a (sub)list or tuple where each is `(instance, validity)` or something like that, then you neither need to call `str` nor have them be hashable.

And when you print them out, you should call `json.dumps` so that what someone sees is the JSON dumped version of the instance, but I can point that out during code review if needed.

dante381:
Ya It worked.
I will finish writing the test asap.
Thank you @Julian

Julian:
Great!

dante381:
Hey @Julian,
I have added the tests but for there seems to be an issue with envsonschema. It is returning false for every instance.
So I used the `js-hyperjump` implementation to validate.

Another issue is,
While displaying the "error" result in summary for implementation errors it works only for test suite as they contain vaild and invalid schemas which follow the implementation format whereas while validating specific schemas against instances. The implementation does not get added to the list returned in `_report.from_input(input)["summary"]`  in line 139 in `bowtie/_cli.py`.
The implementations specified should be returned despite the error so that we can display the error count as well as the validation result.
It is just my understanding of the code upto now.

Is this a different issue? or shall i work on this?

dante381:
> I don't 100% follow what you're describing -- it's basically always easiest if you include a command that shows me what you mean -- what command are you running? I'm not following specifically what you mean by displaying "error" results, I understand you're saying something possibly about data being missing from the report that's needed in order to implement something.

    bowtie validate -i lua-jsonschema <(printf '{}')  <(printf '12') | bowtie summary

The above command displays an error, since the `_report.from_input(input)` method does not return any implementations due to unsupported dialects.
I was thinking that the above method should return the implementation along with the error count.

Julian:
Got it -- yes, that sounds like a good thing to fix in a separate issue (feel free to open one!). It's similar to what happens if you run `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft2020-12/ | bowtie report`, where the implementation doesn't support that dialect, so `bowtie report` fails.

I've been aware of the issue but it's pretty minor so I hadn't fixed it, but it's certainly reasonable to address.

dante381:
Hi @Julian, 
Could you take a look at the changes?
Thank you

bowtie/_cli.py
```diff
@@ -115,12 +115,21 @@ def report(input: Iterable[str], output: TextIO):
...
+    default="validation",
+    type=click.Choice(["failures", "validation"]),
+)
 @click.argument(
     "input",
     default="-",
     type=click.File(mode="r"),
 )
-def summary(input: Iterable[str], format: str | None):
+def summary(input: Iterable[str], format: str | None, show: str | None):
```
Julian, line 132: `show` is never `None` now is it? It's always a `str`?

tests/test_integration.py
```diff
@@ -462,23 +462,46 @@ async def test_validate(envsonschema, tmp_path):
 
 @pytest.mark.asyncio
 async def test_summary(envsonschema, tmp_path):
-    tmp_path.joinpath("schema.json").write_text("{}")
-    tmp_path.joinpath("instance.json").write_text("12")
+    tmp_path.joinpath("schema.json").write_text(
```
Julian, line 465: Rather than modifying the existing test (which tests the existing behavior) you should add a second one which tests your new behavior, and leave the first one in place, because we don't want to *stop* testing the old behavior, we want to make sure some new behavior is *also* tested.

In other words, a second test called, say, `test_summary_show_validation`, which you can mostly copy from this test case, add your `--show validation` option as you have, and then likely the rest of the test as you have modified here.

You may need more than one in fact, perhaps 2 or 3, which cover one case where there's no errors or skipped tests, just valid or invalid instances, and then a second test where there's a skipped test and a test that an implementation errored on. This way we can be sure the code written works without testing it manually.

bowtie/_cli.py
```diff
@@ -153,29 +161,82 @@ def summary(input: Iterable[str], format: str | None):
...
 
+    validity = []
+    for _, schemas in summary._combined.items():
+        schema = []
+        instances = []
+        for result in schemas["results"]:
+            validation = {}
+            for implementation in summary.implementations:
+                valid = result[1].get(implementation["image"], "error")
+                if valid == "error":
```
Julian, line 172: I think this probably only handles 1 error case, which is why adding another test or two like I mentioned below is likely going to be helpful.

A test case can error either at the "case level" or at the "test level". In other words, if I have a schema: `{"foo": "bar"}` and I validate 3 instances against it, an implementation can error:

* by blowing up entirely at the schema `{"foo": "bar"}` because it has some bug interpreting that schema
* by blowing up for 1 of the 3 instances but not others, because it has some bug with a particular edge case

I *think* this may only handle one of the two cases but it'll be clearer if you try to write a test for each -- write a test where you put "crash:1" in the description of a test *case* that you pass to `envsonschema`, and a second test case where you put "crash:1" in the description of an individual *test* you pass but not in the test case. And then we'll be able to see whether we're handling both cases.

Let me know if that doesn't make sense to you!

Julian (commented):
This is getting close. I left some notes on improving/adding another test or two, which may reveal a bug in the code. Let me know if you have trouble and I can show you specific examples!

Julian:
@dante381 (GitHub is slightly confused) but I pushed a branch with some more help [here](https://github.com/bowtie-json-schema/bowtie/commits/bowtie-summary) in a branch called "bowtie-summary". Do you know how to work on top of that? In general it's a lot easier (for both you and me!) if you work on a git branch when you make PRs rather than directly on your fork's main branch. Let me know if you need more help doing so (you can see GitHub is confused now because when I pushed back to your branch I hadn't realized it was your main branch rather than a separate one).

That branch should hopefully give you some tips on the test (which now passes I believe if you run it).

But the code is still broken! Do you want to give a shot to figuring out why?

Specifically, if you run something like this from the branch:

```
⊙  bowtie suite -i python-jsonschema -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft7/type.json | bowtie summary                                                                                                                 julian@Postmac
/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.10/site-packages/click/types.py:82: ResourceWarning: unclosed <ssl.SSLSocket fd=6, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.254.6.90', 56837), raddr=('140.82.113.6', 443)>
  return self.convert(value, param, ctx)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.10/site-packages/click/types.py:82: ResourceWarning: unclosed <ssl.SSLSocket fd=7, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.254.6.90', 56838), raddr=('140.82.112.9', 443)>
  return self.convert(value, param, ctx)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
2023-04-03 16:15.59 [info     ] Will speak                     dialect=http://json-schema.org/draft-07/schema#
2023-04-03 16:16.00 [warning  ] Implicit dialect not acknowledged. Proceeding, but implementation may not have configured itself to handle schemas without $schema. [ghcr.io/bowtie-json-schema/lua-jsonschema] dialect=http://json-schema.org/draft-07/schema# response=StartedDialect(ok=False)
2023-04-03 16:16.00 [info     ] Finished                       count=11
                                      Bowtie                                       
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Schema              ┃                                                           ┃
┡━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│                     │                                                           │
│                     │   Instance   jsonschema (lua)   jsonschema (python)       │
│                     │  ───────────────────────────────────────────────────      │
│                     │   1          valid                                        │
│                     │   1.0        valid                                        │
│ {                   │   1.1        invalid                                      │
│   "type": "integer" │   "foo"      invalid                                      │
│ }                   │   "1"        invalid                                      │
│                     │   {}         invalid                                      │
│                     │   []         invalid                                      │
│                     │   true       invalid                                      │
│                     │   null       invalid                                      │
│                     │                                                           │
├─────────────────────┼───────────────────────────────────────────────────────────┤

```

you'll see that output. Where notice there's only one column shown!

If you'd like to give a shot to fixing that you're welcome to, otherwise I can work a bit more on merging this tomorrow.

Here's 2 hints on what's going on:

* You have a dictionary keyed by the name of an implementation (e.g. `jsonschema`) but alone, that's not unique! See, there's two implementations called "jsonschema". You either need to include the language name too or to not use a dict and to instead keep track of which implementation it is positionally.
* Another reason you may prefer positionally is that even once you fix the above, you may not have the right *order* in the table. You're doing something like `*implementations.values()` but you need to be sure that that actually corresponds to the same order that you have in the column headings. In other words, you might be permuting the columns by showing the wrong implementations' results in some other implementation's column.


As I say, maybe the tweaks I made give you a bit more of a hint, speak up if you're having issues and I can try to give you a hand again!

dante381:
Hey @Julian ,
I opened a new pr and made the changes in the bowtie-summary branch.

Julian:
Perfect! Will have a look in a bit!

Issue #146, opened 2023-02-23
Shrink the raw report page size


Julian:
Right now each HTML report when run on the suite is ~10MB, probably because of all the JSON blobs for schemas and instances that we write into each page.

A compressed zip of the site though is only a few MB total, so clearly stuff compresses well.

We likely anyhow should split the raw data from the DOM-related parts of the report (and have the latter read into the former from some simple client-side JS), but perhaps the above is a reason to even write the raw data out compressed and decompress bits of it client-side.

AgniveshChaubey:
Hey @Julian, I wanna contribute to `Bowtie and the World of JSON Schema`. If you could provide some resources for getting started, it'd be really helpful to me.

Julian:
Hi there! I just left some links [here](https://github.com/postman-open-technologies/gsoc-2023/issues/15#issuecomment-1451328745) -- have a look and let us know if you get stuck!

AgniveshChaubey:
Hello @Julian, I went through the documentation provided in the above link and became somewhat familiar with JSON Schema. Could you please provide some instructions for setting up the environment locally so that I can get familiar with the UI component?

Julian:
Great! What OS are you on? Did the install instructions from the link fail for you or were you able to run any of them?

AgniveshChaubey:
> Great! What OS are you on? Did the install instructions from the link fail for you or were you able to run any of them?

I am having windows. I have successfully installed the bowtie cli but don't know how to set up the website locally! Is the website code hosted publicly?

Julian:
Great! Then if you run one of [these examples](https://bowtie-json-schema.readthedocs.io/en/latest/cli/#running-a-single-test-suite-file), including piping into `bowtie report`, you should end up with a file called `bowtie-report.html` which you can open in a browser, that's the current UI!

Julian:
That error *should* be saying it can't find your docker -- not a container, dockerd itself -- if you run `docker ps` does that work? If not make sure you installed docker.

AgniveshChaubey:
Hello @Julian, all the problems I was having have been fixed, and the bowtie is now fully configured and running properly, thanks for your kind support!

[...]

PR #165, opened 2023-03-12
Reducing the file size of HTML Report 

AgniveshChaubey:
This PR refers to #146 

Issue: The file size of generated HTML report is around 10MB.

Aim: To reduce the size of generated HTML report.

This PR implements `zlib` library to compress the HTML report.

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -12,6 +12,7 @@ import json
 import os
 import sys
 import zipfile
+import zlib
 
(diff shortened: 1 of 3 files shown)


Julian:
@AgniveshChaubey I think going to close this one given you're working on the other angle -- but if we end up getting back to this ticket we can of course open a new PR (and of course if you choose to leave this one alone that's totally fine too!)

Issue #53, opened 2022-10-31
Show the combined errors+failures number in reports


Julian:
This is really the total number of "incorrect" tests run by an implementation regardless of whether they fail or blow up. Right now we show these numbers separately.

aku1310:
Hi @Julian,

I have set up the repo on my system, and going through the code base and documentation. I would like to work on this issue. Is there anything that I should keep in mind as I go through the codebase to resolve this issue? I am also new to this so any guidance would be appreciated.

If possible, can this issue be assigned to me so I can start working on resolving this?

Julian:
Hi there -- great! Super appreciated -- I'm just thinking out loud on a design, but I think for this one a nice thing would maybe be instead of the 3 columns (skipped, failed, errored) to have just 2, skipped and unsuccessful (which should be failed + errored) -- and then to make it so that if you mouse over the number for unsuccessful to get a tooltip which shows the 2 numbers separately (failed and errored) and which also explains what those 2 categories mean -- failed means an implementation didn't crash but got the wrong answer, and errored means it crashed completely when trying to calculate an answer (which it should not).

Does that help? Let me know if you have any further questions.

aku1310:
Thank you @Julian, this surely helps! I will get right into it and see how to make it work.

aku1310:
Hi @Julian, I have made changes in the _cli.py, _report.py, and report.html.j2 files to show the failed and errored as unsuccessful by creating unsuccessful attribute in the _report.py and also added the necessary functions but when running the test on my local system, no changes are being shown in the report created. I feel that has to do with how I'm using the command from [here](https://bowtie-json-schema.readthedocs.io/en/latest/cli/#running-test-suite-tests-from-local-checkouts).

`bowtie suite $(find full/path/to/bowtie/implementations/ -mindepth 1 -maxdepth 1 -type d | sed 's/.*\/implementations\//-i /') full/path/to/JSON-Schema-Test-Suite/tests/draft7/type.json | bowtie report` 

![image](https://user-images.githubusercontent.com/72210364/224478156-8eb7aad4-e494-45e4-b856-a96e5288a392.png)

Could you please point me in the right direction as to what I might be doing wrong and how to go about it?

Julian:
[...]

PR #166, opened 2023-03-13
Showing the combined errors + failures number in reports 

aku1310:
This PR resolves issue [#53](https://github.com/bowtie-json-schema/bowtie/issues/53) and makes the following changes:

1. Shows the combined errors + failures numbers as unsuccessful in the report.
2. Gives the description of errors and failures when hovering over the unsuccessful heading.
3. Shows the errors + failures numbers when hovering over any unsuccessful number in the report.

### DIFF
--- a/bowtie/templates/report.html.j2
+++ b/bowtie/templates/report.html.j2
@@ -90,9 +90,10 @@
               <tr>
                 <th scope="col" class="text-center">errors</th>
 
-                <th scope="col" class="table-bordered text-center">failed</th>
                 <th scope="col" class="table-bordered text-center">skipped</th>
-                <th scope="col" class="table-bordered text-center">errors</th>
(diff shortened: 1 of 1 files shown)


Julian:
Nice! This looks quite good.

Perhaps the only feedback is it seems it can be slightly hard to mouse over a row if you have a tooltip open which is covering a row above it.

In other words if a tooltip is open on the bottom row, you have to move the cursor away and back again if you want to see the tooltip for the second to last row.

It's possibly a bit better for that case if we open the tooltip to the left (so it overlaps the skipped column), rather than centered on top of the row.

Let me know what you think!

aku1310:
That sounds like a better idea for the text to appear over the skipped column instead on top. I will make the required changes as per the suggestion.

Julian:
Great! Think this looks good! Going to merge, really appreciated!

Issue #55, opened 2022-11-05
Generate badges for implementations


Julian:

VishrutAggarwal:
Hi,

I would like to work on this issue, can I get some more details regarding the same?

1. What exactly should the badges should show and represent?
2. What do you mean by supported drafts?
3. Which file should I look out for while working on resolving this issue?

Julian:
Hi @VishrutAggarwal thanks for the offer! This'd be super helpful to do, and apologies for the issue description not being very good.

Here's some detail:

The main goal is to produce badges usable via that shields.io link. There are a few different kinds of badges we'd want to generate for every implementation -- and the way you do so essentially is by producing a JSON file (as part of Bowtie's continuous integration) and then pointing shields.io at the URL for this JSON.

Specifically as an example of each:

 Supported drafts should be a simple badge that indicates which [versions of JSON Schema](https://json-schema.org/specification-links.html) an implementation supports. As an example, if you run `bowtie info -i clojure-json-schema` you'll see that the Clojure jsonschema implementation says:

```
⊙  bowtie info -i clojure-json-schema                                                                                                                                                                                                        julian@Postmac
name: "json-schema"
language: "clojure"
homepage: "https://github.com/luposlip/json-schema"
issues: "https://github.com/luposlip/json-schema/issues"
dialects: [
  "http://json-schema.org/draft-07/schema#",
  "http://json-schema.org/draft-06/schema#",
  "http://json-schema.org/draft-04/schema#"
```

i.e. it supports draft 7, 6 and 4.

So we want to generate a badge that looks like:

![badge](https://img.shields.io/badge/JSON%20Schema%20Versions-4%2C%206%2C%207-Green)

(which I manually generated from this link: https://img.shields.io/badge/JSON%20Schema%20Versions-4%2C%206%2C%207-Green but we would want to generate dynamically for every implementation in the way I mentioned above).

I'll explain the rest of the badges:

* per draft compliance % should be a badge which for each implementation summarizes the results of https://bowtie-json-schema.github.io/bowtie/draft7 (or each draft correspondingly) -- for example, on that draft 7 page, the VSCode JSON Schema implementation fails 106 out of 1176 tests (which is 91% passing) -- so we'd want to generate a badge saying:

![badge](https://img.shields.io/badge/Draft%207-91%25%20Passing-yellowgreen)

[...]

PR #175, opened 2023-03-20
Generate badges for passing test percentages

VishrutAggarwal:
Added a badge that displays the Passing test case percentage per draft.


Julian:
(Moving conversation from https://github.com/bowtie-json-schema/bowtie/issues/55#issuecomment-1480226256 )

This branch seems fine as a start, it seems to already have the right behavior around making the `--badges` argument optional. I made minor improvements, but yeah I think you're already on the way here.

A good idea would be to work now on:

* outputting badges for all implementations, not just one -- this probably means we really need the option to be a *directory* rather than a file. i.e. we ouput `badges/<implementation_name>` rather than one single `badges.json`, as each implementation will have their own badges
* making the colors "correct" for the badge -- right now you hardcode green but probably we should pick some ranges, maybe below 65 is red, above 65 is amber, 85 is yellow, 95 is light green, 100 is green (but we can tweak that range later)
* one or more tests for the expected badge output
* I think you probably are missing accounting for errored and skipped tests in your calculation -- probably the right formula is `(total - skipped - failed - errored) / total`

Let me know if you have more questions but yeah as I say I think this seems like it's definitely a start. Thanks again!

VishrutAggarwal:
Hi,

Here's the progress I've made:
1. Corrected the passing percentage formula.
2. Added a preliminary color code range (it can be tweaked and changed when proper defining of the ranges is done, as of now, it is just for testing purposes).

Here's where I am facing problems:
So I've been trying to run the command `bowtie report --badges badges.json <report.jsonl` and have been so far unsuccessful.
The report gets somehow consumed by the `output.write(template.render(**report))` line.
I've tried copying the report to another variable and use that, somehow it still doesn't work as intended.

I am currently stumped by the matter and haven't been able to test my code either.
Here is the code change I made so that I could test something (if the report still gets consumed and if the interpreter even runs the `if` statement):

image

Here are the traceback calls:

image

P.S. Sorry for the late updates, I have fallen sick again (probably attributing to seasonal changes and new environment). I'll try my best to be regular on the updates and changes.

Julian:
In a shell, if I run `seq 20`, that means "generate all the numbers from 1 to 20 and print them out".

If I run `seq 20 | grep 2`, that means "Generate all the numbers from 1 to 20, and print out only the lines with a 2 on them. The `|` basically takes all the output `seq 20` produces and gives it as *input* to a second command, `grep 2` (one that filters the lines).

If I run `seq 20 >output`, the `>output` means "instead of printing the numbers, put all the output in a new file called `output`" (try it, you'll see no output printed, it'll all be in the file).

I can now say `grep 2 <output`, where now `<output` means "instead of reading input like you would with `|`, make believe the input comes from the file called `output`". So you now should see the same lines printed as if you run `seq 20 | grep 2`, just the output was actually "permanently" stored in the file called `output`. You can run `grep 2 <output` over and over again without running `seq` at all, as the input is stored in the file.

If you run `bowtie suite ... | bowtie report`, you're doing the equivalent of `seq 20 | grep 2` where you immediately process the output of running bowtie suite.

If you instead want to store that in a file, you can run `bowtie suite ... >results.jsonl` to now get a file, and *not* include `| bowtie report` in the command.

Once you have that file, you can now repeatedly run `bowtie report <results.jsonl` to use it as input over and over again.

What you can't/shouldn't do is *both* have `|` *and* `<results.jsonl`, as that doesn't make sense -- are you trying to generate a report using data being produced right now as input? If so `bowtie suite ... | bowtie report`. Are you trying to generate a report using data generated previously and stored in a file? If so `bowtie report <results.jsonl` with *no* `bowtie suite` command.

Does that help?

Julian:
(Sorry, GitHub submitted the comment...) And nice! I'll have a look at the other changes. Hope you feel better.

VishrutAggarwal:
>Does that help?

Yes it does, I was running `bowtie suite ... | bowtie report --badges badges.json <report.jsonl` without understanding that `<` takes an input from an already existing file and doesn't put the previous input into a new file, ie. I had myself mixed between `<` and `>`... and a whole lot of stuff. I need to brush up on some of my scripting skills so that it doesn't happen again.

I have successfully made it work thanks to your help and have fixed the _range bug_ that I found while testing my code.

My plan at the moment:

- [ ] 1. I will work on the other 3 badges, of which I have a decent idea on how to implement the supported dialects.
- [ ] 2. After which I will look into generating a directory with multiple JSON files.

I will require a little guidance on:
1. How to get the `overall compliance | %` badge data
2. How to determine the _keywords with failing test cases_ (I might be misunderstanding the requirement here, are there any pre-defined keywords or do we have to define the keywords ourselves?)
3. Should there be a particular naming convention for the generated JSON files? or should the input be taken from the Command Line Argument itself (and if yes, how should it be implemented) -- (in my opinion, it would be better to keep pre-defined names for the JSON files and just take the name of the generated directory that stores them from the CLA).
4. Does it need to be in such a way that the user chooses which badges to be generated? (which in my opinion would increase the number of arguments unnecessarily) or generate all of them (which would be a lot easier to implement).

Julian:
> After which I will look into generating a directory with multiple JSON files.

Let's work on this first, as once you've done so we can merge the PR immediately and then do the additional badges in a follow-up.

> Should there be a particular naming convention for the generated JSON files?

Let the `--badges` argument specify a directory, and then generate one file in it for each implementation, where the name of the file is the name of the implementation, e.g. `badges/ruby-json_schemer.json`.

> Does it need to be in such a way that the user chooses which badges to be generated?

Always generating all of them sounds good to me!

Julian:
> Do we need to use click module for generating the directory? If yes, I am currently unable to find the corresponding functionality in click's documentation.

Use [click.Path](https://click.palletsprojects.com/en/8.1.x/api/#click.Path) and specify `pathlib.Path` as your `path_type`, which will give you that kind of object, which you can then call the `.mkdir` method on to create the directory.

Julian:
Going to close this given we've merged #175, sorry again for the confusion there, we can pick up the discussion on a follow-up for you if you're still interested, I saw you found the other badge PR that's in progress, which you may want to help out on.

PR #181, opened 2023-03-26
Implemented display schema to show the schema,  when clicked on test

nomandhoni-cs:


nomandhoni-cs:
Hey @Julian, Please have a look to display schema feature when click on the test.
https://user-images.githubusercontent.com/92979541/227791511-0fac846b-6811-487c-9d54-3b43e4d91848.mp4

Julian:
Thanks! Overall this kind of idea seems quite reasonable to me obviously, like we discussed, but exactly as-is I see a couple of things that make me pause.

I think it's jarring if you "expand" an example lower on the page and something appears way up on top of the page - someone expanding an example is trying to look at the sub-rows of that example.

I think it'd be a lot clearer if the new card you're creating was attached or inside of the element that's expanding, rather than detached at the top of the page (particularly if there are many rows on the page and someone is scrolled way down on the page).

So I think it'd be quite a bit more useful if the schema appeared connected to the expanded rows, not at the top.

Does that make sense?

PR #189, opened 2023-03-30
Add runtime info to harness' metadata

AgniveshChaubey:
### What is this PR for?
This PR adds runtime information (including `os`, `os_version`, and `language_version`) to the implementation harness' metadata which is returned by `bowtie info` !

This is a subset PR that adds runtime info to JavaScript and Python based harnesses, and other harnesses' info will be added down the line in this PR itself.

Edit:
Golang based harnesses are also modified to add runtime info.

### DIFF
--- a/bowtie/schemas/io-schema.json
+++ b/bowtie/schemas/io-schema.json
@@ -97,6 +97,7 @@
                     "language_version": {
                       "type": "string",
                       "description": "Version of language used to run the implementation"
+                    }
                   }
                 }
               }
--- a/bowtie/schemas/io-schema.json
+++ b/bowtie/schemas/io-schema.json
@@ -85,7 +85,18 @@
                           "url": { "type": "string", "format": "uri" }
                         }
                       }
-                    }
+                    },
+                    "os": {
+                      "type": "string",
+                      "description": "Operating system the implementation is running on"
+                    },
+                    "os_version": {
+                      "type": "string",
+                      "description": "Version of OS the implementation is running on"
+                    },
+                    "language_version": {
+                      "type": "string",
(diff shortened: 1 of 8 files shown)


Julian:
Hey there! Nice! Well done. I think this probably looks good, can you address the [CI go fmt failure](https://github.com/bowtie-json-schema/bowtie/actions/runs/4561548119/jobs/8047634358?pr=189#step:4:128)?

Julian:
Also, please add the new fields to the [schema](https://github.com/bowtie-json-schema/bowtie/blob/main/bowtie/schemas/io-schema.json) for bowtie (probably by just requiring that they be strings, but adding a small bit of description on what the interpretation or meaning of the value each is expected to be).

AgniveshChaubey:
Go files have been formatted and schema file is also updated with new fields!

Julian:
Well done, looks great! Merging (the 2 CI failures are because we still are having difficulties building Rust implementations in CI unfortunately...)

Will do a release which should produce images containing your changes. Thanks again!

AgniveshChaubey:
It's my pleasure to be a part of this project. :)
 
>  (the 2 CI failures are because we still are having difficulties building Rust implementations in CI unfortunately...)

Can you provide some more details on the 'difficulties' mentioned above

Julian:
It's been fixed today thankfully actually. #192 has some details if you're curious!

AgniveshChaubey:
Got it! The tests were failing due to image size and setting the env variable `CARGO_REGISTRIES_CRATES_IO_PROTOCOL` to `sparse` worked pretty well in this case.

Great to hear it's all sorted now.

PR #192, opened 2023-04-02
Build rust images using cargo's sparse registry option, reducing their memory usage

santhosh-tekuri:
https://github.com/actions/runner-images/discussions/7188

the above issue says it ubuntu-20.04 is not affected by this

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -35,7 +35,7 @@ jobs:
 
   build:
     needs: list
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     permissions:
       packages: write
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -35,7 +35,7 @@ jobs:
 
   build:
     needs: list
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     permissions:
       packages: write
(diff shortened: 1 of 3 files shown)


Julian:
As you can see this still seems to be failing.

I'm certainly ok changing versions if we find one that works. Though I'm slightly skeptical this sort of idea really works at all -- my guess is GitHub upgraded worker versions, (gave them less resources on the newer version) and prioritized the new rollout on newer versions of the runners, so it only seemed like 20.04 still worked because they were using the older runner setup.

But if you manage to get this working definitely happy to merge...

Julian:
Nice, I think now *that* change appears to have made progress! Thanks for finding that option. Can you mention where you found it / how you suspected it'd fix things?

The last remaining failure seems clojure related because of the Ubuntu version change. Perhaps change back to latest now to see whether the new option you added makes everything work, given 20.04 didn't seem related? Otherwise if it turns out we need both 20.04 *and* the sparse registries option you probably need to install some other version of Java on 20.04 for the Clojure build.

(Looks like the other rust build is still running so I guess there's a chance it still fails but hopefully not.)

santhosh-tekuri:
with the new change of using space protocol in rust build, rust-boon is passing.

it is still struck in st-jsonschema because of resources.

can you cancel the build. I will try to fix the closure image failure and try new build

Julian:
Sure, done.

santhosh-tekuri:
the rust build was failing with exit code 127. and I found https://stackoverflow.com/questions/74643818/docker-buildx-gives-exit-code-137-with-cargo

santhosh-tekuri:
rust-boon passed even after reverting back to ubuntu-latest.

rust-jsonschema is still struck: `Job is waiting for a hosted runner to come online.`

santhosh-tekuri:
all images successfully build.

PR #194, opened 2023-04-03
report: add links to other dialects

santhosh-tekuri:
the UI/CSS may not be pretty. but it works

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -45,7 +45,7 @@ jobs:
 
       - name: Generate the Report
         run: |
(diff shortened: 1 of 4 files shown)


Julian:
Nice! I think this definitely is better than the current "no link whatsoever" -- a slight improvement might be conditionally emitting this only if some option is passed (e.g. `--generate-dialect-navigation`) so that someone running locally doesn't end up with a page with broken links on it when they're not generating all of the dialect reports.

But I'm OK with merging this even as-is if you don't feel like trying to do the above.

santhosh-tekuri:
added the flag you mentioned.
I don't have complete dev setup. I just did adhoc editing from vi.
but I am able to test it.

plz do the code review

Julian:
Looks great! Merged, thanks again.

Issue #150, opened 2023-02-24
When PRs touch the implementation directory, only build that one image, not all


Julian:

santhosh-tekuri:
fixed with #197

santhosh-tekuri:
NOTE: if you just change images.yaml workflow, it build no image

PR #195, opened 2023-04-03
ci: build image only when there is some change

santhosh-tekuri:


santhosh-tekuri:
@Julian can you check why precommit failing. I don't get what it is expecting

.github/workflows/images.yml
```diff
@@ -16,27 +16,37 @@ on:
...
-          from pathlib import Path
-          import json
-          paths = [
-              str(path.name)
-              for path in Path("implementations").iterdir()
-              if path.is_dir()
-          ]
-          print(f"images={json.dumps(paths)}")
-          ' >> $GITHUB_OUTPUT
+    - uses: dorny/paths-filter@v2
```
Julian, line 27: I saw this action and am OK with using it (#150 is the relevant issue which you may have seen), but I don't want to maintain this duplicate list of implementations. In other words, I'm OK with merging this only if you dynamically build the `filters` mapping below, or else at worst ensure that every single implementation in the directory is guarded so it must appear in that list.

Julian:
The indentation is off, it's running prettier and finding a difference. You can run pre-commit locally or else it should automatically fix the commit though I think GitHub actions is having some issue this afternoon

PR #197, opened 2023-04-03
ci: build image only when there is some change

santhosh-tekuri:


santhosh-tekuri:
@Julian I am able to get it without listing each impl. plz check

Julian:
Looks OK to me (though it'd be slightly clearer even to split into 2 jobs so you don't have to keep including the conditional, but whatever).

No way to know it works 100% until we try editing an implementation though right? Or have you already double checked that?

Feel free to merge if so.

santhosh-tekuri:
to test we have to merge this.

merging...

PR #198, opened 2023-04-03
ci: test image build

santhosh-tekuri:


santhosh-tekuri:
as expected, it is only building go-jsonschema image.

Julian:
Nice!

PR #200, opened 2023-04-04
Add runtime info to Bowtie report

AgniveshChaubey:
### What is this PR for?
This PR displays runtime information (including `os`, `os_version`, `language` and `language_version`) to Bowtie Report!
Here bootstrap **_modal_** is used to to display the above additional information.

### DIFF
--- a/bowtie/templates/report.html.j2
+++ b/bowtie/templates/report.html.j2
@@ -190,10 +190,9 @@
                   </th>
                   <td>
                     <small class="font-monospace text-muted">{{ implementation.get("version", "") }}</small>
-
-                    <button type="button" class="btn btn-sm btn-secondary" data-bs-toggle="modal" data-bs-target="#implementation-{{ loop.index }}-runtime-info">
-                      more info
-                    </button>
+                    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="#242624" class="bi bi-info-circle-fill" viewBox="0 0 16 16" data-bs-toggle="modal" data-bs-target="#implementation-{{ loop.index }}-runtime-info">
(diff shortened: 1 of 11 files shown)


AgniveshChaubey:
Hi @Julian, Please have a look at this PR!

Julian:
I've seen it and will have a look in a bit, no need to re-ping until at least a day or two has gone by, though I certainly hope to have feedback before then!

AgniveshChaubey:
I'm sorry for this! I thought you missed the notification, that's why... Will take care from now onwards :)

bowtie/templates/report.html.j2
```diff
@@ -190,6 +190,10 @@
                   </th>
                   <td>
                     <small class="font-monospace text-muted">{{ implementation.get("version", "") }}</small>
+
+                    <button type="button" class="btn btn-sm btn-secondary" data-bs-toggle="modal" data-bs-target="#implementation-{{ loop.index }}-runtime-info">
+                      more info
```
Julian, line 195: Let's maybe make this even more conspicuous with one of those [i icons](https://icons.getbootstrap.com/icons/info-circle-fill/). Make sense?

bowtie/templates/report.html.j2
```diff
@@ -363,6 +367,29 @@
...
+        <div class="modal-dialog" role="document">
+          <div class="modal-content">
+            <div class="modal-header">
+              <h5 class="modal-title" id="implementation-{{ loop.index }}-runtime-info-label">
+                <b>{{ implementation.name }} </b> <small class="text-muted"> {{ implementation.language }}</small>
+              </h5>
+              <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            <div class="modal-body">
+                <p><strong>OS Version : </strong>{{ implementation.get("os_version", "") }}</p>
```
Julian, line 382: There are extra spaces in front of the `:`'s here, can you trim those.

Also maybe let's only show values where they're non-empty (so hide any row where we don't have a value for that key).

Julian (changes requested):
No worries at all! Left you two minor comments which maybe you can address, but think this seems pretty good to me! Let me know what you think.

bowtie/templates/report.html.j2
AgniveshChaubey: Sure!

bowtie/templates/report.html.j2
AgniveshChaubey: Extra white space is removed!
Conditionally rendering the row if its value is non-empty

Julian:
Nice! Looks great, thanks again @AgniveshChaubey!

PR #202, opened 2023-04-04
badges: generant pass pct json files

santhosh-tekuri:


santhosh-tekuri:
related issue #55

santhosh-tekuri:
@Julian plz do codereview

santhosh-tekuri:
@Julian yes. I was aware of the work in progress by other PR

since there is no process in PR for week, I jumped in with fresh PR

Julian:
Newer programmers need more time to work on things. Even for a more seasoned developer it's polite to first offer help before immediately jumping in. Please ask next time.

Julian:
Thanks! Merged this after making a few minor tweaks (mainly using `pathlib.Path` objects) which you can see on `main` if you're curious. Otherwise this seems like a start definitely, thanks again.

I guess I was assuming all badges would be produced in one place (if not right now where we have multiple runs, at least eventually). But we can discuss that on the other PR.

Also I guess the badges here should probably be named something to do with their function -- so `badges/implementation/passing/draft7.json` or whatever -- but we can make that tweak once we have a second badge as well. Let me know if you have opinions there.

santhosh-tekuri:
all badges are generated in the dir mentioned in cli argument

Julian:
Right but with #203 someone generating all badges needs to know to run multiple different commands, yes? Or similarly the CI job now needs to run `bowtie info` as well, which it didn't before.

santhosh-tekuri:
@Julian may be we can generate badges both in `info` and `report`;
in CI, we will just use `report` command to generate all badges

Julian:
That's fine with me too yeah.

PR #207, opened 2023-04-05
images: publish must always do fresh build

santhosh-tekuri:


santhosh-tekuri:
@Julian with this PR:
- always build all images when publishing a release
- only build modified images on pull request

Julian:
Nice! Did another release to see that piece.

PR #208, opened 2023-04-05
testing image build

santhosh-tekuri:


santhosh-tekuri:
working as expected; closing

Julian:
Yeah, seems to have done the right thing, both for this and for the release, thanks for the fix.

Though I wish that action worked by not even showing the jobs that are skipped rather than just skipping them, it makes it hard to see which ones ran.

Maybe as a follow up at some point we should emit a GitHub comment onto the PR saying which images were rebuilt.

PR #221, opened 2023-04-11
Revamped the navbar added dropdown and support for mobile menu

nomandhoni-cs:
@Julian please have a look.
# Before
Screenshot_88

# After
Container Fluid

# Changing Reasons
In near future if we have more Dialects than now, we will not have enough space.

# Changes
- Made minor changes to the Navbar by adding a dropdown menu.
- Added JQuery to enable the Bootstrap dropdown menu feature.
- Mobile menu is now fully functional after the addition of JQuery.
- ( In github 'Files changed' section of this PR looks a little bit messy because, I removed unwanted spaces, gap for more readability )


Julian:
The idea looks nice. Thanks! Some notes:

* I don't think jQuery is needed for dropdowns in bootstrap 5.x is it? That at least isn't listed on https://getbootstrap.com/docs/5.3/components/dropdowns/ and I *believe* at one point I played around with adding a dropdown and just followed those instructions
* Can you reverse-sort the dialects? In other words, 2020-12 is the newest dialect, so it's nice to have it on top, and then draft 3 is very old so it's fine to have it on the bottom
* As you mentioned, it's quite hard to review this because it looks like you've reformatted the file. In general *definitely* do that in separate commits, if not separate PRs entirely.

nomandhoni-cs:
Thanks a lot @Julian for guiding me.
- I didn't realized that reformatting the file will blow up the 'Files Changed' section. I should have check before commit by
 `git diff`
- This is gonna be lesson for me, I should have more careful before commit. I hope this same mistake will not happen again.
- When I was testing the dropdown menu, I forgot to remove the old cdn, that's why (old and new cdn was conflicting with each other ) the dropdown wasn't working. 
- To solve that I included JQuery in the last commit.
- I took a look in the code again and removed that old cdn and JQuery . Replaced those with new Bootstrap cdn (Latest)
- Reorganized the dropdown links
## Now everything works as expected
reorganized

bowtie/templates/report.html.j2
```diff
@@ -3,424 +3,437 @@
...
-                <a class="nav-link" href="draft3.html">3</a>
-              </li>
-            {% endif %}
-          </ul>
-        </div>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Bowtie</title>
+  <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/css/bootstrap.min.css" rel="stylesheet">
```
Julian, line 56: You're changing (updating) the version here, and specifically to an alpha version -- I happen to be OK with that if it's necessary for whatever reason (even to the alpha version which would normally not be OK) but when submitting a PR you should definitely call that out specifically -- that you're updating the version -- and tell me the reviewer why/what made you do it. If it's not related to functionality in this PR then doing this change is similar to the formatting change in that it might be good to do but normally you shouldn't do it at the same time as the other changes, it makes things hard to review and hard to debug.

But yeah all that explanation aside, is there a reason it's needed?

Secondly, and also importantly, you dropped the `crossorigin="anonymous"` and `integrity` attributes both here and at the bottom in the other script tag. I have to assume the `integrity` change at least was a mistake, since there's no reason *not* to include it. The `crossorigin="anonymous"` attribute probably does no harm nor no benefit for this kind of thing, but still for both of those the same advice as above applies -- you should be super vocal about removing them if you did it on purpose, and if not, it's a good lesson to make sure you review the `git diff` carefully every time you submit a PR or change -- you may have had difficulty yourself doing so for the same reason a reviewer does/would because of the formatting changes, but still!

Julian:
> This is gonna be lesson for me, I should have more careful before commit. I hope this same mistake will not happen again.

All good! Yeah, it's good to be mindful of but nothing earth shattering!

The changes look better. Left a final comment on the changes.

bowtie/templates/report.html.j2
```diff
@@ -3,424 +3,437 @@
...
+            <a class="nav-link" href="#cases">Details</a>
+          </li>
+          {# If generate_dialect_navigation exists, then render the dropdown navigation  #}
+          {% if generate_dialect_navigation %}
+          <li class="nav-item dropdown">
+            <a class="nav-link dropdown-toggle" href="#" role="button" data-bs-toggle="dropdown" aria-expanded="false">
+              Dialects
+            </a>
+            <ul class="dropdown-menu">
+              <li><a class="dropdown-item" href="draft2020-12.html">2020-12</a></li>
```
Julian, line 85: Also while we're here fixing things, let's maybe use URLs without `.html` at the end, i.e. `href="draft7"` rather than `draft7.html`, etc.

bowtie/templates/report.html.j2
nomandhoni-cs: ## Why I changed cdn to alpha?
I was thinking about how to implement dark mode and light mode in this site with custom JavaScript and CSS,
And When going through the Bootstrap docs, found in v5.3.0 they are added this functionality. That I don't have to write custom css and js for these. I just simply use bootstrap class for this.
- Gained some trust, when saw some of their ( premium themes ) [Bootstrap Premium Themes]( https://themes.getbootstrap.com/) using these and in their home page the are showing the alpha version (not in corner or in small text).
Screenshot_89

nomandhoni-cs: ## Why I dropped crossorigin and integrity ?
I did not dropped those on purpose. I was doing those old fashioned way ( How I used to do before )
cdns
Copy the CDNs from here and paste in the href of <script> tag and in <link> tag for html and css.

After this comment I did some googling and found out
If you do not use the integrity attribute, it is possible for the file to be modified or replaced, which could cause security issues or break your website's functionality.

bowtie/templates/report.html.j2
nomandhoni-cs: Finally, Should I make another commit 
- For adding crossorigin, integrity back again.
- What about keeping the same CDN?
- Or switch to the v5.2.0 and wait until v.5.3.0 get stable then start working on light mode and dark mode themes for the site?

Julian: If it's not necessary for this PR keep the version exactly as-is without changing and you can bump it in a PR for dark mode of course!

Julian:
Great, thanks, looks great, appreciated!

nomandhoni-cs:
Sure! Highlighting the current dialect, where currently user is visiting. I will do that in next commit / pr

Issue #31, opened 2022-09-13
Toggleable / intelligent dark mode in the report


Julian:
Probably once https://github.com/twbs/bootstrap/pull/35857 is merged.

AgniveshChaubey:
Hi @Julian, I have implemented [dark mode](https://github.com/bowtie-json-schema/bowtie/pull/225) feature in the Bowtie report. Please have a look and let me know if any change is required.

PR #225, opened 2023-04-12
Add dark mode to bowtie report

AgniveshChaubey:
### This PR refers to #31 

Added toggleable dark mode feature to Bowtie report using Bootstrap.


bowtie/templates/report.html.j2
```diff
@@ -93,6 +102,15 @@
           {% endif %}
         </ul>
       </div>
+        <div id="theme-toggler" class="" style="margin-right: 1vw;">
+          <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="currentColor" class="bi bi-sun" viewBox="0 0 16 16">
```
Julian, line 106: I think this might need/want to be wrapped with a `button` element -- otherwise when you mouseover it the cursor doesn't change (from a pointer to a hand, say).

I noticed the same thing to be true about the `i` icons we added for runtime info, so if you manage to fix this here perhaps you can fix that one as well the same way.

bowtie/templates/report.html.j2
```diff
@@ -417,6 +435,53 @@
...
+      })
+    </script>
+
+    <script
+    src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/js/bootstrap.bundle.min.js"
+    integrity="sha384-ENjdO4Dr2bkBIFxQpeoTz1HIcje39Wm4jDKdf19U8gI4ddQ3GYNS7NTKfAdVQSZe"
+    crossorigin="anonymous"
+    ></script>
+   <script>
+      const tooltipTriggerList = document.querySelectorAll('[data-bs-toggle="tooltip"]')
```
Julian, line 481: This combined with the line above redeclaring it breaks collapsing the accordion items I think.

Try running some test run (e.g. `bowtie suite -i python-jsonschema $DEVELOPMENT/JSON-Schema-Test-Suite/tests/draft7/ | bowtie report`) and then expand one of the tests and try to collapse it, it won't collapse again for me and I see:

```
Uncaught SyntaxError: redeclaration of const tooltipTriggerList
    <anonymous> file:///Users/julian/Development/bowtie/bowtie-report.html:34509
```

in the console.

Julian (changes requested):
Thanks! Looks nice to start.

Some comments left.

Also, there's a [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media feature which I think it'd be nice to support -- it I think signals whether the user prefers light or dark, and we should use that to decide which version to show initially by default.

E.g. my browser I think is set to prefer dark, but when I open a report it still starts in light.

Can you perhaps give a shot at whether we can easily support that (I'm not sure if Bootstrap's dark mode support added support for it itself and we just need to enable it, or if there's something we'd need to do ourselves, needs looking into).

bowtie/templates/report.html.j2
AgniveshChaubey: accordian collapse issue resolved

bowtie/templates/report.html.j2
AgniveshChaubey: wrapped the icons (both **info** and **theme-toggler**) into `button` elements

bowtie/templates/report.html.j2
```diff
@@ -47,17 +47,26 @@
     color: rgb(112, 112, 112);
     font-size: 0.8rem;
   }
+  .hidden {
```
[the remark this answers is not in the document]
AgniveshChaubey: replaced the custom **hidden** class with bootstrap's inbuilt **d-none** class

AgniveshChaubey:
> Some comments left.
> 
> Also, there's a [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media feature which I think it'd be nice to support

Addressed all requested changes and now working on adding **smart dark mode** based on the user's default preference.

AgniveshChaubey:
Hi @Julian, I have implemented ***intelligent dark-mode*** feature using the 'prefers-color-scheme' CSS media query. One thing I want to ask-- Although all modern browsers support the `matchMedia()` method, it may happen that some older browsers do not. In that case, there will be an error. Should I consider addressing it or just ignore it because there's very less chance the browser does not support it?

bowtie/templates/report.html.j2
```diff
@@ -411,12 +426,47 @@
...
   {% endfor %}
-  <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.2/dist/js/bootstrap.bundle.min.js" integrity="sha384-OERcA2EqjJCMA+/3y+gxIOqMEjwtxJY7qPCqsdltbNJuaOe923+mo//f6V8Qbsw3" crossorigin="anonymous"></script>
+  <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/js/bootstrap.bundle.min.js" integrity="sha384-ENjdO4Dr2bkBIFxQpeoTz1HIcje39Wm4jDKdf19U8gI4ddQ3GYNS7NTKfAdVQSZe" crossorigin="anonymous"></script>
   <script>
     const tooltipTriggerList = document.querySelectorAll('[data-bs-toggle="tooltip"]')
     const tooltipList = [...tooltipTriggerList].map(tooltipTriggerEl => new bootstrap.Tooltip(tooltipTriggerEl))
   </script>
 
+  <script>
+    if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
```
Julian, line 436: This is basically

```suggestion
    var light = !window.matchMedia('(prefers-color-scheme: dark)').matches
```

(or invert the condition to `prefersDark` and you don't need the `!`).

Julian:
Great! I don't think we care about older browsers so that's all fine.

This looks nice I think. I'd love to find a way to remove duplication, it seems like there's a high likelihood someone will forget to update the block you added if there's a new element which needs to swap from light to dark.

But we can maybe handle that as a follow-up. For now I'll merge I think. Thanks!

AgniveshChaubey:
Thank you for the positive feedback! I understand your concern about duplication, and I will definitely try to find a solution to incorporate it in the future. For now, I'm glad to hear that the changes got merged! :)

PR #229, opened 2023-04-13
Implemented Schema Display and removed tooltip, Added SVG Icon name before them

nomandhoni-cs:
#### Implemented Schema Display and removed old way of displaying by tooltip, 
- This will not increase the size of generated html because of displaying single div dynamically
#### Changes
- Used a separate div for display the schema
- When a clicked on the accordion, I insert the same schema display div in the accordion dynamically
- You can one display for all. That's why this will not increase the size of the generated html file
#### Demo Video
https://user-images.githubusercontent.com/92979541/231662325-687fbbcc-cc79-4d87-9027-fa467def8a11.mp4

#### Why I Did this ?
When going through these svg, I have to find out the class and based on the class I have guess this icon is look like this.
This is a kind of hassle. That's why added the Icon name before the icon by comment.

### DIFF
--- a/bowtie/templates/report.html.j2
+++ b/bowtie/templates/report.html.j2
@@ -282,7 +282,7 @@
               {% for test, test_results in results %}
               <tr>
                 <td>
-                  <a href="#schema" class="text-decoration-none" onclick='displayCode(`{{ test.instance | tojson(indent=2) }}`)'>{{ test.description }}</a>
+                  <a href="#schema" class="text-decoration-none" onclick='displayCode(`{{ test.instance | tojson(indent=2) }}`,"instance-info")'>{{ test.description }}</a>
                 </td>
                 {% for implementation in summary.implementations %}
                 {% set implementation_result, incorrect = test_results.get(implementation.image, ({"valid": None},
@@ -420,22 +420,13 @@
           <div class="d-flex align-items-center highlight-toolbar ps-3 pe-2 py-1 border-0 border-top border-bottom">
             <small class="font-monospace text-body-secondary text-uppercase">Schema</small>
             <div class="d-flex ms-auto">
-              <button type="button" id="copy-button" class="btn mt-0 me-0" onclick="copySchemaFromDisplay()" aria-label="Copy to clipboard" data-bs-toggle="tooltip" data-bs-placement="top"
+              <button type="button" id="copy-button-schema" class="btn mt-0 me-0" onclick='copyToClipboard("schema-code","copy-button-schema")' aria-label="Copy to clipboard" data-bs-toggle="tooltip" data-bs-placement="top"
               data-bs-custom-class="custom-tooltip"
               data-bs-title="Copy to clipboard">
               {# Clipboard SVG Icon  #}
                 <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-clipboard" viewBox="0 0 16 16">
                   <path d="M4 1.5H3a2 2 0 0 0-2 2V14a2 2 0 0 0 2 2h10a2 2 0 0 0 2-2V3.5a2 2 0 0 0-2-2h-1v1h1a1 1 0 0 1 1 1V14a1 1 0 0 1-1 1H3a1 1 0 0 1-1-1V3.5a1 1 0 0 1 1-1h1v-1z"/>
                   <path d="M9.5 1a.5.5 0 0 1 .5.5v1a.5.5 0 0 1-.5.5h-3a.5.5 0 0 1-.5-.5v-1a.5.5 0 0 1 .5-.5h3zm-3-1A1.5 1.5 0 0 0 5 1.5v1A1.5 1.5 0 0 0 6.5 4h3A1.5 1.5 0 0 0 11 2.5v-1A1.5 1.5 0 0 0 9.5 0h-3z"/>
-                </svg>
-              </button>
-              <button type="button" class="btn text-nowrap" onclick="closeSchemaCard()" aria-label="Close" data-bs-toggle="tooltip" data-bs-placement="top"
-              data-bs-custom-class="custom-tooltip"
-              data-bs-title="Close">
-              {# Close SVG Icon  #}
-                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-x-square" viewBox="0 0 16 16">
-                  <path d="M14 1a1 1 0 0 1 1 1v12a1 1 0 0 1-1 1H2a1 1 0 0 1-1-1V2a1 1 0 0 1 1-1h12zM2 0a2 2 0 0 0-2 2v12a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V2a2 2 0 0 0-2-2H2z"/>
-                  <path d="M4.646 4.646a.5.5 0 0 1 .708 0L8 7.293l2.646-2.647a.5.5 0 0 1 .708.708L8.707 8l2.647 2.646a.5.5 0 0 1-.708.708L8 8.707l-2.646 2.647a.5.5 0 0 1-.708-.708L7.293 8 4.646 5.354a.5.5 0 0 1 0-.708z"/>
                 </svg>
               </button>
             </div>
@@ -446,7 +437,7 @@
             <div class="d-flex align-items-center highlight-toolbar ps-3 pe-2 py-1 border-0 border-top border-bottom">
               <small class="font-monospace text-body-secondary text-uppercase">Instance</small>
                 <div class="d-flex ms-auto">
-                  <button type="button" id="copy-button" class="btn mt-0 me-0" aria-label="Copy to clipboard" data-bs-toggle="tooltip" data-bs-placement="top"
+                  <button type="button" id="copy-button-instance" onclick='copyToClipboard("instance-info","copy-button-instance")' class="btn mt-0 me-0" aria-label="Copy to clipboard" data-bs-toggle="tooltip" data-bs-placement="top"
                   data-bs-custom-class="custom-tooltip"
                   data-bs-title="Copy to clipboard">
                   {# Clipboard SVG Icon  #}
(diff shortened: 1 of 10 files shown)


Julian:
Nice. Thanks! I'll have a closer look and leave some more specific comments, but at first glance the most obvious thing that stands out is that you're inserting *instances* into the box for schemas.

E.g. in your video, `{"type": "integer"}` is a schema -- it's the schema for all the examples you're clicking on.

37 (or "37" or etc.) are not schemas, they don't belong in the box labelled schema -- they're *instances*, data that the schema is validating. So someone wants to see both of those at the same time (the schema, and the instance that's valid or invalid under it)

We need to think about where the best place to render the instances is if we're going to do that dynamically -- maybe it makes sense to split your schema box into 2 columns, put the schema on the left, and dynamically change the instance on the right?

bowtie/templates/report.html.j2
```diff
@@ -47,6 +50,10 @@
     color: rgb(112, 112, 112);
     font-size: 0.8rem;
   }
+  {# Schema diplay hidden default  #}
+  #display-body{
```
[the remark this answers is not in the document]
nomandhoni-cs: Sure! I missed that bootstrap class.
Thanks

nomandhoni-cs:
That's great, thanks for correcting me. 
Let's see which one looks good, showing instances in a column or in a separate row below. With title 'Instance' above.

Julian:
> There are some schemas, too wide, I think they're gonna take whole space.

I think we'll find there are both schemas *and* instances which are going to be too wide, so no matter what we'll need to handle that.

But I'm ok seeing how it looks with a horizontal split which it sounds like is what you're suggesting to start -- so schema on top and last-selected instance on bottom? Probably it's nice to highlight the row of the instance that's shown if we do that.

bowtie/templates/report.html.j2
```diff
@@ -2,6 +2,9 @@
 <html lang="en">
 
 <style>
+  a:link{
+    text-decoration: none;
```
Julian, line 6: Here too -- can you elaborate on which link you're adding this for? I would also suspect Bootstrap should have some class for this though I didn't check.

bowtie/templates/report.html.j2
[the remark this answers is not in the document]
Julian: You resolved the comment but this looks like it's still in the diff, no?

bowtie/templates/report.html.j2
```diff
@@ -410,13 +419,93 @@
...
+        displayBody.classList.remove("d-none");
+        displayBody.classList.add("d-block");
+        codes.innerHTML = `<pre>${code}</pre>`;
+      }
+      function schemaDisplay(parentBodyId, code) {
+        displayBody.classList.remove("d-none");
+        displayBody.classList.add("d-block");
+        displayCode(code);
+        const isDisplayed = displayBody.parentElement === document.getElementById(parentBodyId);
+        // If the code segment is displayed, remove it
```
Julian, line 470: A lot of the comments above are fine/definitely improvements for understanding SVG icons, but these kind of comments seem less so, they just repeat what some fairly simple code seems to be doing, or repeat the name of a function.

bowtie/templates/report.html.j2
```diff
@@ -410,13 +419,93 @@
...
+        document.execCommand("copy");
+        document.body.removeChild(tempElement);
+        {# Change the Copy Button icon after copied #}
+        {# Checkmark SVG Icon  #}
+        document.getElementById("copy-button").innerHTML = `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-check" viewBox="0 0 16 16">
+                                                              <path d="M10.97 4.97a.75.75 0 0 1 1.07 1.05l-3.99 4.99a.75.75 0 0 1-1.08.02L4.324 8.384a.75.75 0 1 1 1.06-1.06l2.094 2.093 3.473-4.425a.267.267 0 0 1 .02-.022z"/>
+                                                            </svg>`;
+setTimeout(() => {
+  {# Clipboard SVG Icon setting as a default again after copy  #}
+  document.getElementById("copy-button").innerHTML = `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-clipboard" viewBox="0 0 16 16">
```
Julian, line 501: I think rather than continuing to copy paste SVG elements, let's just use [Bootstrap Icons](https://icons.getbootstrap.com/#icon-font) at this point? Which I think already has most of the SVGs both that were already present as well as the new ones you're adding.

nomandhoni-cs: In the html report, every SVG icon is from bootstrap. Yes we can Directly use CDN for icons and, Instead of using the svg code we will use icons like `<i class="bi bi-0-circle"></i>` this.

bowtie/templates/report.html.j2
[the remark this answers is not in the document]
nomandhoni-cs: Opps! There was a mistake. First Time I made a new commit with every thing but I forgot to git pull before commiting new changes. Then I had to stash that commit and rewrite these. In that time I forgot to remove this css.

bowtie/templates/report.html.j2
nomandhoni-cs: Currently There are no need to use `<a href="#">` for these links schema and test name, I just wanted to remove default link style (Underline)
Screenshot_90

bowtie/templates/report.html.j2
Julian: Ok let's decide in a follow-up and stick as-is for now.

bowtie/templates/report.html.j2
nomandhoni-cs: Done! Used bootstrap class instead of hard coded css

bowtie/templates/report.html.j2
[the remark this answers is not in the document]
nomandhoni-cs: Done! Used Bootstrap class ( Thanks for correcting me )

nomandhoni-cs:
Hey @Julian,
I think this pull request is ready to be merged!
- Solved my silly mistakes ( Using hard coded css instead of using bootstrap class )
- Implemented display schema and instance info display in the same row
- Functional copy button
- Removed close button ( I Will add this later, currently thinking what is the best way to do this with less code )
Screenshot_92
 

https://user-images.githubusercontent.com/92979541/232242189-71d1a979-a212-40f0-854e-8baa4ef77c32.mp4

## Functionality I will bring in next PR
- Highlight the clicked instance
- Close button
- Find out more enhancement

bowtie/templates/report.html.j2
nomandhoni-cs: Removed some of the comments

bowtie/templates/report.html.j2
```diff
@@ -410,13 +413,100 @@
...
+      }
+      {# Copy Schema from display  #}
+      function copyToClipboard(targetTextToCopy, btnId){
+        const textToCopy = document.getElementById(targetTextToCopy).innerText;
+        const tempElement = document.createElement("textarea");
+        tempElement.value = textToCopy;
+        document.body.appendChild(tempElement);
+        tempElement.select();
+        tempElement.setSelectionRange(0, 99999);
+        document.execCommand("copy");
```
Julian, line 493: I think this is the old API for doing this -- on not-super-old browsers I think you can just use [the clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API).

bowtie/templates/report.html.j2
```diff
@@ -278,8 +282,7 @@
               {% for test, test_results in results %}
               <tr>
                 <td>
-                  <a href="#" data-bs-toggle="tooltip" data-bs-html="true"
-                    data-bs-title='<pre>{{ test.instance | tojson(indent=2) }}</pre>'>{{ test.description }}</a>
+                  <a href="#schema" class="text-decoration-none" onclick='displayCode(`{{ test.instance | tojson(indent=2) }}`,"instance-info")'>{{ test.description }}</a>
```
Julian, line 285: I think this `onClick` should go higher -- right now it only renders the instance if you click on the text, but someone is likely to click on the *row*, even beyond the text. So maybe on the `tr`? Until/unless we have some way of displaying the actual results more clearly.

Julian:
Thanks Noman! Left a few comments directly on the PR. Two others:

* Something's broken when collapsing accordion items -- see e.g.:

https://user-images.githubusercontent.com/329822/232501371-edab2165-2b46-40b5-9ab6-1c2b80006efe.mov

where after collapsing one, reopening it doesn't render the schema and instance anymore.

* Can you rebase or merge the main branch into yours, it bumps the bootstrap version so let's just be sure it still works after your changes

nomandhoni-cs:
Thanks, I didn't think of that testing, Closing, and opening the same accordion. I the I have to save the state with boolean and based on that I have to execute the function.

nomandhoni-cs:
Thanks a lot @Julian,
There was this little mistake, that I forgot to handle the edge case, that if I click on the same accordion for open and close.
- Now handled that with condition to see if the accordion already open or not
- Based on that execute the function
https://user-images.githubusercontent.com/92979541/232547118-013ec25b-e25a-45cd-a88a-e2baebf980ea.mp4

Julian:
Great -- I see you have a few more minor style issues -- I just added a linter to try and catch some, though I'm not sure it'll catch all of them yet. But can you again merge main into your branch and see if you can fix some of the remaining spacing / style issues?

nomandhoni-cs:
Merged from my main branch to fix spacing / style issue as you told me to do.

bowtie/templates/report.html.j2
```diff
@@ -1,6 +1,54 @@
 <!DOCTYPE html>
 <html lang="en">
 
+<style>
+  .details-required {
```
Julian, line 5: This might be a merge mistake -- the style tag belongs in the head tag if we're going to have it -- though a few more things below look like they should be doable with normal bootstrap classes -- but can you move all this into the style tag in the head tag?

nomandhoni-cs: and Done!
I cannot express my gratitude enough for your incredible patience throughout this journey. Your unwavering support and guidance have been invaluable to me. Your help in identifying and correcting my mistakes has been particularly meaningful. Thank you so much for everything.

Julian: No worries at all, definitely not a race! Thanks for the efforts!

Julian:
Looks nice! Merging, really appreciated! Will deploy in just a bit.

PR #231, opened 2023-04-17
Displays JSON format for `bowtie info`

### LINKED ISSUE #188: Add --format json to bowtie info
Like `bowtie summary`, `bowtie info` could benefit from a `--format json` option to make its output more easily machine-processable.

dante381:
Hey @Julian,
Fixes #188 ,
Added the format flag to the info function and created a dictionary that stores the metadata of the implementations.

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -510,7 +510,7 @@ async def _info(image_names: list[str], format: str):
             else:
                 click.echo(
                     "\n".join(
-                        f"{k}:{json.dumps(v, indent=2)}"
+                        f"{k}: {json.dumps(v, indent=2)}"
                         for k, v in metadata.items()
                     ),
                 )
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -450,6 +450,14 @@ def validate(
 @main.command()
 @click.pass_context
 @IMPLEMENTATION
+@click.option(
+    "--format",
+    "-f",
+    help="What format to use for the output",
+    default=lambda: "pretty" if sys.stdout.isatty() else "json",
+    show_default="pretty if stdout is a tty, otherwise JSON",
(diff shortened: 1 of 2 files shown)


Julian:
Nice. Remember though, we don't generally want to *remove* existing tests, which you're doing by no longer testing the old behavior. You want to add a new one. You might need to modify the old one slightly, but otherwise the new behavior is essentially self-contained and can live in a new test.

bowtie/_cli.py
```diff
@@ -484,21 +492,37 @@ async def _info(image_names: list[str]):
...
+                        kv[0] == "dialects",
+                        kv[0],
+                    ),
+                )
+            }
+            if format == "json":
+                click.echo(json.dumps(metadata, indent=2))
+            else:
+                click.echo(
+                    "\n".join(
```
Julian, line 512: I think you also can make do without calculating this a second time here, you should just be able to at least partially reuse `metadata` here without recalculating it from` implementation.metadata`. I.e. something like:

`"\n".join(f"{k}: {json.dumps(v, indent=2)}" for k, v in metadata.items()}` probably should do it though I haven't double checked.

tests/test_integration.py
```diff
@@ -445,10 +447,10 @@ async def test_info(envsonschema):
     assert proc.returncode == 0, (stdout, stderr)
     assert stdout.decode() == dedent(
         """\
-        name: "envsonschema"
-        language: "python"
-        issues: "https://github.com/bowtie-json-schema/bowtie/issues"
-        dialects: [
+        name:"envsonschema"
```
Julian, line 450: The fact you had to modify this should have pointed out you're now missing a space :) -- use the tests to find bugs! That's what they're there for.

Julian:
Great! Thanks! Looks great!

PR #233, opened 2023-04-19
Removed $ sign before cli

nomandhoni-cs:
## Why this is needed
- Every time when I copy a line from copy button this include a $ and space before the line
Screenshot_94
- And then I have manually move my cursor to remove that
- I thought, let's remove the $ sign before that line, and make it developer friendly

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -12,7 +12,7 @@ To run the draft 7 ``type``-keyword tests on the Lua ``jsonschema`` implementati
 
 .. code:: sh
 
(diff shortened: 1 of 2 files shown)


Julian:
The better way to solve this is to actually tweak the [sphinx_copybutton settings](https://sphinx-copybutton.readthedocs.io/en/latest/use.html). I forget which one controls this (in fact I forgot that it doesn't do it by default). But perhaps you want to try doing that?

The prompt characters are pretty common to show, and indicate to someone that you're intended to run these at a shell.

I definitely think it's reasonable to make sure that when someone copies the line they don't get them copied though -- have a look at the doc above and let me know if you can't see which one is the right setting.

Julian:
The tweak I mentioned is done in fed6de8.

PR #234, opened 2023-04-19
Highlight clicked Instance row for better User Experience

nomandhoni-cs:
https://user-images.githubusercontent.com/92979541/233057401-581b075c-6887-4273-9d9d-9492d264aea4.mp4

## Features
- Highlight the clicked instance row
- Did this because, If user scroll and forgot which one they clicked to see instance info. Now he / she can see that highlighted row

### DIFF
--- a/bowtie/templates/report.html.j2
+++ b/bowtie/templates/report.html.j2
@@ -476,7 +476,6 @@
 
   <script>
       const displayBody = document.querySelector("#display-body");
-      const codes = document.querySelector("#schema-code");
       let currentParrentBodyId = '';
       {# Highlight the clicked row   #}
       function rowHighlightListener(rowClass) {
@@ -519,9 +518,11 @@
         }
       }
 
+      const clipBoardIcon = document.querySelector('.bi-clipboard');
       {# Copy Schema from display #}
       function copyToClipboard(targetTextToCopy, btnId) {
         const textToCopy = document.getElementById(targetTextToCopy).innerText;
+        console.log(clipBoardIcon);
         navigator.clipboard.writeText(textToCopy)
         {# Change the Copy Button icon after copying #}
         {# Checkmark SVG Icon #}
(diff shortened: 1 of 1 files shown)


bowtie/templates/report.html.j2
```diff
@@ -496,16 +509,20 @@
...
             accordionBody.insertBefore(displayBody, accordionBody.firstChild);
           }
         }
       }
 
+      const clipBoardIcon = document.querySelector('.bi-clipboard');
       {# Copy Schema from display #}
       function copyToClipboard(targetTextToCopy, btnId) {
         const textToCopy = document.getElementById(targetTextToCopy).innerText;
+        console.log(clipBoardIcon);
```
Julian, line 525: I'm surprised the linter still doesn't catch this, but you have a stray `console.log` statement left in.

bowtie/templates/report.html.j2
```diff
@@ -170,7 +170,7 @@
 
       <div class="card-body">
         <table class="table table-sm table-hover">
-          <thead class="table-header table-light">
+          <thead>
```
Julian, line 173: Are you sure the `table-header` class also should be removed?

nomandhoni-cs: Yes, 100% sure, this was used for get this specific table header in JS to implement dark mode for a single table header, But dark mode should be applicable for all tables across whole site. That's why removed this and used dark mode and light mode in whole table not in a single table header

bowtie/templates/report.html.j2
nomandhoni-cs: Opps! Silly mistake happened!

nomandhoni-cs:
## Before Table dark mode
https://user-images.githubusercontent.com/92979541/233086840-486ad5a2-ea08-44e0-9e19-dc380e7ab41d.mp4

## After table dark mode

https://user-images.githubusercontent.com/92979541/233086978-ae4e4f8f-1462-4adf-9985-db5a186b0e29.mp4

Julian:
Cool, looks great now I think, thanks!

Issue #157, opened 2023-03-07
Display adherence of implementations by keyword or vocabulary


jviotti:
So we could link to this from https://www.learnjsonschema.com

jviotti:
No worries! https://www.learnjsonschema.com provides documentation of specific keywords, or specific vocabularies. 

Now imagine that Bowtie has pages, on its website, for checking implementation adherence of specific keywords or vocabularies. 

For example, I go to the Bowtie website, select the `unevaluatedProperties` keyword from the Unevaluated vocabulary and can see which JSON Schema implementations properly support this keyword and which ones don't very well.

As a vocabulary example, imagine I go to the Bowtie website, select the Validation vocabulary, and can see which JSON Schema implementations support this vocabulary well and which ones don't.

If Bowtie had these pages, I could directly link to them from https://www.learnjsonschema.com.

AgniveshChaubey:
Does this mean that we will need to create a separate subpage for each vocabulary/keyword? If so, won't it lead to a large number of files in the local root directory? If we do have to create these subpages, I believe creating a subfolder would be a better option for storing those files. This is what I think you are referring to. Please let me know if I am going in the right direction.

jviotti:
I don't know how this project is structured, but as long as there is a different URL for each keyword and vocabulary, that'd do the trick.

AgniveshChaubey:
Ah, I see! So you're looking for individual URLs for each keyword and vocabulary. I have started exploring possible ways to make that happen and will bring up some solutions shortly!

(And sorry for the delayed replies - actually, submission week is going on, and most of my time is being eaten up by writing assignments.)

jviotti:
Exactly! No worries. Good luck!

AgniveshChaubey:
> (I think one or two of you probably even gave doing so a shot and ran into trouble trying to make that happen). 

Actually, it was me only! :)

> Yesterday I got a moment to do that work myself and try and stimulate a bit there without disrupting the existing report...

Can you share more about how you did this? (I was facing problem in converting the Python dict. to json string)

> Specifically, now that we have the raw data available separately, a page that queries it and dynamically renders stuff by keyword or vocabulary is a lot more feasible.

Do you mean to suggest creating an entirely separate page, independent of the Bowtie report page, that can be dynamically rendered based on the query? Or do you think it would be more practical to add a new section to the existing Bowtie report that can be dynamically rendered based on the query?

Julian:
[...]

Issue #42, opened 2022-09-29
Investigate autogenerating Python classes from the schema


Julian:
Right now we [hand-write classes corresponding to the types in the `io-schema.json` file](https://github.com/python-jsonschema/bowtie/blob/main/bowtie/_commands.py).

It'd be nice to pick a library (or write one) which does so for us and generates attrs-using classes.

PR #241, opened 2023-04-21
Draft PR : Migrating Bowtie Report

AgniveshChaubey:
As per our discussion in [this](https://github.com/bowtie-json-schema/bowtie/issues/157) ticket, I have commenced the task of migrating the report that is currently being developed using JavaScript.

However, I encountered an issue when attempting to decompress the JSON file, specifically draft7.json.gz. The resulting output comprised various objects that did not conform to the proper JSON format. (It's not even an array of objects)
Something like...
```
{key: value}
{key: value}
{key: value}
{key: value}
```

I wonder how to retrieve data from this file! Nevertheless, I have started the process of migrating the report by utilizing a dummy JSON file that shares similarities with the report's data file.

As of now, 30-40% of the work has been completed, and the remaining tasks are still in progress. (We only need to make slight adjustments to the variables to correctly render the data from the original report).

This draft PR is intended to demonstrate the progress made and to confirm that we are moving in the right direction. (this is just a temporary file)

### DIFF
--- a/bowtie/templates/data.json
+++ b/bowtie/templates/data.json
@@ -1,0 +1,4 @@
+{
+  "dialect": "Draft7",
+  "version": "0.82.01",
+}
--- a/bowtie/templates/data.json
+++ b/bowtie/templates/data.json
@@ -1,6 +1,23 @@
 {
-  "dialect": "Draft7",
+  "run_info": {
+    "dialect_shortname": "Draft7",
+    "bowtie_version": "0.82.01",
+    "dialect": "http://json-schema.org/draft-07/schema#"
+  },
+  "summary": {
+    "total_cases": 67,
+    "total_tests": 33,
+    "errored_cases": 22,
+    "skipped_tests": 17,
+    "implementation": [
+      { "name": "js-hyperjump", "language": "javascript", "version": "4.2.5" },
+      { "name": "python-jschon", "language": "python", "version": "2.8.1" }
+    ],
+    "counts": {
+      "errored_cases": 30,
+      "skipped_tests": 11
+    }
+  },
   "dialectLink": "http://json-schema.org/draft-07/schema#",
-  "version": "0.82.01",
-  "runStarted":"2023-04-21T09:30:00"
+  "runStarted": "2023-04-21T09:30:00"
 }
--- a/bowtie/templates/data.json
+++ b/bowtie/templates/data.json
@@ -1,4 +1,6 @@
 {
   "dialect": "Draft7",
-  "version": "0.82.01"
+  "dialectLink": "http://json-schema.org/draft-07/schema#",
+  "version": "0.82.01",
+  "runStarted":"2023-04-21T09:30:00"
 }
--- a/bowtie/templates/data.json
+++ b/bowtie/templates/data.json
@@ -1,4 +1,4 @@
 {
   "dialect": "Draft7",
-  "version": "0.82.01",
+  "version": "0.82.01"
 }
--- a/bowtie/templates/migrate-report.html
+++ b/bowtie/templates/migrate-report.html
@@ -110,9 +110,7 @@
               </tr>
             </thead>
 
-            <tbody id="tbody" class="table-group-divider">
-             
-            </tbody>
+            <tbody id="tbody" class="table-group-divider"></tbody>
 
             <tfoot>
               <tr>
@@ -188,6 +186,7 @@
           //updating summary table
           tbody = document.getElementById("tbody");
           for (i = 0; i < data.summary.implementation.length; i++) {
+            //th section of summary table's data
             var tr = document.createElement("tr");
             var th = document.createElement("th");
             th.setAttribute("scope", "row");
@@ -199,33 +198,83 @@
             );
             a.textContent = `${data.summary.implementation[i].name} `;
 
-            const small = document.createElement("small");
-            small.setAttribute(
+            const smallTH = document.createElement("small");
+            smallTH.setAttribute(
               "id",
               `implementation-${data.summary.implementation[i].language}`
             );
-            small.setAttribute("class", "text-muted");
-            small.textContent = `${data.summary.implementation[i].language}`;
+            smallTH.setAttribute("class", "text-muted");
+            smallTH.textContent = `${data.summary.implementation[i].language}`;
 
             th.appendChild(a);
-            th.appendChild(small);
+            th.appendChild(smallTH);
             tr.appendChild(th);
+            
+
+            //td section of summary table's data
+            var td = document.createElement("td");
+            const smallTD = document.createElement("small");
+            smallTD.setAttribute("class", "font-monospace text-muted");
+            smallTD.textContent = `${data.summary.implementation[i].version}`;
+
+            const buttonTD = document.createElement("button");
+            buttonTD.setAttribute("class", "btn border-0");
+
+            const svgTD = document.createElementNS(
+              "http://www.w3.org/2000/svg",
+              "svg"
+            );
+            svgTD.setAttribute("width", "16");
+            svgTD.setAttribute("height", "16");
+            svgTD.setAttribute("fill", "currentColor");
+            svgTD.setAttribute("class", "bi bi-info-circle-fill");
+            svgTD.setAttribute("viewBox", "0 0 16 16");
+            svgTD.setAttribute("data-bs-toggle", "modal");
+            svgTD.setAttribute(
+              "data-bs-target",
+              `#implementation-runtime-info`
+            );
+
+            const pathTD = document.createElementNS(
+              "http://www.w3.org/2000/svg",
+              "path"
+            );
+            pathTD.setAttribute(
+              "d",
+              "M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16zm.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"
+            );
+
+            svgTD.appendChild(pathTD);
+            buttonTD.appendChild(svgTD);
+
+            var td = document.createElement("td");
+            td.appendChild(smallTD);
+            td.appendChild(buttonTD);
+
+            tr.appendChild(td);
+
+
+            //final section of table body
+            var td2 = document.createElement('td');
+            td2.setAttribute('class', 'text-center');
+            td2.textContent = `${data.summary.counts.errored_cases}`;
+
+            var td3 = document.createElement('td');
+            td3.setAttribute('class', 'text-center');
+            td3.textContent += `${data.summary.counts.skipped_tests}`;
+
+            var td4 = document.createElement('td');
+            td4.setAttribute('class', 'text-center');
+            td4.textContent += `${
+              data.summary.counts.errored_cases +
+              data.summary.counts.skipped_tests
+            }`;
+            tr.appendChild(td2);
+            tr.appendChild(td3);
+            tr.appendChild(td4);
+            
             tbody.appendChild(tr);
           }
-
-          var countsErroredCases = document.getElementById(
-            "counts-errored_cases"
-          );
-          countsErroredCases.textContent = `${data.summary.counts.errored_cases}`;
-          var countsSkippedTests = document.getElementById(
-            "counts-skipped_tests"
-          );
-          countsSkippedTests.textContent += `${data.summary.counts.skipped_tests}`;
-          var countsTotalCases = document.getElementById("counts-total-tests");
-          countsTotalCases.textContent += `${
-            data.summary.counts.errored_cases +
-            data.summary.counts.skipped_tests
-          }`;
 
           //summary total
           var summaryErroredCases = document.getElementById(
--- a/bowtie/templates/migrate-report.html
+++ b/bowtie/templates/migrate-report.html
@@ -50,7 +50,26 @@
         </a>
       </div>
     </nav>
-    <div id="report"></div>
+
+    <div class="container p-4">
+      <div class="card mx-auto mb-3 w-75" id="run-info">
+        <div class="card-header">Run Info</div>
+        <div class="card-body">
+          <table class="table table-sm table-hover">
+            <tr>
+              <td>Dialect</td>
+              <td id="dialect-link"></td>
+            </tr>
+            <tr>
+              <td>Ran</td>
+              <td>
+                <time id="run-started"></time>
+              </td>
+            </tr>
+          </table>
+        </div>
+      </div>
+    </div>
 
     <script>
       let dark = window.matchMedia("(prefers-color-scheme: dark)").matches;
@@ -59,14 +78,48 @@
       fetch("data.json")
         .then((response) => response.json())
         .then((data) => {
-          console.log(data);
           const titleElement = document.querySelector("title");
           titleElement.textContent += ` ${data.dialect}`;
 
           //adding bowtie version
           var bowtieVersion = document.getElementById("bowtie-version");
           bowtieVersion.textContent += `${data.version}`;
+
+          //adding dialect link in #run-info
+          var dialectLink = document.querySelector("#dialect-link");
+          dialectLink.textContent = `${data.dialectLink}`;
+
+          let ago = (Date.now() - Date.parse(`${data.runStarted}`)) / 1000;
+          console.log(ago);
+          let agoUnits = "seconds";
+          if (ago > 60) {
+            ago /= 60;
+            agoUnits = "minutes";
+            if (ago > 60) {
+              ago /= 60;
+              agoUnits = "hours";
+              if (ago > 24) {
+                ago /= 24;
+                agoUnits = "days";
+                if (ago > 7) {
+                  ago /= 7;
+                  agoUnits = "weeks";
+                }
+              }
+            }
+          }
+          const roundedAgo = Math.round(ago);
+          const formattedAgo = Number.isFinite(roundedAgo)
+            ? new Intl.RelativeTimeFormat("en", { numeric: "auto" }).format(
+                -roundedAgo,
+                agoUnits
+              )
+            : "";
+          console.log(roundedAgo);
+          document.getElementById("run-started").textContent = formattedAgo;
         });
+
+      //setting datetime attribute
 
       //changing theme icon
       //moon icon
--- a/bowtie/templates/migrate-report.html
+++ b/bowtie/templates/migrate-report.html
@@ -60,7 +60,56 @@
           const titleElement = document.querySelector("title");
           titleElement.textContent += ` ${data.dialect}`;
 
-          
+          const themeTogglerButton = document.getElementById("theme-toggler");
+          let dark = window.matchMedia("(prefers-color-scheme: dark)").matches;
+          if (dark) {
+            var themeSvg = document.createElementNS(
+              "http://www.w3.org/2000/svg",
+              "svg"
+            );
+            themeSvg.setAttribute("xmlns", "http://www.w3.org/2000/svg");
+            themeSvg.setAttribute("width", "20");
+            themeSvg.setAttribute("height", "20");
+            themeSvg.setAttribute("fill", "currentColor");
+            themeSvg.setAttribute("class", "bi bi-moon-fill");
+            themeSvg.setAttribute("viewBox", "0 0 16 16");
+            const path = document.createElementNS(
+              "http://www.w3.org/2000/svg",
+              "path"
+            );
+            path.setAttribute(
+              "d",
+              "M6 .278a.768.768 0 0 1 .08.858 7.208 7.208 0 0 0-.878 3.46c0 4.021 3.278 7.277 7.318 7.277.527 0 1.04-.055 1.533-.16a.787.787 0 0 1 .81.316.733.733 0 0 1-.031.893A8.349 8.349 0 0 1 8.344 16C3.734 16 0 12.286 0 7.71 0 4.266 2.114 1.312 5.124.06A.752.752 0 0 1 6 .278z"
+            );
+            themeSvg.appendChild(path);
+
+          } else {
+            var themeSvg = document.createElementNS(
+              "http://www.w3.org/2000/svg",
+              "svg"
+            );
+            themeSvg.setAttribute("xmlns", "http://www.w3.org/2000/svg");
+            themeSvg.setAttribute("width", "20");
+            themeSvg.setAttribute("height", "20");
+            themeSvg.setAttribute("fill", "currentColor");
+            themeSvg.setAttribute("class", "bi bi-sun");
+            themeSvg.setAttribute("viewBox", "0 0 16 16");
+
+            const path = document.createElementNS(
+              "http://www.w3.org/2000/svg",
+              "path"
+            );
+            path.setAttribute(
+              "d",
+              "M8 11a3 3 0 1 1 0-6 3 3 0 0 1 0 6zm0 1a4 4 0 1 0 0-8 4 4 0 0 0 0 8zM8 0a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 0zm0 13a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 13zm8-5a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2a.5.5 0 0 1 .5.5zM3 8a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2A.5.5 0 0 1 3 8zm10.657-5.657a.5.5 0 0 1 0 .707l-1.414 1.415a.5.5 0 1 1-.707-.708l1.414-1.414a.5.5 0 0 1 .707 0zm-9.193 9.193a.5.5 0 0 1 0 .707L3.05 13.657a.5.5 0 0 1-.707-.707l1.414-1.414a.5.5 0 0 1 .707 0zm9.193 2.121a.5.5 0 0 1-.707 0l-1.414-1.414a.5.5 0 0 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .707zM4.464 4.465a.5.5 0 0 1-.707 0L2.343 3.05a.5.5 0 1 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .708z"
+            );
+            themeSvg.appendChild(path);
+          }
+          themeTogglerButton.appendChild(themeSvg);
+
+          var bowtieVersion = document.getElementById('bowtie-version');
+          bowtieVersion.textContent += `${data.version}`;
+        });
     </script>
   </body>
 </html>
--- a/bowtie/templates/migrate-report.html
+++ b/bowtie/templates/migrate-report.html
@@ -69,6 +69,68 @@
           </table>
         </div>
       </div>
+
+      <div class="card mx-auto mb-3 w-75" id="summary">
+        <div class="card-header">Summary</div>
+
+        <div class="card-body">
+          <table class="table table-sm table-hover">
+            <thead>
+              <tr>
+                <th
+                  colspan="2"
+                  rowspan="2"
+                  scope="col"
+                  class="text-center align-middle"
+                >
+                  implementation
+                </th>
+                <th colspan="1" class="text-center">
+                  <span id="total_cases" class="text-muted">cases</span>
+                </th>
+                <th colspan="3" class="text-center">
+                  <span id="total_tests" class="text-muted">tests</span>
+                </th>
+                <th colspan="1"></th>
+              </tr>
+
+              <tr>
+                <th scope="col" class="text-center">errors</th>
+
+                <th scope="col" class="table-bordered text-center">skipped</th>
+                <th
+                  scope="col"
+                  class="table-bordered text-center details-required"
+                >
+                  <div class="hover-details details-desc text-center"></div>
+                  unsuccessful
+                </th>
+
+                <th scope="col"></th>
+              </tr>
+            </thead>
+
+            <tbody id="tbody" class="table-group-divider">
+             
+            </tbody>
+
+            <tfoot>
+              <tr>
+                <th scope="row" colspan="2">total</th>
+
+                <td id="summary-errored_cases" class="text-center"></td>
+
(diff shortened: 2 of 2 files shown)


Julian:
Great! Will have a closer look after the weekend, but:

> The resulting output comprised various objects that did not conform to the proper JSON format. (It's not even an array of objects)

This kind of file is a line-based JSON file. It's quite useful and common (if you want to think about why, it doesn't require reading the entire file into memory in order to deserialize it). It's in fact so common that it was in widespread use before it was even "specified", if I recall correctly. These days there's e.g. [jsonl](https://jsonlines.org/) which does so -- but it's so simple that most people who use it likely don't know that exist I expect.

Anyways, to decode it, simply iterate over each of the lines in the file and decode each line as JSON!

You can look at the code in `bowtie._report` which does this, and which is essentially what you're replicating from Javascript.

Julian:
Looking again at this -- this is sort of in the right direction yeah, though your sample file is sort of "cheating" at the actual work, in the sense that you're essentially avoiding using the file and have created pre-rolled-up data. Of course the point of dynamically generating the data is to not have static data. Hopefully the tip above about how to parse the file helped, but you essentially want to look at how the existing report file does the rollup you did (presumably by hand) and essentially do something like it from Javascript.

Let me know if that makes sense.

AgniveshChaubey:
Yeah, I know the sample file was a bit of a cheat, but I was in a rush and needed to test whether I was on the right track. My bad!

I was having trouble reading the JSON file because I didn't know about JSON lines, but thanks to your tip, I think I can get the actual report data now.
However, the end-sem exam is going on right now, so it might take me a bit to get around to making the changes. But I'll try to work on it as soon as I can.

Thanks again for your help! :)

AgniveshChaubey:
Update:

Hi @Julian, I have successfully retrieved the data from the original report and updated some of the sections. Although I have done this primarily for draft7 at present, I will soon add the other drafts for conditional rendering, as soon as exams are over.

Julian:
Don't worry, there's no rush, take care of your schoolwork, but I'm going to hold off on reviewing until your PR doesn't contain the JSON rollup file, just FYI :)

> I will soon add the other drafts

I'd expect you to do this by basically (dynamically) making use of the current page URI -- in other words, if the current page URI is `/foo`, then retrieve `/foo.json.gz` -- then your page will work correctly for all drafts at once, if someone loads `/draft7`, they should see the data coming from `/draft7.json.gz`.

Good luck on the exams!

AgniveshChaubey:
Hi @julian, I'm having some queries regarding structuring the data. 
As per my understanding, the `bowtie report` command calls the `from_input()` function from **_report** file to structure the data suitable for the report. And in the case of the compressed file (which we are grabbing from the URL - `https://bowtie-json-schema.github.io/bowtie/draft7.json.gz`), the data it is having is what's being spitted by `bowtie suite` command. 

So, in this case, too, we will have to structure that data.

For the former case, the data is being structured by the following code 
```
def from_input(
    input: Iterable[str],
    generate_dialect_navigation: bool = False,
) -> ReportData:
    """
    Create a structure suitable for the report template from an input file.
    """
    lines = (json.loads(line) for line in input)
    run_info = RunInfo(**next(lines))
    summary = run_info.create_summary()

    for each in lines:
        if "case" in each:
            summary.add_case_metadata(**each)
        elif "caught" in each:
            summary.see_error(**each)
        elif "skipped" in each:
            del each["skipped"]
            summary.see_skip(_commands.CaseSkipped(**each))
        elif "did_fail_fast" in each:
            summary.see_maybe_fail_fast(**each)
        else:
            summary.see_result(_commands.CaseResult.from_dict(each))
    return ReportData(
        summary=summary,
        run_info=run_info,
        generate_dialect_navigation=generate_dialect_navigation,
    )
``` 

This means that I have to create equivalent code in JS too, to structure data, but do I need to define the **RunInfo**, **CaseSkipped**, **CaseResult**, etc classes again in JS? 

Please correct me if I am wrong. I have been stuck on this for the last 2 days and would appreciate any hints to help me proceed further.

Julian:
> This means that I have to create equivalent code in JS too, to structure data, but do I need to define the RunInfo, CaseSkipped, CaseResult, etc classes again in JS?

Good question! Yes, I think that's exactly what I was expecting to happen.

To some extent, if JSON Schema had good tooling to do this, we'd actually try generating those classes from a JSON Schema itself! And then we could use them from both Python and JS (and in fact I wouldn't have needed to manually create those classes in the first place) Doing that is #42, but I actually don't know of any tooling good enough to do it for us.

So for now I indeed think the best plan is to try to swag those out from JavaScript as well. Lemme know if you run into trouble.

bowtie/utilclasses.js
nomandhoni-cs, line 1: More specifically this file?

Or migrate-report.html?

AgniveshChaubey: Both have to be reviewed, but firstly reviewing migrate-report.html would be better!

AgniveshChaubey:
Hi @Julian, 

The UI part is completed, but the data rendering in some sections is still remaining as of now. 
At this point two of the main issues I am facing are...

1. The `failed_tests()` function is not rendering accurate data. I tried creating this function similar to the **Python** one, but as there are a lot of function calls one inside the other, I am feeling messed up. I think still I'm not able to fully understand the raw data (more specifically results of testing ones --- how they are being utilized) and that's why I'm facing difficulties in creating custom functions. 

2. Secondly, the **_results_** object yielded by [flat_results()](https://github.com/bowtie-json-schema/bowtie/pull/241/files#diff-f0f9e791a2914d1e218977a8a9341c25ff9f5f45d38255fd63625019bfe7a471R230) is not having any content (it's empty-- don't know why). Temporarily I've added some content explicitly to avoid errors, but we'll need to fix that.

I need your assistance in solving these issues. And would really appreciate if you could help me understand the raw data more clearly.

bowtie/utilclasses.js
```diff
@@ -0,0 +1,443 @@
...
+          return 1;
+        } else {
+          return 0;
+        }
+      },
+    );
+    let sorted_CombinedObject = Object.fromEntries(sorted_CombinedArray);
+    // console.log(sorted_CombinedObject)
+    for (let seq in sorted_CombinedObject) {
+      let each = sorted_CombinedObject[seq];
```
VishrutAggarwal, line 245: Hi, I don't think this line is required here. `seq` is already the "each" in `sorted_CombinedArray` and seems to be the problem. `seq` is not a key, it's a value (in context to a key-value pair).

VishrutAggarwal (changes requested):

bowtie/utilclasses.js
```diff
@@ -0,0 +1,443 @@
...
+  case_results() {
+    return Object.values(this._combined).map((each) => ({
+      case: each.case,
+      registry: each.registry || {},
+      results: each.results,
+    }));
+  }
+
+  *flat_results() {
+    let sorted_CombinedArray = Object.entries(this._combined).sort(
```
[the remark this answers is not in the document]
AgniveshChaubey: `this._combined` is an object (dict in python) and `Object.entries` function converts the key-value pairs of object into array of arrays. For eg---
```
{
"name": "Agnivesh",
"age": "20"
}
```

to

```
[
["name", "Agnivesh"], ["age", "20"]
]
```

The purpose of doing this was to sort the object in ascending order (direct sorting on object is not possible, that's why...)

bowtie/utilclasses.js
AgniveshChaubey: No, in the raw data, `seq` is key.

bowtie/report/public/index.html
```diff
@@ -0,0 +1,29 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Bowtie –</title>
+    <link
+      href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/css/bootstrap.min.css"
```
harrel56, line 7: Since there is now a bundling process for UI, maybe *bootstrap* could be put as a dependency in `package.json` file?

bowtie/report/public/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <script
+      src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/js/bootstrap.bundle.min.js"
+      integrity="sha384-ENjdO4Dr2bkBIFxQpeoTz1HIcje39Wm4jDKdf19U8gI4ddQ3GYNS7NTKfAdVQSZe"
+      crossorigin="anonymous"
+    ></script>
+
+    <script>
+      const script = document.createElement("script");
+      script.src = "https://cdn.jsdelivr.net/npm/pako@2.0.3/dist/pako.min.js";
+      document.head.appendChild(script);
```
harrel56, line 26: I guess this lib could be moved to `package.json` too, but I feel like we shouldn't really be unpacking gzips on frontend? I think probably that build/deploy process should take care of unpacking

bowtie/report/src/components/AccordionSvg.js
```diff
@@ -0,0 +1,36 @@
...
+      <div>
+        <svg
+          xmlns="http://www.w3.org/2000/svg"
+          width="16"
+          height="16"
+          fill="currentColor"
+          className="bi bi-check-circle-fill"
+          viewBox="0 0 16 16"
+        >
+          <path d="M16 8A8 8 0 1 1 0 8a8 8 0 0 1 16 0zm-3.97-3.03a.75.75 0 0 0-1.08.022L7.477 9.417 5.384 7.323a.75.75 0 0 0-1.06 1.06L6.97 11.03a.75.75 0 0 0 1.079-.02l3.992-4.99a.75.75 0 0 0-.01-1.05z" />
```
harrel56, line 29: Maybe all SVG icon definition like this could be put into separate files and then imported in JS code? I remember that there was some common solution for that

bowtie/report/src/index.js
```diff
@@ -0,0 +1,25 @@
+import { createRoot } from "react-dom/client";
+import App from "./App";
+import { CountsDataProvider } from "./data/CountsDataContext";
+
+document.addEventListener("DOMContentLoaded", () => {
+  const Dialect = "2020-12";
+
+  //fetching json data from url
+  fetch(`https://bowtie-json-schema.github.io/bowtie/draft${Dialect}.json.gz`)
```
harrel56, line 9: Is there a way to use data generated by local bowtie?

bowtie/report/src/data/report.js
```diff
@@ -0,0 +1,107 @@
...
+    // console.log(this.counts)
+    this.initializeCounts();
+  }
+  initializeCounts() {
+    for (const each of this.implementations) {
+      this.counts[each.image] = new Count();
+    }
+  }
+
+  get total_cases() {
```
harrel56, line 26: Is there a reason for using a snake case? Generally, I would consider camel case to be an accepted standard for JS

bowtie/report/public/index.html
AgniveshChaubey: I have used cdn in order to keep the application lightweight. (Also it might increase the loading speed because of caching).

bowtie/report/public/index.html
AgniveshChaubey: At this stage, the UI is fetching data from the local bowtie webpage (`https://bowtie-json-schema.github.io/bowtie/`${dialect_version}`.json.gz`). Also we are 6 **_dialects_** (`2020-12`, `2019-09`, `7`, `6`, `4`, `3`) --- that means 6 data files. I think it would be more feasible to dynamically fetch the data based on user's choice of **_dialect_** (to keep the app lightweight) rather than collecting all the data at once and bundling them in our app. However, let me know what do you think.

bowtie/report/src/components/AccordionSvg.js
AgniveshChaubey: Ya I have kept them in a separate svg folder initially, but there was an issue in mode toggling. --- I stored svgs in .svg file and but in that case, I had to use the path as **value** to **src** property. And so, the SVG is now an image and we can't add value of `fill` attribute dynamically  based on the current theme. (means we have to use two different svgs --- I was trying to use only one). 

Btw, there are other ways too. I'll do that!

bowtie/report/src/index.js
AgniveshChaubey: Yes of course! The raw data can be found here (in compressed form) ----
`https://bowtie-json-schema.github.io/bowtie/draft2020-12.json.gz` (just replace the **_dialect_** version of your choice)

bowtie/report/src/data/report.js
AgniveshChaubey: Not actually!😅 I'll change that.

bowtie/report/public/index.html
harrel56: I mean, either way you still gonna need to download the same amount of bytes. I can see few advantages of putting it in package.json:
- you have all your dependencies in one place - it's easier to manage and e.g. dependabot could see this dependencies
- not relying on 3rd party sites - we don't need to worry this URL stopping to work (or changing) spontaneously. Each UI build/bundle would be independent and it's behaviour should be reproducible
- bundle produced by UI could also be cached as a whole or if it's necessary it can be split to chunks with different cache mechanics for each chunk.

bowtie/report/public/index.html
harrel56: I'm only suggesting that if we need this data to be unpacked we could have it uploaded in unpacked version. Currently, each user's browser is doing the unpacking process (affecting user experience I guess? But i don't know how much time does this process take), when it could be done just once during building/deployment or at any stage it is convenient

bowtie/report/src/components/AccordionSvg.js
harrel56: I remember there were also whole icon packs (used as any other dependencies) which could be imported in JS. I'm guessing we don't use any custom icons?

bowtie/report/src/index.js
harrel56: Hmm, to be precise I meant e.g. using some command like this: `bowtie suite ... / bowtie report` and then using produced data by this command to populate UI

bowtie/report/public/index.html
Julian: The original reason to investigate was that the unpacked data for each draft is ~10MB or so (#146), and the compressed data is ~50KB. so it was a noticeable reduction in load time I think. But we can reconsider of course (it'd definitely be nice if GitHub supported Content-Type and then we'd obviously get the browser automatically decompressing the gzip before the app saw it)

bowtie/report/src/index.js
AgniveshChaubey: Ohh! My bad! Do you mean populating the local UI or the hosted one?

bowtie/report/public/index.html
harrel56: Hmm, that is what I was thinking - the GH pages should expose static files which are **NOT** gzipped, but the underlying server should handle their compressing if client requires so. And the browser should unpack it automatically. So we could use URL like `https://bowtie-json-schema.github.io/bowtie/draft2020-12.json` (without gzip extension), and if client sends request with header `Accept-Encoding=gzip` it should receive gzipped response (with response header `Content-Encoding=gzip`). It looks like e.g. getting bowtie HTML via https://bowtie-json-schema.github.io/bowtie is working exactly like this:

- without `Accept-Encoding` header: ~8.6mb
- with `Accept-Encoding` header: ~200kb

Haven't had time to read the whole discussion from #146 so maybe I missed some important details. Let me know if this approach makes sense

bowtie/report/src/index.js
harrel56: I mean local UI with local report data. Now the URL is hardcoded for specific host (https://bowtie-json-schema.github.io), maybe this URL can be relative so it could work locally too? Also, if in the future hosting platform gonna change then it shouldn't really affect code behaviour. So I would suggest something like `/bowtie/draft${Dialect}.json.gz`, but you gonna need to have these report files locally. And somehow expose them on live local server

frontend/src/assets/svg/Svg.js
```diff
@@ -0,0 +1,109 @@
...
+      return (
+        <svg
+          xmlns="http://www.w3.org/2000/svg"
+          width="20"
+          height="20"
+          fill="currentColor"
+          className="bi bi-moon-stars-fill"
+          viewBox="0 0 16 16"
+        >
+          <path d="M6 .278a.768.768 0 0 1 .08.858 7.208 7.208 0 0 0-.878 3.46c0 4.021 3.278 7.277 7.318 7.277.527 0 1.04-.055 1.533-.16a.787.787 0 0 1 .81.316.733.733 0 0 1-.031.893A8.349 8.349 0 0 1 8.344 16C3.734 16 0 12.286 0 7.71 0 4.266 2.114 1.312 5.124.06A.752.752 0 0 1 6 .278z" />
```
harrel56, line 13: It looks better separated like that, but I had something like that in mind: https://www.npmjs.com/package/react-bootstrap-icons. Let me know what you think about this approach

I only wonder why it's not part of official react-bootstrap package

frontend/src/components/Cases/AccordionItem.js
```diff
@@ -0,0 +1,199 @@
...
+import Svg from "../../assets/svg/Svg";
+
+const AccordionItem = ({ eachCase, implementations, caseImplementation }) => {
+  const seq = eachCase.seq;
+  const description = eachCase.case.description;
+  const schema = eachCase.case.schema;
+  const tests = eachCase.case.tests;
+
+  function result(index, implementation) {
+    var testResult;
```
harrel56, line 12: Any reason why you use `var` in some places? Checked few of them and looks like they can be replaced by `let` or `const`. I think that as a general rule of thumb the latter two should be used - they are much more intuitive in terms of scope and visibility

frontend/src/components/Cases/AccordionItem.js
```diff
@@ -0,0 +1,199 @@
...
+    if (implementation.skipped && implementation.skipped == true) {
```
harrel56, line 13: Wouldn't `if (implementation.skipped)` be enough? Or if you really want to check for `true` you could write `if (implementation.skipped === true)`. Same goes for conditions below

frontend/src/components/Cases/AccordionItem.js
```diff
@@ -0,0 +1,199 @@
...
+      return (testResult = "skipped");
```
harrel56, line 14: What is the reason for assigning value to `testResult` on return?

frontend/src/components/Cases/AccordionItem.js
```diff
@@ -0,0 +1,199 @@
...
+    } else if (implementation.results && implementation.expected) {
+      var caseResults = implementation.results.filter(
+        (element) => element.skipped,
+      );
+      if (caseResults.length > 0) {
+        return (testResult = "skipped");
+      }
+      if (
+        implementation.results.every(
+          (each) => each.hasOwnProperty("errored") && typeof each === "object",
```
harrel56, line 26: If `each` indeed can be something different than `object`, the AND conditions should be reversed

frontend/src/components/Cases/AccordionSvg.js
```diff
@@ -0,0 +1,44 @@
+import Svg from "../../assets/svg/Svg";
+
+const AccordionSvg = ({ testResult }) => {
+  if (testResult === "skipped") {
```
harrel56, line 4: Maybe `switch-case` can be used instead?

.gitignore
```diff
@@ -2,6 +2,12 @@
 bowtie-report.html
 /badges/
 
+# files to be ignored within report directory
```
Julian, line 5: (This comment is slightly inaccurate now, but probably fine to just remove it)

.gitignore
```diff
@@ -153,7 +159,7 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+# .idea/
```
Julian, line 162: Didn't you mean to uncomment this? Right now it just adds a space, but the comment character (`#`) is still there.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Julian, line 1: The `.vscode` folder is `gitignored` but looks like this file is committed before doing so (so probably you want to run `git rm -f .vscode/settings.json` or thereabouts so that you're not proposing adding it in this PR.

README.rst
```diff
@@ -41,3 +41,10 @@ Looks like a bowtie, no?
 Also because it's elegant – we hope.
 
 For more information, see `Bowtie's documentation <https://bowtie-json-schema.readthedocs.io/>`_.
+
+
+To run the bowtie UI locally, follow the below steps.
+
+Make sure you have node and npm installed in your system.
+Navigate to path_to_bowtie/report
```
Julian, line 49: This is `frontend/` now I think.

frontend/package.json
```diff
@@ -0,0 +1,29 @@
+{
```
Julian, line 1: I can't comment on it directly (because the silly GitHub UI doesn't allow it) but if you look above, it looks like you're also committing an empty `bowtie/templates/migrate-report.html` file as part of this PR.

frontend/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "bowtie",
+  "version": "1.0.0",
+  "description": "Schema validator app",
```
Julian, line 4: Possibly copy the description from the `pyproject.toml` file here (or copy it and mention this is the UI for it).

frontend/public/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Bowtie –</title>
```
Julian, line 5: Is there supposed to be a template variable or something here after the –? I'm no react expert, does something magically append page names to the `<title>` or yeah is there supposed to be something after that?

frontend/src/assets/svg/Svg.js
Julian: Yeah ideally this whole file should be replaced by that I hope.

frontend/src/components/NavBar.js
```diff
@@ -0,0 +1,125 @@
...
+                <a className="nav-link" href="#summary">
+                  Summary
+                </a>
+              </li>
+              <li className="nav-item">
+                <a className="nav-link" href="#cases">
+                  Details
+                </a>
+              </li>
+              {generateDialectNavigation && (
```
Julian, line 57: This is more of a TODO than something to address now, but now that we have a dynamic site, we should likely also have a dynamic way to populate this navigation -- specifically rather than hardcoding which dialects we support, we should probably simply automatically populate this list by querying which dialects are available. Perhaps just file an issue for us to do so after this PR is merged?

frontend/src/components/RunInfo/RunInfoSection.js
```diff
@@ -0,0 +1,85 @@
+const RunInfoSection = (props) => {
+  let ago = (Date.now() - Date.parse(props.runInfo.started)) / 1000;
```
Julian, line 2: Now that we have some package management I suspect you should be able to find some humanize JS library that does this better than my silly hand rolled one as well. I think GitHub has a library they published for doing this maybe?

frontend/src/components/Summary/ImplementationRow.js
```diff
@@ -0,0 +1,162 @@
+import "./ImplementationRow.css";
+import Svg from "../../assets/svg/Svg";
+
+const ImplementationRow = ({ lines, implementation, index }) => {
+  const implementationArray = lines.filter((element) => element.implementation);
+  const caseArray = lines.filter((element) => element.case);
+  //   const {
```
Julian, line 7: Looks like there's some dead code here.

frontend/src/utilities/utilFunctions.js
```diff
@@ -0,0 +1,81 @@
+class _Summary {
+  constructor(implementations) {
+    // console.log(typeof(implementations.implementations))
```
Julian, line 3: And more dead code here.

frontend/src/utilities/utilFunctions.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+export function copyToClipboard(targetTextToCopy, btnId) {
+  const clipBoardIcon = document.querySelector(".bi-clipboard");
+  const textToCopy = document.getElementById(targetTextToCopy).innerText;
+  navigator.clipboard.writeText(textToCopy);
+  document.getElementById(
+    btnId,
+  ).innerHTML = `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-check" viewBox="0 0 16 16">
```
Julian, line 75: Here too should be replaced by some icon package I hope.

report/src/components/Modals/DetailsButtonModal.js
```diff
@@ -0,0 +1,246 @@
...
+                Object.keys(element).length === 1 &&
+                "valid" in element,
+            )
+          ) {
+          }
+        } else if (seqImplementation.results) {
+          var caseResults = seqImplementation.results.filter(
+            (seqImplementation) => seqImplementation.context,
+          );
+          // console.log(caseResults);
```
Julian, line 65: More dead code sprinkled in this file -- good to do a pass and just look for any lines you've commented out -- perhaps also look at whether eslint can raise warnings for such things, I suspect/hope it can, so we should enable that -- let me know if you want help figuring out how.

Julian (commented):
Left some additional minor notes -- agree with @harrel56's comments too.

Looking nice though, good progress so far!

bowtie/report/public/index.html
AgniveshChaubey: Ya, that's totally reasonable. It indeed increased the loading speed!

AgniveshChaubey: Ya, that's totally reasonable. Till now, I too was thinking that pages didn't support `Accept-Encoding`

frontend/src/assets/svg/Svg.js
AgniveshChaubey: Sure, I'll do that!

frontend/src/components/Cases/AccordionItem.js
AgniveshChaubey: No. I'll change then to `let`/`const`

AgniveshChaubey: No. I'll change them to `let`/`const`

frontend/src/components/Cases/AccordionItem.js
AgniveshChaubey: Made the changes!

frontend/src/components/Cases/AccordionItem.js
AgniveshChaubey: Ya, it's unnecessary I guess. Removed!

frontend/src/components/Cases/AccordionItem.js
AgniveshChaubey: I think it unintentionally changed during some random try. Fixed it now!

frontend/src/components/Cases/AccordionSvg.js
AgniveshChaubey: Done

frontend/src/assets/svg/Svg.js
AgniveshChaubey: Done

README.rst
AgniveshChaubey: Yes. updated

frontend/public/index.html
AgniveshChaubey: The dialectShortName is being dynamically appended by the following code in `App.js` file.

```
  document.getElementsByTagName("title")[0].textContent +=
    " " + runInfo.dialectShortName;

```

frontend/src/components/NavBar.js
AgniveshChaubey: Sure, I'll file an issue after this PR gets merged.

frontend/src/components/Summary/ImplementationRow.js
AgniveshChaubey: oops! I was trying to use React's context API to pass the data to the parent component but then used the state-lifting method. Either way we can do that. Btw removed the dead code.

frontend/src/utilities/utilFunctions.js
AgniveshChaubey: addressed!

report/src/components/Modals/DetailsButtonModal.js
AgniveshChaubey: Cleaned it up!

AgniveshChaubey: > perhaps also look at whether eslint can raise warnings for such things...

Ya please help me to understand how ESLint works.

bowtie/report/src/index.js
AgniveshChaubey: "We can definitely add conditional fetching of data based on where the app is being used, whether to fetch data from the **_remote URL_** or from a **_local path_**. Regarding the idea of using a **_relative URL_**, I believe we can address it in a separate pull request in the near future. @Julian, please let me know your thoughts on this.

Julian: Yeah totally fine with me, @harrel56 we definitely want to preserve that behavior but yeah we can perhaps get this one merged and then restore it i in a next PR (I think being able to "upload" a local file is probably the most straightforward way to resupport it but we can brainstorm)

report/src/components/Modals/DetailsButtonModal.js
Julian: Definitely! We can probably hit this too in a follow-up PR, no need to block this one, but yeah I'll show you how it works!

Julian:
Can you also by the way merge main again, that should fix the CI failures

Julian:
Think we're about ready to merge this and handle some of the follow-ups in a next few PRs -- anyone have last comments here? Otherwise I'll take a final look myself and probably merge in a few hours! Thanks @AgniveshChaubey!

frontend/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test"
+  },
+  "dependencies": {
+    "bootstrap": "^5.3.0",
+    "pako": "^2.1.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "react-scripts": "5.0.0"
```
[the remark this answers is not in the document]
AgniveshChaubey: Ah! I missed the `--save` flag

Julian:
Going to merge, congrats @AgniveshChaubey -- plenty left to do and I see a bunch of minor improvements we can make, but we can do em in further PRs!

PR #260, opened 2023-05-18
Added Supported Draft Versions Badge

VishrutAggarwal:
This PR adds the Supported Drafts Badge as mentioned in Issue #55.

### DIFF
--- a/bowtie/_report.py
+++ b/bowtie/_report.py
@@ -351,9 +351,7 @@ class _Summary:
             }
             supp_dir = target_dir / f"{lang}-{name}"
             supp_dir.mkdir(parents=True, exist_ok=True)
-            badge_path_supp_drafts = (
-                supp_dir / "supported_versions.json"
-            )
+            badge_path_supp_drafts = supp_dir / "supported_versions.json"
             badge_path_supp_drafts.write_text(json.dumps(badge_supp_draft))
 
 
--- a/bowtie/_report.py
+++ b/bowtie/_report.py
@@ -333,8 +333,6 @@ class _Summary:
                 - counts.skipped_tests
             )
             pct = (passed / total) * 100
-            impl_dir = target_dir / f"{lang}-{name}"
-            impl_dir.mkdir(parents=True, exist_ok=True)
             r, g, b = 100 - int(pct), int(pct), 0
             badge_per_draft = {
                 "schemaVersion": 1,
@@ -348,8 +346,15 @@ class _Summary:
                 "message": supported_drafts,
                 "color": "lightgreen",
             }
-            badge_path_per_draft_compliance = impl_dir / f"{label.replace(' ', '_')}_perDraftCompliance.json"
-            badge_path2_supported_drafts = impl_dir / f"{label.replace(' ', '_')}_supportedVersions.json"
+            comp_dir = target_dir / "compliance" / f"{lang}-{name}"
+            comp_dir.mkdir(parents=True, exist_ok=True)
+            
+            supp_dir = target_dir / "supported_versions" / f"{lang}-{name}"
+            supp_dir.mkdir(parents=True, exist_ok=True)
+            
+            badge_path_per_draft_compliance = comp_dir / f"{label.replace(' ', '_')}.json"
+            badge_path2_supported_drafts = supp_dir / f"{label.replace(' ', '_')}.json"
+            
             badge_path_per_draft_compliance.write_text(json.dumps(badge_per_draft))
             badge_path2_supported_drafts.write_text(json.dumps(badge_supported_draft))
 
--- a/bowtie/_report.py
+++ b/bowtie/_report.py
@@ -350,7 +350,6 @@ class _Summary:
                 "color": "lightgreen",
             }
             supp_dir = target_dir / f"{lang}-{name}"
-            supp_dir.mkdir(parents=True, exist_ok=True)
             badge_path_supp_drafts = supp_dir / "supported_versions.json"
             badge_path_supp_drafts.write_text(json.dumps(badge_supp_draft))
 
--- a/bowtie/_report.py
+++ b/bowtie/_report.py
@@ -317,8 +317,9 @@ class _Summary:
             dialect_versions = impl["dialects"]
             if dialect not in dialect_versions:
                 continue
-            supported_drafts  = ", ".join(
-                _DIALECT_URI_TO_SHORTNAME[each].removeprefix("Draft ") for each in reversed(dialect_versions)
+            supported_drafts = ", ".join(
+                _DIALECT_URI_TO_SHORTNAME[each].removeprefix("Draft ")
+                for each in reversed(dialect_versions)
             )
             name = impl["name"]
             lang = impl["language"]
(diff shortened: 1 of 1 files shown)


VishrutAggarwal:
Hi, 

Sorry for the delay in this @Julian.
I have generated the supported drafts badge in the report file itself along with the per draft compliance badge

bowtie/_report.py
```diff
@@ -314,8 +314,20 @@ def flat_results(self):
     def generate_badges(self, target_dir: Path, dialect: str):
         label = _DIALECT_URI_TO_SHORTNAME[dialect]
         for impl in self.implementations:
+            dialectVersions = impl["dialects"]
```
Julian, line 317: The convention in Python is different from Javascript, variables are generally named in snake case, so this should be

```suggestion
            dialect_versions = impl["dialects"]
```

bowtie/_report.py
```diff
@@ -314,8 +314,20 @@ def flat_results(self):
...
             if dialect not in impl["dialects"]:
```
Julian, line 318: I assume creating a variable was in order to be able to reuse it, so you should do that here

```suggestion
            if dialect not in dialect_versions:
```

bowtie/_report.py
```diff
@@ -314,8 +314,20 @@ def flat_results(self):
...
                 continue
+
+            SupportedVersions: str = []
```
Julian, line 321: You'll see CI is failing, because this type hint is wrong, you're saying `supported_versions` is a string, but it isn't, it's a `list[str]`. But all of these lines are spelled `str.join` in Python -- so you want simply:

```suggestion
            supported = ", ".join(
                each.removeprefix("Draft ") for each in dialect_versions
            )
```

bowtie/_report.py
```diff
@@ -330,14 +342,26 @@ def generate_badges(self, target_dir: Path, dialect: str):
...
             }
-            badge_path = impl_dir / f"{label.replace(' ', '_')}.json"
-            badge_path.write_text(json.dumps(badge))
+            badgeSupportedDraft = {
+                "schemaVersion": 1,
+                "label": "JSON Schema Versions",
+                "message": SupportedDrafts,
+                "color": "lightgreen",
+            }
+            badge_path1 = (
```
Julian, line 357: You were using names for different badges up until here, let's not switch to opaque numbers now :), call these something related to what the badge is about as well.

bowtie/_report.py
```diff
@@ -330,14 +342,26 @@ def generate_badges(self, target_dir: Path, dialect: str):
...
+                impl_dir / f"{label.replace(' ', '_')}_perDraftCompliance.json"
```
Julian, line 358: We wanted to put these in folders rather than suffixes, so let's introduce separate directories for the two kinds of badges (one probably called `compliance` and the other `supported-versions`).

Julian (changes requested):
Great! No worries, thanks again for giving it a shot!

Left some comments, let me know if any need clarifying!

Thanks again.

VishrutAggarwal:
Okay, should I create a new PR after making the changes or should I just commit the changes to this PR itself?

Julian:
Generally whenever you make a PR you basically commit any changes from feedback to that PR unless it's something unrelated (in which case someone will usually say "let's do XYZ after this" or the like), so yeah here is fine!

bowtie/_report.py
```diff
@@ -327,17 +334,38 @@ def generate_badges(self, target_dir: Path, dialect: str):
                 - counts.skipped_tests
             )
             pct = (passed / total) * 100
-            impl_dir = target_dir / f"{lang}-{name}"
-            impl_dir.mkdir(parents=True, exist_ok=True)
             r, g, b = 100 - int(pct), int(pct), 0
-            badge = {
+            badge_per_draft = {
```
Julian, line 338: This isn't per draft, is it? It's for all drafts I think.

VishrutAggarwal: When I saw the issue #55 , It had the per draft compliance checked off, so I assumed that was the badge for it.

Julian: Oh no it is per draft? I have to look at this again, I'm still confused a bit as to what directory structure you're proposing.

bowtie/_report.py
```diff
@@ -327,17 +334,38 @@ def generate_badges(self, target_dir: Path, dialect: str):
...
                 "schemaVersion": 1,
                 "label": label,
                 "message": "%d%% Passing" % int(pct),
                 "color": f"{r:02x}{g:02x}{b:02x}",
             }
-            badge_path = impl_dir / f"{label.replace(' ', '_')}.json"
-            badge_path.write_text(json.dumps(badge))
+            badge_supported_draft = {
```
Julian, line 344: It would definitely be clearer, at least to me, if you grouped this by badge, not by functionality -- right now you do some calculation for a bunch of badges, then build a dict of bunch of badges, then write out a bunch of badges to files.

It's a lot clearer to me understanding I think to fully handle each badge one by one -- so calculate, build, write the first badge, then the second. And it may even then be clear there's some helper function that makes it even clearer. But I'm going to read this more carefully to remind myself what the existing badge means.

Julian: The merge conflicts are because each time you're pushing, you're failing the linter (which is checking your code style), and so you're getting a commit that fixes that automatically.

There's a few ways to fix it, either look at what's being fixed and avoid doing that :) (in this case it's leaving whitespace on lines I think), or [install `pre-commit`](https://pre-commit.com/#install) which is what's being run, and then run that locally whenever you commit (it can do so automatically if you run `pre-commit install` inside the Bowtie repository, then it will run every time you commit).

Or alternatively, you can run `git pull` before you commit, and that will pull the commit that exists each time you're pushing.

bowtie/_report.py
VishrutAggarwal: As of now, the code creates the directory in the following structure:

- for per-draft-compliance:
`{path-specified-in-cli-command}/compliance/{language-implementation}/{dialect-name}`

- for supported-draft-compliance:
`{path-specified-in-cli-command}/compliance/{language-implementation}/{dialect-name}`

This file structure feels a little redundant to me, if there's a better structure in mind, please share. I'll make the changes accordingly.

bowtie/_report.py
VishrutAggarwal: Okay, that helps a lot, I'll do that.

bowtie/_report.py
Julian: `{path-specified-in-cli-command}/{language-implementation}/compliance/{dialect-name}.json` and `{path-specified-in-cli-command}/{language-implementation}/supported-versions.json` seem less redundant to me (and implementers probably will like having all their badges in one "spot").

VishrutAggarwal:
Let me know if the code requires any more changes.

bowtie/_report.py
```diff
@@ -327,17 +332,27 @@ def generate_badges(self, target_dir: Path, dialect: str):
...
+            comp_dir.mkdir(parents=True, exist_ok=True)
+            badge_path_per_draft = comp_dir / f"{label.replace(' ', '_')}.json"
+            badge_path_per_draft.write_text(json.dumps(badge_per_draft))
+            badge_supp_draft = {
+                "schemaVersion": 1,
+                "label": "JSON Schema Versions",
+                "message": supported_drafts,
+                "color": "lightgreen",
+            }
+            supp_dir = target_dir / f"{lang}-{name}"
```
Julian, line 352: This is `impl_dir` from before the changes (so you could have left it as-is and just made `comp_dir` a child of `impl_dir`), but if you're going to do it this way now, there should be no need for the `.mkdir` on the next line -- the directory must always exist, it's a parent of hte directory you created on line 342, right?

Julian:
One last thing I think, then this is likely to be ready!

VishrutAggarwal:
Done with the changes.

Julian:
Nice, looks great! E.g. here's the new badge for `lua-jsonschema`:



Thanks again, really appreciated!

PR #263, opened 2023-05-21
Added Overall Compliance Badge

VishrutAggarwal:
This PR works on the issue #55.

The code generates the Overall Compliance Badge for the implementations

### DIFF
--- a/bowtie/_report.py
+++ b/bowtie/_report.py
@@ -313,6 +313,8 @@ class _Summary:
 
     def generate_badges(self, target_dir: Path, dialect: str):
         label = _DIALECT_URI_TO_SHORTNAME[dialect]
+        total_pct : float = 0
+        num_of_implementations : int = 0
         for impl in self.implementations:
             dialect_versions = impl["dialects"]
             if dialect not in dialect_versions:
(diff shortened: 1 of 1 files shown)


Julian:
Thanks! I don't think this calculates the right thing though. This seems to calculate something like the average overall compliance across all implementations for one dialect.

But the idea of that badge was not that, it's to calculate the overall percentage for each implementation but across all dialects. Which is actually tricky to do the way we have things set up because right now we run one dialect at a time and generate therefore only badges for one at a time.

But this badge is meant to say that if one implementation gets 99 out of 100 tests right for each of the versions it supports, and it supports 6, 7 and 2020-12, then we generate a badge showing (99+99+99)/300 which is its overall compliance across all versions.

Doing so will likely require separating badge generation from report generation at the minute unfortunately I think...

VishrutAggarwal:
Okay, I think I get your point.

I think this will need to be closed at the moment then (or a will fix label could be added though deleting this branch would be easier in the longer run), I will work on the keywords badge then until this can be discussed upon or if there is any way you could guide me on how to do that I could look into it.

Julian:
Yep sounds good to me

PR #278, opened 2023-06-07
Update report.yml

AgniveshChaubey:
Data compression removed

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -45,14 +45,14 @@ jobs:
 
       - name: Generate the Report
         run: |
-          bowtie suite $(find implementations/ -mindepth 1 -maxdepth 1 -type d | sed 's/implementations\//-i /') https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/${{ matrix.version }} | tee >(gzip >${{ matrix.version }}.json.gz) | bowtie report --badges badges --generate-dialect-navigation --out ${{ matrix.version }}.html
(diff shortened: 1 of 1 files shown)


.github/workflows/report.yml
```diff
@@ -45,14 +45,14 @@ jobs:
 
       - name: Generate the Report
         run: |
-          bowtie suite $(find implementations/ -mindepth 1 -maxdepth 1 -type d | sed 's/implementations\//-i /') https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/${{ matrix.version }} | tee >(gzip >${{ matrix.version }}.json.gz) | bowtie report --badges badges --generate-dialect-navigation --out ${{ matrix.version }}.html
+          bowtie suite $(find implementations/ -mindepth 1 -maxdepth 1 -type d | sed 's/implementations\//-i /') https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/${{ matrix.version }} | > ${{ matrix.version }}.jsonl | bowtie report --badges badges --generate-dialect-navigation --out ${{ matrix.version }}.html
```
Julian, line 48: You still need the `tee`, just not the `>(...)` -- that's called "process substitution" (and was previously redirecting some things into `gzip`, but we still do need `tee` in order to both write out a file as well as pass output along), it probably should be:

```suggestion
          bowtie suite $(find implementations/ -mindepth 1 -maxdepth 1 -type d | sed 's/implementations\//-i /') https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/${{ matrix.version }} | tee ${{ matrix.version }}.jsonl | bowtie report --badges badges --generate-dialect-navigation --out ${{ matrix.version }}.html
```

You can try it out locally by manually replacing the template variable (and picking just one implementation to make it run faster), e.g.:

```
bowtie suite -i python-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft2020-12 | tee draft2020-12.jsonl | bowtie report --badges badges --generate-dialect-navigation --out draft2020-12.html
```

Issue #296, opened 2023-06-16
UI: Render the Dialect list dynamically


AgniveshChaubey:
At present, the `dialect` [list ](https://github.com/bowtie-json-schema/bowtie/blob/main/frontend/src/index.js#L8)is hard-coded, but in the future, when the list updates, we'll have to manually update it in the UI. So, it would be better to render it dynamically instead of hard-coding it.

PR #297, opened 2023-06-16
Feature: Upload locally generated report to bowtie UI

AgniveshChaubey:
This PR enables users to upload locally generated JSON data to the Bowtie UI. It allows for a more informative and comprehensive representation of their local reports.


harrel56 (commented):
I have one major design suggestion, which boils down to:

- instead of using `fetch()` directly in component we could use `loaders` from react-router. As loaders can be parametrized we could have just one route for all the routes that require report data from web. https://reactrouter.com/en/main/route/loader

- ideally, `App` component could be just a component for pure presentation (maybe this component should get a rename?) - by that I mean it won't make any HTTP request or any other things like that, but it will receive report data (`lines`) via prop or via beforementioned `loader` data

- as for page where user is expected to upload local files - it could be implemented as a component that wraps `App` component, and it either renders the file drop area or the `App` with data extracted from provided file

 

I could use `loaders` from the beginning, but now I think it gets more relevant as we have 2 sources of data. Also, if `loader` API doesn't work for us we could provide out own abstraction for providing data from the web, which probably would be another version of wrapper for `App`.

Let me know what you think about this approach or if anything is unclear

AgniveshChaubey:
Was reading about `loaders` and yes, it does make sense to utilize the loader for loading the data. However, can you further explain what do you mean by -**_could have just one route for all the routes that require report data from web_** 

Yes, It's better to fully utilize the feature of  resuibility of components. Will create a wrapper component that will render either `file drop area` or the`App` component

harrel56:
I had in mind that we could simplify routes configuration by having more generic routing, like:
```ts
createHashRouter([
  {
    path: "/:draftName",
    element: <App .../>,
    loader: ({ params }) => {
      return someFetchingFunction(params.draftName);
    },
  },
]);
```

AgniveshChaubey:
tried this loader function -
```js
    {
      path: "/:draftName",
      element: <ReportDataHandler />,
      loader: ({ params }) => {
        console.log(params.draftName);
        return fetch(
          `https://bowtie-json-schema.github.io/bowtie/${params.draftName}.jsonl`
        )
          .then((response) => response.text())
          .then((jsonl) => {
            const dataObjectsArray = jsonl.trim().split(/\n(?=\{)/);
            const lines = dataObjectsArray.map((line) => JSON.parse(line));
            return lines;
          });
      },
    },
```
Even though the draftName variable correctly reflects the draft name in the console, why is it resolved as **undefined** in the URL?
getting this error `GET https://bowtie-json-schema.github.io/bowtie/undefined.jsonl 404`

harrel56:
Are you sure the request call doesn't come from somewhere else in the code? Because if it prints it right it should also send request to a valid URL. If you still won't be able to fix it, you could commit your changes and then I could have a look

AgniveshChaubey:
Hmm, the call was from another component! 😐

AgniveshChaubey:
Yes, the call was from another component! 😐

Julian:
(Merged #301, so the conflicts are from there, sorry, should be minor to fix)

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,144 @@
...
+  const [fileUploaded, setFileUploaded] = useState(false);
+  const [invalidJson, setInvalidJson] = useState(false);
+  const [lines, setLines] = useState();
+  const inputRef = useRef(null);
+
+  const flyingAnimation = useSpring({
+    transform: fileUploaded ? "translateY(-100%)" : "translateY(0%)",
+  });
+
+  const handleDrag = function (e) {
```
harrel56, line 19: Why not split it to 2 different functions if you need to check for event type inside?

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,144 @@
...
+    inputRef.current.click();
+  };
+
+  const handleFiles = (file) => {
+    const fileName = file.name;
+    if (fileName.endsWith(".json") || fileName.endsWith(".jsonl")) {
+      const reader = new FileReader();
+      reader.onload = (e) => {
+        try {
+          const dataObjectsArray = e.target.result.trim().split(/\n(?=\{)/);
```
harrel56, line 61: Is there a reason for searching for a newline and `{` on next position? Can we just change this regex to `\n`?

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,144 @@
...
+          setLines(dataObjectsArray.map((line) => JSON.parse(line)));
+        } catch (error) {
+          console.error(" :", error);
```
harrel56, line 64: I guess this should also show some kind of visual indication that something went wrong

frontend/src/components/DragAndDrop/DragAndDrop.css
```diff
@@ -0,0 +1,29 @@
+#form-file-upload {
```
harrel56, line 1: I think it's generally more preferred to apply styles to classes instead of ids, but that's not a big deal

frontend/src/index.js
```diff
@@ -2,37 +2,31 @@ import "bootstrap/dist/css/bootstrap.min.css";
...
 
 import { createRoot } from "react-dom/client";
-import App from "./App";
+import ReportDataHandler from "./ReportDataHandler";
 import { createHashRouter, RouterProvider } from "react-router-dom";
+import DragAndDrop from "./components/DragAndDrop/DragAndDrop";
 
 const router = createHashRouter([
   {
     path: "/",
```
harrel56, line 11: I would say root path should still point to newest draft version. Updating local report is rather unusual action for the user, but I might be wrong 😄

Julian: Perhaps coming around to this as well, so sorry for misleading you a bit @AgniveshChaubey -- let's maybe go with this still then (redirecting to latest), and we can think about how we "automatically" make it a bit more prominent when someone's using the UI offline (i.e. locally) later.

frontend/src/ReportDataHandler.js
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from "react";
+import { useLoaderData } from "react-router";
+
+import App from "./App";
+import LoadingAnimation from "./components/LoadingAnimation";
+
+const ReportDataHandler = () => {
+  const [loading, setLoading] = useState(true);
+  let [lines, setLines] = useState([]);
```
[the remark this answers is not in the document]
AgniveshChaubey: Yes, I think it's unnecessary to set again the same thing!

frontend/src/ReportDataHandler.js
```diff
@@ -0,0 +1,22 @@
...
+  const loaderData = useLoaderData("draftName");
+
+  useEffect(() => {
+    setLoading(true);
+    setLines(loaderData);
+
+    setLoading(false);
+  }, [loaderData]);
+
+  return <>{loading ? <LoadingAnimation /> : <App lines={lines} />}</>;
```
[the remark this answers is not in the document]
AgniveshChaubey: We can't directly return JSX.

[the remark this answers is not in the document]
harrel56: What do you mean? React components should mainly return JSX. It could be something like this: `loading ? <LoadingAnimation /> : <App lines={lines} />` or even:
```js
if (loading) {
  return <LoadingAnimation/>
}
return <App ../>

```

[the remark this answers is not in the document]
harrel56: And BTW it looks like the loading screen only works on initial component mount. Maybe this hook can be useful for implementing loading screens? https://reactrouter.com/en/main/hooks/use-navigation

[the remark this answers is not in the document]
AgniveshChaubey: > What do you mean?

Oops, sorry for the misspelling! I meant we can't use `javascript` directly in return statement that's why used Fragment.

> `loading ? <LoadingAnimation /> : <App lines={lines} />`

Oh! I see, I was using an extra return statement!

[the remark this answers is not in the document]
AgniveshChaubey: > And BTW it looks like the loading screen only works on initial component mount...

Yes, working on it.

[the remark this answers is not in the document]
AgniveshChaubey: > What do you mean?....

Sorry for the misspelling! I meant we can't directly use `return { ... }`. It would be considered an object.

> `loading ? <LoadingAnimation /> : <App lines={lines} />`

Oh! I see, you meant to remove `{...}` too and not just the fragment.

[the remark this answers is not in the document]
AgniveshChaubey: > And BTW it looks like the loading screen only works on initial component mount.....

Fixed

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: Yes, you're right. If the data is valid line-based json, then we can skip that extra search.

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: Bootstrap's `toast` element added!

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: If split, then three separate attributes will be required to be added to the form element!
BTW done!

frontend/src/ReportDataHandler.js
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from "react";
+import { useLoaderData } from "react-router";
+
+import App from "./App";
+import LoadingAnimation from "./components/LoadingAnimation";
+
+const ReportDataHandler = () => {
+  const [loading, setLoading] = useState(true);
+  let [lines, setLines] = useState([]);
+  const loaderData = useLoaderData("draftName");
```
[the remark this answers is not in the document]
AgniveshChaubey: fixed! 
BTW it was not causing any errors. Maybe because it's referring to the same route?

frontend/src/index.js
AgniveshChaubey: Yes, I agree with Harrel here too!

AgniveshChaubey:
used Context APIs to make the data available centrally!

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,196 @@
...
+  const handleFiles = (file) => {
+    const fileName = file.name;
+    if (fileName.endsWith(".json") || fileName.endsWith(".jsonl")) {
+      const reader = new FileReader();
+      reader.onload = (e) => {
+        try {
+          const dataObjectsArray = e.target.result.trim().split(/\r?\n/);
+          setLines(dataObjectsArray.map((line) => JSON.parse(line)));
+        } catch (error) {
+          handleShowToast();
```
harrel56, line 80: I don't understand why there are 2 representations of errors? Here you use toast, but if uploaded file has wrong extension then it looks completely different. Can we use just one approach for representing errors?

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,196 @@
...
+                  <p className="card-text">
+                    Drag and drop the JSON report file generated by Bowtie in
+                    your local repository!
+                  </p>
+                  <button className="btn btn-primary" onClick={onButtonClick}>
+                    Upload report
+                  </button>
+                  <h5
+                    className={`pt-3 text-danger ${
+                      !invalidJson ? "d-none" : ""
```
harrel56, line 181: Maybe it'll be more readable if instead of conditionally adding css class, you would conditionally render whole `<h5>` tag?

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,196 @@
...
+                onChange={handleChange}
+              />
+
+              <div
+                id="label-file-upload"
+                className={dragActive ? "drag-active" : ""}
+                style={{ backgroundColor: `${invalidJson ? "#f00b0b39" : ""}` }}
+              >
+                <div className="text-center">
+                  <animated.div style={flyingAnimation}>
```
harrel56, line 169: Is this animation necessary? I think I don't even know how to trigger it - maybe adding a whole separate library for that is not worth it?

frontend/src/context/ThemeContext.js
```diff
@@ -0,0 +1,26 @@
+import { createContext, useState, useEffect } from "react";
+
+export const ThemeContext = createContext();
+
+const ThemeContextProvider = ({ children }) => {
+  const [theme, setTheme] = useState(
```
harrel56, line 6: Putting this in context is really nice. But `theme` is basically a boolean variable so maybe something like `[isDarkMode, setDarkMode]` will be more appropriate?

frontend/src/context/ThemeContext.js
```diff
@@ -0,0 +1,26 @@
...
+  );
+
+  useEffect(() => {
+    document
+      .querySelector("body")
+      .setAttribute("data-bs-theme", theme ? "dark" : "light");
+  }, [theme]);
+
+  const handleTheme = () => {
+    setTheme(!theme);
```
harrel56, line 17: I would also rename `handleTheme` to something like `toggleDarkTheme`? You could also use another variant of setting state: `setDarkMode(mode => !mode)` - this will ensure that it always toggles appropriately even if it is called multiple times in a single update batch - but that's not a big deal though

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: sure!

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: Okay then! gonna remove it.

frontend/src/context/ThemeContext.js
AgniveshChaubey: done

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: `not have enough info` has not been handled yet. Though it's having very less chance of occurrence.

Julian: I haven't tried to preview what this looks like yet, so I could be off base, but definitely don't be checking the extension of the file -- you don't know what it will be, someone can redirect into a file of any name.

And I don't think it's very different if the file isn't valid line-delimited JSON or if it's somehow otherwise invalid -- these files are internal implementation details of Bowtie (i.e. the contents of them aren't really hand authored by anyone, Bowtie produces them and Bowtie then consumes them) -- so an invalid file basically only has one explanation, namely "you're uploading something that wasn't produced by Bowtie" -- if anything else is happening it's likely a Bowtie bug.

frontend/src/index.js
Julian: > If I am not wrong, you mean rendering the upload report component at / route when users are using UI locally?
However, the option to upload a report will still be there in navbar's dropdown list.

Yes -- in other words if we detect that the base URL is `https://localhost` (or the reverse, not `https://bowtie-...`) then maybe we skip showing any dialects and only show the local upload. This will probably be more obvious as part of #296 because the obvious way to implement that I think will *force* us to do something different for the local case.

frontend/src/ReportDataHandler.js
```diff
@@ -0,0 +1,18 @@
...
+import App from "./App";
+import LoadingAnimation from "./components/LoadingAnimation";
+
+const ReportDataHandler = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const loaderData = useLoaderData();
+
+  useEffect(() => {
+    setIsLoading(!isLoading);
+  }, [useNavigation().state === "loading"]);
```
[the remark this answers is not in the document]
AgniveshChaubey: In this case, `useNavigation` will only change its state when there is a path change. And for that case, we will require to trigger the useEffect. So I don't think there any problem here.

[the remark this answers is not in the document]
harrel56: But there's no need to use `useEffect`, you just need simple variable assignment:
```js
const isLoading = navigationData.state === 'loading'
```
If simplified it could look like this:
```js
const ReportDataHandler = () => {
  const loaderData = useLoaderData()
  const {state} = useNavigation()

  return state === 'loading' ? <LoadingAnimation /> : <App lines={loaderData} />
}
```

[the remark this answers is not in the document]
AgniveshChaubey: Hmm! got it!

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: > I haven't tried to preview what this looks like yet, so I could be off base, but definitely don't be checking the extension of the file -- you don't know what it will be, someone can redirect into a file of any name (I regularly redirect into some file called output with no extension).

Hmm, I didn't think that way.

So to check the validity of the report, I think we must be checking if it contains at least three json lines (one containing `started` property, one containing `case` property and one containing `implementation` property). And if any of these is missing then we can consider it as invalid.

Julian: Right, don't check any of that explicitly though -- simply start parsing the file assuming it is valid -- if you get an error, the file is invalid (this is essentially what the Python code does) and then you report to the user "hey you gave me something other than a Bowtie report".

frontend/src/index.js
Julian: Actually an even better way of doing the above might just be trying to fetch the dialect file and whenever it fails falling back to showing the "upload report" widget.

AgniveshChaubey: I think checking for `localhost` would be good option because the fetching process is being done in loader function and the fetching URL itself is dependent on route `params` and also its corresponding component is dependent on the status of loader function!

AgniveshChaubey: But ya, if you want that way, we still can achieve that by redirecting to `upload report` component from the loader function itself if it fails fetching the data. But in that case we'll have to compromise with the URL (as currently `<ReportDatahandler />` is being rendered at `/` route)

AgniveshChaubey:
Added the conditional rendering of components (`<ReportData />` and `<DragAndDrop />`) based on the URL's hostname.

Most of the issues related to this specific PR have been resolved, and I think it is now ready to merge!

harrel56 (approved):

frontend/src/components/DragAndDrop/DragAndDrop.js
```diff
@@ -0,0 +1,146 @@
...
+    <div>
+      {lines ? (
+        <App lines={lines} />
+      ) : (
+        <div>
+          <NavBar />
+          <div
+            aria-live="polite"
+            aria-atomic="true"
+            style={{
```
Julian, line 78: (As usual not being a React or even frontend dev) -- is there a reason you use inline styles here but still have a separate CSS file?

Julian (approved):
One minor comment/question, after that feel free to merge!

frontend/src/components/DragAndDrop/DragAndDrop.js
AgniveshChaubey: Actually, this is not even required anymore! Removed it!

PR #301, opened 2023-06-18
Change report files to .json

harrel56:
Also changed regex for split function, it should handle Windows new lines as well

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           name: report
           path: |
-            ${{ matrix.version }}.jsonl
+            ${{ matrix.version }}.json
             badges/
 
   deploy-frontend:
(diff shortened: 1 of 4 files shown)


harrel56:
I guess it makes local development a little harder because all the fetches will fail. Can be fixed by manually adding report data to `frontend/public` folder

Julian:
Nice, thanks!

> Can be fixed by manually adding report data to frontend/public folder

We could also have two separate env vars / base URIs -- i.e. be able to separately specify a base URI for the site and one for where reports live (and then when you develop locally you still use the "production" URI for the latter, but if you don't provide one we default to using the same URI for both).

But I honestly haven't tried developing on the UI yet personally so I trust the two of you more than myself :D

Regardless, given you mentioned this seems to fix the issue, going to merge, thanks again!

PR #305, opened 2023-06-23
Refactor/switch from create-react-app to vite

harrel56:
As CRA is no longer recommended for developing react apps, I thought I can give Vite a try. This change at least clears some of the vulnerabilities which came with `react-scripts`. Also it looks like it builds bundles faster :)
I had to rename each .js file which uses JSX to .jsx, which I didn't mind because it seems to be more correct


Julian:
Very nice, looks great to me, thanks!

PR #306, opened 2023-06-23
Feature/add eslint

harrel56:
PR on top of https://github.com/bowtie-json-schema/bowtie/pull/305.

Adds eslint to build workflow, few fixes were required


.github/workflows/build-frontend.yml
```diff
@@ -18,6 +18,8 @@ jobs:
           node-version: 18.x
       - run: npm ci
         working-directory: ./frontend
+      - run: npm run lint
```
Julian, line 21: I think even if it's a bit less like what's "normal" for fully JS projects that it's better here to run eslint via `pre-commit`, at least in CI -- speciifically via adding [the eslint hook](https://github.com/pre-commit/mirrors-eslint) to the pre-commit-config.yaml.

Otherwise though looks good to me.

frontend/src/data/runInfo.js
```diff
@@ -1,14 +1,5 @@
 import Summary from "./summary";
 
-const _DIALECT_URI_TO_SHORTNAME = {
-  "https://json-schema.org/draft/2020-12/schema": "Draft 2020-12",
```
Julian, line 4: I'm guessing eslint flagged this as not being used right?

But I think that's probably actually incorrect in the code -- right now if you go to the homepage you see `Bowtie-draft202012` as the `<title>` but I think this mapping is/was meant to convert that into some more human readable name (even if it's only slightly different)

.github/workflows/build-frontend.yml
harrel56: Alright, now it does linting in pre-commit :)

Julian: Hooray! LGTM (merge at will) -- the failures I think are dumb rate limiting to GitHub's own runners, so it may require kicking the build again or something, but yeah changewise, lgtm.

PR #307, opened 2023-06-24
Refactor/routing refactor

harrel56:
PR on top of https://github.com/bowtie-json-schema/bowtie/pull/305 & https://github.com/bowtie-json-schema/bowtie/pull/306.

There were few doubts how to handle routing on local dev environment and this is my proposition. I would advocate to keep differences between local <-> GH pages as minimal as possible - so now the only thing that is different is fetch URI. Let me know what you think.

Apart from minor fixes (my local sonarlint complains a lot :)), I also fixed loading screen if switching between `/local-draft` -> `/:draftName`


frontend/src/components/Summary/SummaryTable.jsx
```diff
@@ -57,15 +56,7 @@ const SummaryTable = ({ lines }) => {
       }
       let seq = element.seq;
       const caseImplementation = caseArray.find((each) => each.seq === seq);
-      if (
-        element.caught === true ||
-        element.caught === false ||
```
AgniveshChaubey, line 62: Here we can't ignore the `caught` property as it also shows _**errored**_ cases.
To make the expression simple, maybe we can do something like `element.caught !== null`

frontend/src/context/BowtieVersionContext.jsx
```diff
@@ -0,0 +1,14 @@
+import { createContext, useMemo, useState } from "react";
+
+export const BowtieVersionContext = createContext();
+
+export const BowtieVersionContextProvider = ({ children }) => {
+  const [version, setVersion] = useState();
```
AgniveshChaubey, line 6: Instead of creating a specific Context Provider for `version`, I was planning something like creating a general Context Provider for the entire `data`, in order to avoid prop drilling.

frontend/src/data/runInfo.js
```diff
@@ -1,14 +1,5 @@
 import Summary from "./summary";
 
-const _DIALECT_URI_TO_SHORTNAME = {
-  "https://json-schema.org/draft/2020-12/schema": "Draft 2020-12",
```
AgniveshChaubey, line 4: I actually missed reviewing this file when submitting my own PR. This code represents my previous attempts at structuring the data. As Julian pointed out, it would be beneficial to map the draft names to more readable and formatted ones. We can modify the code accordingly. No worries, I'll handle it in a separate PR.

frontend/package.json
```diff
@@ -2,31 +2,25 @@
   "name": "bowtie",
   "version": "1.0.0",
   "description": "The UI displays the outcomes generated by Bowtie, which serves as a meta-validator for the JSON Schema specification",
-  "homepage": "/bowtie",
   "scripts": {
```
AgniveshChaubey, line 5: Don't we need to add the `homepage` or is it something related to _**vite**_?

frontend/src/MainContainer.jsx
```diff
@@ -0,0 +1,14 @@
+import NavBar from "./components/NavBar";
+import { useNavigation } from "react-router-dom";
+import LoadingAnimation from "./components/LoadingAnimation";
+
+export const MainContainer = ({ children }) => {
+  const { state } = useNavigation();
+
+  return (
+    <div>
+      <NavBar />
```
AgniveshChaubey, line 10: It's nice to move the `NavBar` to the top-level component, but I think here bowtie_version will not be rendered. As we have created a context provider for the version, we can utilize it here.

AgniveshChaubey (commented):
Hi @harrel56, left a few comments, else everything looks fine to me.

frontend/src/data/runInfo.js
harrel56: I missed your comment and already added the name mapping, hope you didn't start changing it on your end 😄

AgniveshChaubey (approved):

frontend/src/context/BowtieVersionContext.jsx
harrel56: Hmm, I needed access to version and couldn't use `useLoaderData` so I put it in context. If you think more general context is needed you are free to refactor this one in your later PRs, but honestly I don't know if putting everything in context is a good solution. But if you have any other ideas how to handle version i nav bar, please share

frontend/src/MainContainer.jsx
harrel56: What do you mean? I access version context in `NavBar` component

frontend/src/components/Summary/ImplementationRow.jsx
```diff
@@ -70,11 +70,7 @@ const ImplementationRow = ({ lines, implementation, index }) => {
         if (
           element.caught === true ||
           element.caught === false ||
-          (element.results &&
-            element.results.every(
-              (each) =>
-                typeof each === "object" && each.hasOwnProperty("errored"),
```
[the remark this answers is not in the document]
harrel56: I rewrote it to avoid using `typeof` & `hasOwnProperty`. But I was not sure how the data exactly looks like so I might've done it wrong. This should evaluate to `true` if all items in a list have field `errored` equal to `true`, right?

frontend/src/components/Summary/SummaryTable.jsx
harrel56: Hmm, I guess it's not as simple as I thought. Do you know what does `caught` signifies? When it's set to `true` and when to `false`?

frontend/package.json
harrel56: Setting homepage was a thing for `create-react-app`, it indicated on which path the app was hosted. Now this part of config was moved to `vite.config.js` file

frontend/src/MainContainer.jsx
AgniveshChaubey: Oops! Sorry, I overlooked it. :)

frontend/src/components/Summary/ImplementationRow.jsx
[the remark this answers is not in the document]
AgniveshChaubey: Yes, it's totally correct!

frontend/src/components/Summary/SummaryTable.jsx
AgniveshChaubey: I am not 100% sure, but I guess when tests are running and get failed due to some reason, then the `caught` sets to `false`, and if some error occurred and the tests didn't run due to some reason, then the `caught` sets to `true`. 

I have gone through the data of all the drafts and built the logic based on that. I might have taken brute force approach, but I can surely say we do require both conditions to be considered (i.e. `caught` and `errored`). Skipping `caught === false` may not cause any issue in the latest draft, but it indeed misses out on some data in older drafts.(`draft3`, `draft4`)

However, @Julian, can you please confirm?

harrel56: Yeah, I looked briefly into the generated reports and now I kind of get it. Bottom line is that you're right and the `caught` field should be checked. And also this change is from previous PR and is already in main - gonna fix it with null check you suggested

PR #311, opened 2023-06-28
Refactor/transform data once

harrel56:
I wanted to separate presentation and data parsing logic. General idea was to parse report data to a structure which could be used during presentation more easily and without any additional computation (like iterating over all lines of a report). 
I tried to limit the scope of changes to a minimum, but well, it was hard to split it to smaller chunks and the changes turned up to be broader than I would like.

Brief list of changes:
- as the structures (report json lines and also the data for presentation) were really complicated and I could hardly understand what was going on - I introduced a little bit of typescript to describe this data.
- all parsing data logic is in `parseReportData.ts` file (well, a little more processing was required in details modal) - it's ugly but at least mostly in one place :)
- `DetailsButtonModal.jsx` & `RunTimeInfoModal.jsx` have been changed to not use css based bootstrap logic - generally, I think this approach is more efficient: no HTML duplication and it's only rendered on click.

Let me know what you think.

As a side note: I was expecting to improve loading/rendering performance a bit, but there not much noticeable difference. It turns out that the real bottleneck is rendering almost 400 accordions with tables :)


AgniveshChaubey:
Wow, this PR looks significant!
As I was mainly focused on getting the UI up as soon as possible, there were many instances where I repeated the code for data processing in several components. And yes, separating the data layer and presentation layer was necessary to address this issue. I'll review it soon and let you know if anything is missing or can be further improved.

frontend/src/data/parseReportData.ts
```diff
@@ -0,0 +1,192 @@
...
+          (res, idx) => {
+            if (res.errored) {
+              const errorMessage = res.context?.message ?? res.context?.stderr;
+              implementationData.erroredTests++;
+              return { state: "errored", message: errorMessage };
+            } else if (res.skipped) {
+              implementationData.skippedTests++;
+              return { state: "skipped", message: res.message };
+            } else {
+              const successful = res.valid === caseData.tests[idx].valid;
```
AgniveshChaubey, line 63: If `successful` resolves to true, then we have two cases...
- _**successfully ran**_ and its value is `true` (which is represented by `check` svg)
- _**successfully ran**_ and its value is `false` (which is represented by `cross` SVG)

frontend/src/data/parseReportData.ts
```diff
@@ -0,0 +1,192 @@
...
+  cases: Map<number, CaseResult[]>;
+  erroredCases: number;
+  skippedTests: number;
+  unsuccessfulTests: number;
+  erroredTests: number;
+}
+
+interface CaseResult {
+  state:
+    | "successful"
```
AgniveshChaubey, line 148: Changes required (in reference to the above comment)

AgniveshChaubey (commented):
Hi, everything looks good, just a minor correction is needed. left some comments regarding those.

frontend/src/components/Cases/AccordionSvg.jsx
```diff
@@ -4,43 +4,14 @@ import {
...
-      svgComponent = <CheckCircleFill />;
-      break;
-    case "unexpectedlyInvalid":
-      svgComponent = <XCircleFill />;
-      break;
-    default:
-      return null;
+const AccordionSvg = ({ result }) => {
+  if (result.state === "skipped" || result.state === "errored") {
+    return <td className='text-center text-bg-warning'><ExclamationOctagon /></td>
```
AgniveshChaubey:
```html
return <td className=`text-center ${result.state === "skipped" ? "text-bg-warning" : "text-bg-danger"}`><ExclamationOctagon /></td>
```

AgniveshChaubey (approved):
One last change, then I think we are good to go!

frontend/src/components/Cases/AccordionSvg.jsx
harrel56: Sorry, I didn't see you last comments. I believe the background for skipped and errored cases were the same? But still, I agree with you that it should look like differently, so changed it. Added also a title to provide even more information (when I was using the site I felt like I don't really know what these colors and icons mean :))

Julian:
Think this looks good to me too! Thanks!

Issue #303, opened 2023-06-20
Add the networknt implementation - Java


Relequestual:
Previously, the networknt Java implementation reported that it supported 2019-09 and 2020-12, however the support is so weak that the support indicator on the official website listing was [removed](https://github.com/json-schema-org/json-schema-org.github.io/pull/525)!

While the implementation provides its own compliance report, it would be nice if it could be added here.

Logging an issue as requested.

Julian:
We have at least one more thing to do here, namely the harness doesn't yet support registries apparently (thanks @harrel56 for pointing that out).

PR #316, opened 2023-07-03
Add support - networknt/json-schema-validator

AgniveshChaubey:
This PR addresses #303.


.github/dependabot.yml
```diff
@@ -73,6 +73,11 @@ updates:
     schedule:
       interval: "daily"
 
+  - package-ecosystem: "gradle"
+    directory: "implementations/java-json-schema-validator"
```
Julian, line 77: The directory here is named differently than the real one, so this will fail to run I think -- let's actually keep this as is and rename the directory.

.github/dependabot.yml
```diff
@@ -163,3 +168,8 @@ updates:
     directory: "/tests/fauxmplementations/lintsonschema"
     schedule:
       interval: "daily"
+
+  - package-ecosystem: "docker"
+    directory: "/implementations/java-json-schema-validator"
```
Julian, line 173: Same here.

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+    DialectRequest dialectRequest = objectMapper.treeToValue(
+      node,
+      DialectRequest.class
+    );
+    if (DIALECTS.contains(dialectRequest.dialect())) {
+      output.println(
+        objectMapper.writeValueAsString(new DialectResponse(true))
+      );
+    } else {
+      output.println(
```
Julian, line 103: I'm not sure if you copied this from @harrel56 's harness (and am too lazy to look at the minute :) but technically this would be a Bowtie bug to ever reach this -- `{"ok": false}` isn't meant to mean "I don't understand that dialect", it's meant to mean "I don't support setting dialects at all outside of a schema" -- Bowtie sending you (the harness) a dialect request that isn't one of the ones supported by the implementation (i.e. that you responded with in `start`) would just be a bug, that should never happen, and would seemingly be the only way one could get here in the harness.

Julian (commented):
Awesome, well done, looks great!

I haven't tried this out yet (so have just read it) and @harrel56 might have some Java-specific suggestions for the harness, but other than the comments I left and me having to test it by building it sometime on Wednesday probably after the US holiday, looks good to me!

implementations/java-jsonschema-validator/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+
+def jsonProvider = [
+        group: 'com.fasterxml.jackson.core',
+        name: 'jackson-databind',
+        version: '2.15.2'
+]
+
+dependencies {
+    implementation "$harnessImplementation.group:$harnessImplementation.name:$harnessImplementation.version"
+    implementation "$jsonProvider.group:$jsonProvider.name:$jsonProvider.version"
```
harrel56, line 33: This whole "json provider" dependency probably should be removed, because the "networknt" by itself depends on other json libraries (I mean, it just happens that it also uses `jackson` libraries) and all the transitive dependencies will be handled automatically. 
So, the line 
```groovy
implementation "$harnessImplementation.group:$harnessImplementation.name:$harnessImplementation.version"
```
will be enough

implementations/java-jsonschema-validator/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+
+jar {
+    archiveBaseName = 'harness'
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+    from {
+        configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
+    }
+    manifest {
+        attributes "Main-Class": "BowtieJsonSchemaValidator"
+        attributes "Implementation-Group": harnessImplementation.group
```
harrel56, line 44: This is not read in harness runtime so it could be removed

implementations/java-jsonschema-validator/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+        attributes "Implementation-Name": harnessImplementation.name
+        attributes "Implementation-Version": harnessImplementation.version
+        attributes "Provider-Group": jsonProvider.group
+        attributes "Provider-Name": jsonProvider.name
+        attributes "Provider-Version": jsonProvider.version
```
harrel56, line 49: As mentioned above, these "Provider-*" attributes are unnecessary and can be removed

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+import com.networknt.schema.JsonSchemaFactory;
+import com.networknt.schema.ValidationMessage;
+import java.io.*;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+
+public class BowtieJsonSchemaValidator {
```
harrel56, line 16: The name of the file should be capitalized (pascal case). You even refer to it with capitalized name in gradle script - I don't even know how it finds this file, honestly 😄

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
harrel56: So I got it incorrectly, gonna need to adjust it in my harness too

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+  private static final List<String> DIALECTS =
+      List.of("https://json-schema.org/draft/2020-12/schema",
+              "https://json-schema.org/draft/2019-09/schema",
+              "http://json-schema.org/draft-07/schema#",
+              "http://json-schema.org/draft-06/schema#",
+              "http://json-schema.org/draft-04/schema#");
+
+  private final ObjectMapper objectMapper = new ObjectMapper().configure(
+      DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+  private final JsonSchemaFactory schemaFactory =
```
harrel56, line 27: This field is unused

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.networknt.schema.JsonSchema;
+import com.networknt.schema.JsonSchemaFactory;
+import com.networknt.schema.ValidationMessage;
+import java.io.*;
+import java.util.List;
+import java.util.Map;
```
harrel56, line 11: This import is unused

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.Set;
+import java.util.jar.Attributes;
```
harrel56, line 13: This import is unused

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+        .tests()
+        .stream()
+        .map(test -> {
+          JsonSchemaFactory schemaFactory = JsonSchemaFactory.getInstance();
+          JsonSchema schema = schemaFactory.getSchema(
+            runRequest.testCase().schema()
+          );
+
+          ObjectMapper objectMapper = new ObjectMapper();
+          JsonNode instanceJson = test.instance();
```
harrel56, line 129: These 2 local variables are unused

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+    output.println(objectMapper.writeValueAsString(startResponse));
+  }
+
+  private void dialect(JsonNode node) throws JsonProcessingException {
+    DialectRequest dialectRequest = objectMapper.treeToValue(
+      node,
+      DialectRequest.class
+    );
+    if (DIALECTS.contains(dialectRequest.dialect())) {
+      output.println(
```
harrel56, line 99: When setting specific dialect you need to perform some action on the validator. It looks like you should call e.g. `JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4)`, but ofc with specific dialect

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
```diff
@@ -0,0 +1,215 @@
...
+      );
+    }
+  }
+
+  private void run(JsonNode node) throws JsonProcessingException {
+    RunRequest runRequest = objectMapper.treeToValue(node, RunRequest.class);
+    try {
+      if (runRequest.testCase().registry() != null) {
+        JsonSchemaFactory schemaFactory = JsonSchemaFactory.getInstance();
+        JsonSchema schema = schemaFactory.getSchema(
```
harrel56, line 114: Does it do something? It should somehow register all schema from `runRequest.testCase().registry()`, but unfortunately I don't know the API of networknt lib

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
Julian: (Lemme know if there's some better place to make that even more explicit!)

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
Julian: Does the linter not catch unused imports + variables? If not can you suggest how we configure it to? Or are we not running one, I forget.

harrel56: It seems there is no java linter supported in pre-commit? I'm not familiar with this tool tbh. Usually you would use `sonarlint` or `checkstyle` with java

Julian: https://github.com/gherynos/pre-commit-java it looks like provides a hook for `checkstyle` at least.

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
AgniveshChaubey: That's strange! It keeps automatically changing to camelCase. Is there a git command I might have executed incorrectly? I only did one reset, and since then, it has been automatically converting to camelCase.

Julian: Are you on macOS?

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
AgniveshChaubey: Yes, the `getSchema` method is responsible for collecting the schema that needs to be validated. [This ](https://github.com/networknt/json-schema-validator/blob/master/doc/quickstart.md)is the reference I have used.

AgniveshChaubey: Yes, the `getSchema` method is responsible for collecting the schema that needs to be validated. [This ](https://github.com/networknt/json-schema-validator/blob/master/doc/quickstart.md)is the reference I have used to for this implementation.

AgniveshChaubey: Thank you for taking a close look at the PR! It's quite late today, so I will address your comments and resolve/respond to them tomorrow.

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
harrel56: nice - I could add it later if you don't think it's not strictly needed in this PR

Julian: Yep fine with me

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
harrel56: Sure thing, there is no rush

harrel56: So, if I recall correctly the structure returned from `runRequest.testCase().registry()` is something like this:
```json
{
  "https://host.com/schema1": { ...here goes actual schema definition },
  "https://host.com/schema2": { ...here goes actual schema definition },
  ...
}
```
so you would need to iterate over all fields in this JSON and call this `getSchema()` with value of each field.

But I have other concern that `getSchema` only creates Schema object and it does not store it somewhere in its internal registry (but I might be wrong, haven't had time to thoroughly check it). Maybe implementing custom fetcher is a way to go? https://github.com/networknt/json-schema-validator/blob/master/doc/cust-fetcher.md
In this fetcher you would just serve schemas from beforementioned registry json object

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
AgniveshChaubey: Windows

AgniveshChaubey: Windows

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
AgniveshChaubey: fixed

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
AgniveshChaubey: I think we should maintain the `dialect` method exclusively for configuring the dialects without the `$schema` property. All actions associated with validation should be added in the `run` method.

implementations/java-jsonschema-validator/bowtieJsonSchemaValidator.java
AgniveshChaubey: used ArrayList to store the results

AgniveshChaubey:
Hi, I apologize for the delay. I was busy with some academic tasks. I have just updated the PR and made an effort to address all the comments. Hopefully, this should work fine. Whenever you have some time, please take a look at it, @harrel56  and @Julian

Julian:
No worries! Will have a look today

Julian:
This looks great! I want to see what it looks like and still haven't figured out why GitHub actions runners seem to lose their docker images after building them, so going to merge -- if there are other comments we can address in a follow-up I think, but yeah well done!

PR #325, opened 2023-07-08
Hyperjump - Fix test harness issue

jdesrosiers:
I noticed today that bowtie was reporting that I had some failing tests. I looked into it and it turned out to be a test harness issue. I couldn't manage to get bowtie to recognize and run my changes locally, so I wasn't able to make sure the fix works as intended. It looks like it should work :man_shrugging:.

### DIFF
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -116,10 +116,9 @@ const cmds = {
       });
     } else {
       try {
-        const fakeURI =
-          "https://example.com/bowtie.sent.schema." +
-          args.seq.toString() +
-          ".json";
+        const idToken = args.dialect === "http://json-schema.org/draft-04/schema#" ? "id" : "$id";
+        const host = testCase.schema?.[idToken].startsWith("file:") ? "file://" : "https://example.com";
+        const fakeURI = `${host}/bowtie.sent.schema.${args.seq.toString()}.json`;
         addSchema(testCase.schema, fakeURI, dialect);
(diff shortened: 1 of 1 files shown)


Julian:
Thanks -- can you explain the change, I'm not sure I follow -- this looks like what's happening is that schemas that contain internal IDs can't be given an external ID with some other scheme?

jdesrosiers:
Good guess, but not quite. I don't allow references to a `file:` scheme unless the base URI also uses the `file:` scheme. This is a security precaution I borrowed from web browsers.

In my implementation, `file:` URIs are how you reference a schema on the file system. Technically, a user could load a schema into local storage with a `file:` URI, but I always treat `file:` URIs as a file-based schema because I don't always know where it came it from. Likewise, an `http(s)` schema may have been retrieved from the web and I don't always know in all contexts where the schema came from, so from a security point of view, I have to assume that it came from the web.

Therefore, if I download a schema (say from schemastore.org), that schema should not be allowed to make a reference to a schema on my filesystem. That would either be a mistake or an attack attempt of some sort. The other way around is allowed. From a file-based schema, a non-filesystem reference is allowed, but that non-file-based schema can't reference the filesystem even though the original context was file-based.

jdesrosiers:
In this case, the retrieval URI uses `https`. Then the `file:` URI declared in `$id` is a change from `https:` to `file:` which is not allowed. Using `file:` for the retrieval URI if the `$id` uses `file:` fixes the problem.

Julian:
Ah great thanks just wanted to be sure this wasn't going to be the same story with some third scheme. Thanks for the fix.

PR #326, opened 2023-07-08
fix: bug causing error when uploading local report

AgniveshChaubey:


Julian:
Great. Thanks! Didn't try it out but I assume you did, if it fixes things looks good to merge from reading.

harrel56:
Nice, must have missed it, thanks

PR #334, opened 2023-07-10
chore:  adjust dev.harrel:json-schema

harrel56:
Needed to adjust harness a little with new release, also:
- disable skipping vocabulary tests,
- fix response to dialect message


Julian:
Thanks!

PR #336, opened 2023-07-11
feature/add java linter

harrel56:


harrel56:
@Julian so, GH action `CI / pre-commit` is running fine, but the `pre-commit.ci - pr` is failing due to docker not being found. Will appreciate any help. 
https://github.com/gherynos/pre-commit-java explicitly states that it requires docker - tried to find some alternatives, but found only PMD static analysis tool (not familiar with that) which states to require `Courier` (no idea what's that 🥲 ) and I suppose it will fail in similar way.

Julian:
Bleh fun ok let me have a look, yeah, `pre-commit.ci` has fewer capabilities around installing additional dependencies (in exchange for being tons faster and automatically committing fixes).

Julian:
The other thing is `clang-format` and the new linter are fighting right? It looks like `clang-format` is failing expecting different formatting -- for that probably we should simply disable `clang-format` on Java files right, it's probably duplicating what this linter is doing or could be doing?

harrel56:
Hmm, they are fighting, yes, but they are doing two different things. I'm pretty sure checkstyle cannot format code, it only gives feedback about what's wrong

Julian:
So the (style) changes here were done manually? Can we configure checkstyle to accept the style `clang-format` is asking for, or vice versa?

harrel56:
I think if the "pre-commit.ci" would commit changes from formatter then all would be fine. There were few conflicting rules though, but I think that I disabled them all in checkstyle config

harrel56:
because now it fails on formatter step and checkstyle seems to have no complaints

Julian:
> I think if the "pre-commit.ci" would commit changes from formatter then all would be fine. 

Ah got it ok -- I'm not sure I have high hopes there, but I'm going to ask in the pre-commit channel when the US wakes up if anyone has a java linter that works with `pre-commit.ci`...

Julian:
> But I wonder if it would need to be totally separate workflow from the one with pre-commit?

`pre-commit.ci` doesn't run via GitHub actions at all (i.e. it's not in the workflow) -- right now we actually run pre-commit twice (for no reason really) -- once via pre-commit.ci which is entirely external and simply reads the pre-commit config file and runs pre-commit and then reports on it, and then again from GitHub actions from within the workflow.

The latter we have full control over, and as you can see works (it's what's producing those diffs).

The former we don't, it has a fixed set of available dependencies (and they don't include docker). But using this is much much faster, and has some other benefits -- and all the other linters currently work fine with it. So.. yeah :/ will see what they say but if you have ideas I'm certainly all ears. I have to dig through the pre-commit.ci docs possibly too to remind myself.

harrel56:
Right, so I did it like you said - moved java linting to separate workflow. Had to create special config file for this linter and wasn't sure where to put, so put in in project root. Let me know what you think about this setup - generally we could stick to such division between formatters (pre-commit with autofixes) & linters (separate GH action).

PS: there is a load of rules for this PMD, tried to achieve some balance here, but I'm pretty sure that in the future the rules will need to get adjusted (lots of rules are totally unnecessary, at least IMO)

harrel56:
And maybe if we decide to go with such setup we could move eslint from pre-commit too

Julian:
> generally we could stick to such division between formatters (pre-commit with autofixes) & linters (separate GH action).

I'm not sure I understand the benefit of doing this in general -- to me this is more about "we can't put this one in pre-commit", I'd rather as many things be in the former as possible, because it obviates someone who doesn't necessarily know anything about the language they're dealing with from needing to figure out how to install things in it, everything happens automatically.

As is here, someone needs to learn about Java to figure out how to run `checkstyle` -- so if all of a sudden we find a pre-commit hook that *does* work in `pre-commit.ci` I'd very much prefer we go back and drop the extra action and move it back into pre-commit.

But I'm certainly OK with leaving this in the short term, having it is way better than not having it.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
+name: Lint code
```
Julian, line 1: I suspect it would have been easier to leave this more or less as it was via `pre-commit` and just had a second `.non-pre-commit-ci.yaml` file and used `pre-commit run --config` -- in fact I suspect that would have been a two line change to the existing setup (just adding a second `uses: pre-commit/action` via [custom invocation](https://github.com/pre-commit/action#using-this-action-with-custom-invocations) to include the extra config file.

The advantage again would have been that someone can still run this without needing to read this workflow and figure out how to reproduce it to run the linter.

But it's not a huge deal I suppose, so if you've done the work already here and it works (which it seems to) I'm ok with merging -- though if you do merge, please rename the file and workflow title to something quite a bit more "Java linters"-specific, and yeah can we put `pmd-ruleset.xml` somewhere way less conspicuous, e.g. `.java-implementations-pmd-ruleset.xml` or `implementations/java-implementations-pmd-ruleset.xml` or something?

harrel56: I mean, the changes would have been way narrower, but I decided to give PMD a try instead of checkstyle

harrel56:
So I thought about having two pre-commit config files and honestly if you're fine with it then it's probably the simplest way to do this. Gonna refactor it later

harrel56:
Riight, it's way much more hassle than I anticipated - generally had problems that PMD hook always said that are no files to process :) it's not ready yet, but I'm gonna be out for a week or so, so gonna get back to it then or if you feel like it @Julian you could take it from there (generally I feel like the most problems are with pre-commit config, maybe I missed some little details there)

Julian:
Hah no problem thanks for all the effort, will have a look, probably Monday.

Julian:
(I also had a few other things come up this week so haven't really looked here -- hoping you're enjoying some time off, obviously no urgency on this so either of us will end up bothered enough to finish it off whenever even if it means going back to something like what you had before :/).

harrel56:
I guess it works now - turned out that pre-commit action adds `--all-files` argument by default and it was getting lost when providing additional `extra_args` (https://github.com/pre-commit/action/blob/main/action.yml). Also noticed small issue with this action - the cache key is calculated from default config file (`pre-commit-config.yaml`) and it doesn't take into account the config passed in extra args. I don't know how this is affecting things but if we want to get it right then probably the easiest way would to be to hardcode fixed action in this repo?

Another question is if it's better to have this special pre-commit config for GH CI to contain ALL stuff from "normal" config + problematic hooks configurations or it should just contain these problematic hooks configurations?

Julian:
Wow well done.

> I don't know how this is affecting things but if we want to get it right then probably the easiest way would to be to hardcode fixed action in this repo?

I *think* the effect should just be that it never caches the built docker image, which would be annoying I guess, but given:

> Another question is if it's better to have this special pre-commit config for GH CI to contain ALL stuff from "normal" config + problematic hooks configurations or it should just contain these problematic hooks configurations?

I'd prefer the latter (just problematic), so just the Java linter, and given the above that means just running that one will be slower, but I think relative to the rest of CI no one will notice probably? Especially if eventually we get it combined somehow (seems unlikely given `pre-commit.ci` doesn't support pre-built Java hooks, but maybe some day.)

implementations/java-implementations-pmd-ruleset.xml
```diff
@@ -0,0 +1,315 @@
+<?xml version="1.0"?>
```
Julian, line 1: Sorry, one last nitpick -- I sometimes use `ls implementations/ | wc -l` as a quick way to count how many we support.

Can you hide this file (i.e. rename to `.java-implementations-pmd-ruleset.xml`) so it doesn't cause minor inconvenience (of needing to change to filter directories) there?

harrel56:
> Sorry this was an ordeal, thanks for figuring it out!

I mean it was not so bad, just needed to have a fresh look at it instead of committing furiously ;)

Issue #264, opened 2023-05-22
Add documentation for using Bowtie as a GitHub Action


Julian:
Bowtie is now enabled as a [GitHub action](https://github.com/marketplace/actions/setup-bowtie).

This allows someone to use Bowtie to validate instances against one or more implementations Bowtie supports from within their own workflows.

We should add a page to [Bowtie's documentation](https://bowtie-json-schema.readthedocs.io/en/latest/) showing how to do so in a simple example, perhaps for someone who wishes to validate a specific file in their repository against a schema which also lives within it.

This may involve creating a second repository as an example repository (e.g. `bowtie-json-schema/actions-example`) though if we can avoid doing so somehow for the example it'd be a bit cleaner for the organization.

VishrutAggarwal:
Hi, I'd like to work on this issue

VishrutAggarwal:
Hi @Julian,

I apologize for the absence.
I was thinking of creating a repository on my account and trying out the action while documenting the steps first before making it cleaner.

Will that be fine or do I need to do it directly in a way that keeps it clean (This would require me to first think of the best way to do it first)

Again, I'd like to apologize for the infrequent updates, I have prior college commitments which take up a lot of my time.

Julian:
No problem! Appreciate the update! Creating a repository on your account sounds good to me, we can always clean it up afterwards.

VishrutAggarwal:
Update!

I have created a repo on my account which you can find [here](https://github.com/VishrutAggarwal/bowtie-actions-example).

I have also added a schema, an instance and a workflow YAML file to install bowtie.

Julian:
Awesome! That looks great! I think we can probably skip using `smoke` and just focus on validating using the schema and instance you have -- i.e. using `bowtie validate`

Julian:
Whoops, hit the wrong button, but yeah ^^ that's probably the right next step (maybe even the last one, looks pretty good!)

VishrutAggarwal:
Hey, I'm trying to put in the `bowtie validate` command but I keep on receiving this error:

> Error: Invalid value for 'SCHEMA': 'bowtie-actions-example/schema.json': No such file or directory

I have tried different paths to the file and have been unsuccessful so far.

I've tried:
- `schema.json`
- `./schema.json`
- `bowtie-actions-example/schema.json`

Is there a way to reference the repository itself? or am I approaching this the wrong way?

Julian:
[...]

PR #350, opened 2023-07-19
[Docs] Added GitHub Actions Documentation

VishrutAggarwal:
This PR adds a documentation on how to use Bowtie as a GitHub Action as specified in #264


Julian:
@VishrutAggarwal if you update the branch to pull in the commit I just pushed to main you should see a preview of how this renders show up in the PR.

docs/gh-action.rst
```diff
@@ -0,0 +1,136 @@
...
+    jobs:
+    sets-up-bowtie:
+        runs-on: ubuntu-latest
+
+        steps:
+        - name: Install Bowtie
+            uses: bowtie-json-schema/bowtie@v2023.05.12
+
+
+`The code snippet will install bowtie every time you push onto the repository.`
```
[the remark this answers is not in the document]
VishrutAggarwal: It still shows the same test failing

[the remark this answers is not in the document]
VishrutAggarwal: Update: It took some time to update, but it's fine now

Julian:
Ah no sorry, it's because I set it to only add them for newly opened PRs, so this one didn't get included because it was open already. The next one should have it automatically, yours is here though: 

https://bowtie-json-schema--350.org.readthedocs.build/en/350/

VishrutAggarwal:
>  this one didn't get included because it was open already

Makes sense.

Though if there are changes that are required before merging the PR, please let me know. I have a couple of hours free and can contribute a major chunk of it to this.

docs/index.rst
```diff
@@ -22,6 +22,7 @@ Contents
     cli
     implementers
     contributing
+    gh-action
```
Julian, line 25: Let's call the file `github-actions.rst` (and then update here).

docs/gh-action.rst
```diff
@@ -0,0 +1,136 @@
+==============================
+Using Bowtie in GitHub Actions
+==============================
+
+You can now use bowtie as a GitHub Action. Follow the steps to get started with it in a few simple steps:
```
Julian, line 5: Rather than saying "now", which will go out of date, best to just stick to the present tense, i.e.:

```suggestion
Bowtie can be used from within GitHub Actions.
Follow these steps to get started with it in a few simple steps:
```

Also have a quick look at [this style guide on semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) -- in short, generally put each sentence on its own line in the file here (that will still get combined into paragraphs in the rendered docs).

docs/gh-action.rst
```diff
@@ -0,0 +1,136 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+        - name: Install Bowtie
+            uses: bowtie-json-schema/bowtie@v2023.05.12
+
+
+*The code snippet will install bowtie every time you push onto the repository.*
+
+2. Verification Steps
```
Julian, line 36: I think it's OK for people to skip this step in general and not print the version, that's perhaps mostly useful just during debugging, but otherwise there's no reason to always run it, the version will match what you use in the `uses` line, i.e. `bowtie-json-schema/bowtie@v2023.05.12` should generally give you a `--version` of `v2023.05.12``.

So think you can likely skip this section.

docs/gh-action.rst
```diff
@@ -0,0 +1,136 @@
...
+* Open the drop down of the ``Run Bowtie`` step.
+* If it shows something like this, bowtie is running successfully:
+
+.. code:: sh
+
+    bowtie, version 2023.6.4
+
+.. admonition:: Note
+
+    IF THE RUN WAS NOT SUCCESSFUL, YOU WILL SEE A RED CIRCLE WITH A CROSS. RECHECK THE CODE YOU HAVE WRITTEN, AND CORRECT IN CASE OF ANY DISCREPANCY.
```
Julian, line 80: Let's assume someone reading this page already knows the basics of GitHub Actions -- definitely useful to have resources on it if not, so perhaps let's add a link to the GitHub documentation at the top, but we can skip describing something like this, which the person should know if they've used GitHub actions in some other context.

docs/gh-action.rst
```diff
@@ -0,0 +1,136 @@
...
+* ``bowtie validate``: It tells which command to run.
+* ``-i``: It is a required flag for ``validate`` command and specifies which implementation to be used, which in this case is ``lua-jsonschema``.
+* ``schema.json``: It is the name of the file containing the defined schema against which instances will be validated.
+* ``instance.json``: It is the name of the file containing the instances to be validated. This can also be the path to a directory that contains all the instances.
+
+
+*You will see that the implementation is skipped and thus does not validate the instances. This is because the lua implementation does not support the default 2020-12 draft.*
+
+
+2. Change the Dialect
```
Julian, line 111: This sort of information I think belongs in the CLI documentation, it's not specific to using Bowtie in GitHub actions, so perhaps have a look there and see if it already contains this info -- if it doesn't, we should add it there, but if it does, you can probably just link to that page instead of repeating it here.

Julian (commented):
Thanks! Left a few comments -- I think basically we can/should shorten this a bit, either by linking to existing documentation (like for GitHub Actions basics) or to Bowtie documentation (for in depth information about how to use the CLI).

That'll let this page be short and focus just on the 2 or 3 steps in running Bowtie from GitHub actions without needing too much extra detail.

docs/gh-action.rst
VishrutAggarwal: I wrote it here specifically because I did not find it in the CLI docs. I will add it to that page instead of here, I agree with it being more suited there.

Julian:
(Otherwise if/when someone else gets to this I'm sure we'll credit your work!)

VishrutAggarwal:
Hi, the checks are failing due to the link not being valid yet.

I suppose it'll pass the checks once the transfer of repository is complete.

Julian:
Great! There's a tiny tweak or two to make, but I think I'll perhaps merge this and simply add it after.

Thanks for your work!

PR #357, opened 2023-07-24
Update tutorial of bowtie documentation

tcskiran:
The image building is getting an error when done in docker.
``` bash
docker build --quiet -f Dockerfile -t bowtie-lua-jsonschema
```
Updated the documentation so that it works now in both podman, docker.


Julian:
Thanks, appreciated!

There's a second `podman build` command lower on the page, can you make the same change there?

PR #370, opened 2023-08-03
Add runner for kmp-json-schema-validator

OptimumCode:
Hi, I would like to add an integration with Bowtie for a new Kotlin multiplatform library [json-schema-validator](https://github.com/OptimumCode/json-schema-validator)

----
:books: Documentation preview :books:: https://bowtie-json-schema--370.org.readthedocs.build/en/370/


implementations/kmp-json-schema-validator/.gitattributes
```diff
@@ -0,0 +1,8 @@
+#
+# https://help.github.com/articles/dealing-with-line-endings/
+#
+# Linux start script should use lf
+/gradlew        text eol=lf
```
harrel56, line 5: Do we even need this file as there are no wrapper scripts in repo?

implementations/kmp-json-schema-validator/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM gradle:8.2.1-jdk11
+WORKDIR /opt/app
+COPY gradle.properties .
+COPY settings.gradle.kts .
+COPY build.gradle.kts .
+RUN gradle dependencies > /dev/null
```
harrel56, line 6: What this line do? Does it have some effect?

implementations/kmp-json-schema-validator/gradle.properties
```diff
@@ -0,0 +1,2 @@
+org.gradle.parallel=true
+org.gradle.caching=true
```
harrel56, line 2: Is caching needed? There is only one call to install/build happening, right?

implementations/kmp-json-schema-validator/gradle.properties
```diff
@@ -0,0 +1,2 @@
+org.gradle.parallel=true
```
harrel56, line 1: If it would mean that this file can be removed, I would just use `--parallel` when calling gradle. But it's just my aesthetic preference :)

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
```diff
@@ -0,0 +1,214 @@
...
+                )
+            )
+        )
+    }
+
+    private fun requireDialect() {
+        require(currentDialect.isNotEmpty()) { "dialect is not set" }
+    }
+
+    private fun handleDialect(command: Command.Dialect): Result {
```
harrel56, line 169: I think `dialect` message is the most misleading in the whole harness <-> bowtie communication. This is probably fine, but you can look at this comment for clarification: https://github.com/bowtie-json-schema/bowtie/pull/316#discussion_r1251030816

harrel56 (commented):
Left a few comments. I'm not really familiar with kotlin but I guess it looks good to me :)

harrel56:
Oh, and I think you also should add dependabot config for your implementation (`.github/dependabot.yml`). Same for pre-commit config for formatting and ideally linting. As this is the first kotlin implementation it may not be as straightforward. But at least it seems that PMD also supports kotlin?

And as there is a convention for naming implementation directory it should start with language (so probably it should be `kotlin-json-schema-validator`).

implementations/kmp-json-schema-validator/.gitattributes
OptimumCode: Initially there was a wrapper but I have decided that it shouldn't be committed (other impls that using Gradle doesn't have it). But I forgot about this file

implementations/kmp-json-schema-validator/Dockerfile
OptimumCode: Yes, it is used to initialize some part of the Gradle caches and reuse it in future builds.

Docker caches each step. So, if nothing is changed in gradle.properties, settings.gradle.kts or build.gradle.kts the docker starts right from the cached image that was produced by this step. In this case the command `RUN gradle install` will be executed faster (at least it is my understanding of that and it seems to be working on my machine)

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
OptimumCode: I read the comment. If I get it correctly, then it means it is better to finish the program with an error if it receives an unsupported dialect. Correct me, please, if I am wrong

implementations/kmp-json-schema-validator/gradle.properties
OptimumCode: Yes, it will be equivalent. I will remove gradle.properties file)

OptimumCode:
Thank you @harrel56 for comments.

I will try to update required configurations.

Regarding the implementation name: if it is okay I would like to keep `kmp` prefix because the implementation is written for [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform.html). I want to leave `json-schema-validator` name for other implementations that are not multiplatform (e.g. for JVM only) 🙂
So, I hope the name `kotlin-kmp-json-schema-validator` will be okay)

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private fun shouldSkipTest(caseDescription: String, testDescription: String): Boolean {
+        // TODO: check tests to skip
+        return false
+    }
+
+    private fun handleRun(command: Command.Run) {
+        requireStarted()
+        requireDialect()
```
Julian, line 104: I think @harrel56 already pointed out some of the slightly trickiness here, but this line / this check too is not necessarily needed/true -- if your harness is sent the schema `{"$schema": "http://json-schema.org/draft-07/schema#", "type": "integer"}` no dialect request is needed/required, the appropriate dialect is already indicated in the schema.

The dialect request is essentially needed only for schemas that don't have a `$schema` property.

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
```diff
@@ -0,0 +1,214 @@
...
+                    ),
+                ),
+            )
+            return
+        }
+        val results: List<TestResult> = command.case.tests.map { test ->
+            runCatching {
+                if (shouldSkipTest(command.case.description, test.description)) {
+                    TestResult.Skipped(
+                        message = "test ${test.description} in the list of ignored tests",
```
Julian, line 138: This message is really meant for someone trying to understand why the test is skipped, so it'd be a bit better to either link to an issue in your repository which explains when the functionality might be expected (or otherwise to just say "ref resolution is still being developed" or something).

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
Julian: That's correct, this message is to ask your implementation to treat schemas without `$schema` a certain way.

Julian (changes requested):
Very nice, thanks for the PR!

Left a few minor additional comments, also agree with Harrel's on reducing the unneeded files and adding `kotlin` to the directory name.

Appreciate the PR and good luck with your implementation!

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
OptimumCode: There are no tests skipped at the moment but I will use the same mechanism here as I used for skipping cases

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
OptimumCode: Okay, I get it, I think. Will change the implementation

implementations/kmp-json-schema-validator/src/main/kotlin/BowtieSampsonSchemaValidatorLauncher.kt
Julian: Right!

OptimumCode:
> But at least it seems that PMD also supports kotlin?

@harrel56 sorry, I missed the part about PMD. I could find only a few rules [there](https://pmd.github.io/pmd/pmd_rules_kotlin_bestpractices.html). Probably, I can add a hook for [detekt](https://detekt.dev/) execution

harrel56:
> > But at least it seems that PMD also supports kotlin?
> 
> @harrel56 sorry, I missed the part about PMD. I could find only a few rules [there](https://pmd.github.io/pmd/pmd_rules_kotlin_bestpractices.html). Probably, I can add a hook for [detekt](https://detekt.dev/) execution
> 

Yep, that would be great. Just use whatever you consider best for kotlin

OptimumCode:
Hi, @harrel56 and @Julian.
I would like to ask for your opinion: what do you think about creating a local pre-commit hook?

I have found only this existing pre-commit hook for executing `detekt` checks - https://github.com/dustinsand/pre-commit-jvm.
However, it looks like the last update was 5 month ago and it uses the old version of `detekt`. And it also requires [coursier](https://get-coursier.io/) to be installed. So, it does not look like a good option.

If you don't mind I would add a local hook that will execute script like this (https://detekt.dev/docs/gettingstarted/git-pre-commit-hook#only-run-on-staged-files). It also will download the required version of `detekt` and execute it.

Julian:
I'd prefer one in a separate repo honestly, either yours if you care to create it or otherwise one we create in parallel to this one -- a local one is unlikely to get anyone to contribute to it / take over maintaining it, and is essentially an equal amount of work to put together. At least in a separate repo others could find and use it.

OptimumCode:
@Julian @harrel56 One more thing (missed that): ktlint requires Java to run it. Is it okay to add it to the GitHub workflow?

Julian:
You'd add Java to the dockerfile that runs the pre-commit hook. If you look at the `pre-commit-java` hook I suspect it'd look similar

OptimumCode:
> You'd add Java to the dockerfile that runs the pre-commit hook. If you look at the `pre-commit-java` hook I suspect it'd look similar

So, you suggesting to create my own pre-commit hook here as well (using docker), am I right?

Julian:
Looking more carefully now,

> However, it looks like the last update was 5 month ago and it uses the old version of detekt. And it also requires [coursier](https://get-coursier.io/) to be installed. So, it does not look like a good option.

This repository seems to work for me, what's making you say it's not a good option? I was able to set it up pretty easily on my machine, and perhaps more importantly it has a high likelihood of working in pre-commit.ci as well. If the version is old (which it doesn't seem very, it's 1.22 and the newest seems like 1.23.1) perhaps we simply send a PR to that repository?

That would certainly save the work of creating a whole new repository with the exact same setup to run `detekt` and `ktlint`. But yes the desired end goal is both of those running in `pre-commit`!

OptimumCode:
Thank you, @Julian, for the clarification.

> This repository seems to work for me, what's making you say it's not a good option?

Yes, it works fine, but you must install [coursier](https://get-coursier.io/). And you also need to install it in the CI (I will double-check that moment). If it does not require any changes in the CI I will use them. But if the changes are required I will create a repo with hooks that uses docker for execution

OptimumCode:
So, it looks like if I use those existing pre-commit hooks the CI must be changed (coursier must be installed).
If @harrel56 and @Julian are okay with installing it I will add it to the CI. Otherwise, I will continue with creating my own hooks that use docker to run

<details><summary>Job run</summary>
[CI/pre-commit]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/2-composite-3.sh] user= workdir=</br>
| [INFO] Initializing environment for https://github.com/gherynos/pre-commit-java.</br>
| [INFO] Initializing environment for https://github.com/dustinsand/pre-commit-jvm.</br>
| [INFO] Installing environment for https://github.com/gherynos/pre-commit-java.</br>
| [INFO] Once installed this environment will be reused.</br>
| [INFO] This may take a few minutes...</br>
| [INFO] Installing environment for https://github.com/dustinsand/pre-commit-jvm.</br>
| [INFO] Once installed this environment will be reused.</br>
| [INFO] This may take a few minutes...</br>
| An unexpected error has occurred: AssertionError: pre-commit requires system-installed "cs" or "coursier" executables in the application search path</br>
| Check the log at /root/.cache/pre-commit/pre-commit.log</br>
</details>

Julian:
Can you link to the run that ran on?

OptimumCode:
> Can you link to the run that ran on?

It was a local run using [act](https://github.com/nektos/act) (a very handy tool). But I can reproduce it in GitHub if you want

Julian:
I don't think that's going to reproduce in GitHub actions, because in GitHub actions it is going to run in `pre-commit.ci` (and not in GitHub actions)

OptimumCode:
> in GitHub actions it is going to run in `pre-commit.ci` (and not in GitHub actions)

Didn't know about that. Thank you for the information :-)

But what about this: https://github.com/bowtie-json-schema/bowtie/blob/main/.github/workflows/ci.yml#L17
This one will be running in GitHub actions, or am I wrong here?

Julian:
> But what about this: https://github.com/bowtie-json-schema/bowtie/blob/main/.github/workflows/ci.yml#L17
This one will be running in GitHub actions, or am I wrong here?

That's basically only running the actions which *don't* run in `pre-commit.ci` correctly -- which right now is only the Java linters... though the repo you found actually seems to also contain `pmd` as well! So if this works correctly we may actually be able to use the `pmd` hook in that repo for Java too.

Julian:
CI is failing but that's because you added the new linter to both pre-commit files (sorry, this is annoyingly complicated because of the aforementioned Java linters).

But indeed it looks like `pre-commit.ci` worked fine -- meaning, if you remove your linter from here: https://github.com/bowtie-json-schema/bowtie/pull/370/files#diff-6d46c09d13a3d14107caa8c9822f12530775483c6f06b5113ec8c5ad61d08fe1R14 -- i.e. from the `non-pre-commit-ci` config, everything should work.

Julian:
> Do you want me to move pmd t[here](https://github.com/dustinsand/pre-commit-jvm/blob/main/.pre-commit-channel/pmd.json) as well (using the new hook from here)?

I think it's probably wise to merge this as soon as everything passes, and then we can change the Java one in a second PR

Julian:
Great!

> Looks like Bowtie does not want to use a local image and tries to download one from the GitHub registry (but it is not published there)

Indeed, this is a "known issue" in that there's something I don't understand about how GitHub runners seem to throw away container state in between running steps in a workflow.

I'm going to merge this (well done!) and do a release so we can see how the results look, if something needs fixing you can stick in a follow-up PR.

Julian:
You're showing up on https://bowtie.report/#/draft7 !

OptimumCode:
Great! Thank you @Julian and @harrel56 for the review and suggestions

PR #373, opened 2023-08-05
fix: Support for multiple BASE_URL variables

harrel56:
Things got a little complicated with custom domain name. It now requires base url to be equal to `/`, but for fork deployments it still should be `/bowtie` (e.g. link `https://harrel56.github.io/assets/index-e150a41d.js` should be `https://harrel56.github.io/bowtie/assets/index-e150a41d.js`). This PR introduces a way to control base url using GH variable (`BOWTIE_BASE_URL`). I did it so if the variable is absent it uses `/bowtie` value. This way all future fork deployments should work without any additional configuration, but the original bowtie repo should set up a variable (`BOWTIE_BASE_URL` with value `/`) in order to work correctly.

@Julian Let me know what you think about this approach, couldn't come up with anything simpler and ofc I'm open to other ideas.

----
:books: Documentation preview :books:: https://bowtie-json-schema--373.org.readthedocs.build/en/373/

### DIFF
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -18,7 +18,7 @@ jobs:
           node-version: 18.x
       - run: npm ci
         working-directory: ./frontend
-      - run: npm run build
+      - run: npm run build -- --base=${{ vars.BOWTIE_BASE_URL || '/bowtie' }}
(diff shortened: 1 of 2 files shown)


Julian:
Nice! This is good with me!

Issue #248, opened 2023-04-25
Redesign the site path structure


Julian:
*This is a WIP / raw thoughts on redesigning the path structure of the UI.*

- [ ] `/dialects/<dialect-name>` - a cross-implementation page showing compliance summaries across all implementations supporting the dialect
- [ ] `/dialects/<dialect-name>?language=<language>` - a cross-implementation page filtered to a given programming language
- [ ] `/dialects/<dialect-name>?vocabulary=<vocabulary>` - a cross-implementation page showing compliance summaries across all implementations for a specific JSON Schema Vocabulary
- [ ] `/dialects/<dialect-name>?keyword=<keyword>` - a cross-implementation page showing compliance summaries for a specific JSON Schema keyword
- [ ] `/implementations/<lang>-<implementation-name>` -- a per-implementation page showing compliance summaries across all supported dialects for the implementation
- [ ] `implementations/<lang>-<implementation-name>/dialects/badge` - a per-implementation badge showing the known supported dialects for the implementation
- [ ] `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>` - a per-implementation page showing compliance summaries across a specific dialect
- [ ] `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>/badges/compliance` - a per-implementation badge showing the pass percentage for all tests within the given dialect
- [ ] `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>/badges/failing-keywords` - a per-implementation badge showing the names of all JSON Schema keywords within the dialect with failing tests

PR #379, opened 2023-08-07
Feature/dialects path

harrel56:
Part of path structure redesign (https://github.com/bowtie-json-schema/bowtie/issues/248 - `/dialects/<dialect-name>`).

----
:books: Documentation preview :books:: https://bowtie-json-schema--379.org.readthedocs.build/en/379/

### DIFF
--- a/frontend/src/components/NavBar.jsx
+++ b/frontend/src/components/NavBar.jsx
@@ -67,12 +67,18 @@ const NavBar = () => {
                 </a>
                 <ul className="dropdown-menu">
                   <li>
-                    <NavLink className="dropdown-item" to="/dialects/draft2020-12">
+                    <NavLink
+                      className="dropdown-item"
+                      to="/dialects/draft2020-12"
+                    >
                       2020-12
                     </NavLink>
                   </li>
                   <li>
-                    <NavLink className="dropdown-item" to="/dialects/draft2019-09">
+                    <NavLink
+                      className="dropdown-item"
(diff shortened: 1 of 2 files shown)


frontend/src/components/NavBar.jsx
```diff
@@ -25,7 +25,7 @@ const NavBar = () => {
         }`}
       >
         <div className="container-fluid">
-          <Link className="navbar-brand mb-0 h1" to="/draft2020-12">
+          <Link className="navbar-brand mb-0 h1" to="/">
```
Julian, line 28: (I haven't tried this, I'm just reading it) but this is sending the homepage to `/` right, but where is it that makes that render  or redirect to `/dialects/draft2020-12`, and/or what's the homepage going to show after this?

harrel56: Routes definitions are located in `index.jsx` file, see call of `createHashRouter()` function. Basically, it should render the same thing it used to render before, which is the report page for `draft2020-12` dialect. There are no redirects, but ofc we could do this the other way around - navigating to `bowtie.report` could always redirect to `bowtie.report#/dialects/draft2020-12`. Either approaches are fine I guess

Julian: Ah I see this change is irrelevant to all that, this is just tweaking some unrelated thing about whether we link to the homepage or to draft2020-12 from the navbar, I see (and yeah this way seems more correct to me).

PR #381, opened 2023-08-08
Feature/filter by language

harrel56:
Should resolve `/dialects/<dialect-name>?language=<language>` part from https://github.com/bowtie-json-schema/bowtie/issues/248.

I've decided to add support for filtering by multiple languages. I guess there can be some considerations from UX/UI perspective - and I'm open to suggestions.

There is one issue that has been present from the beginning, but now it's more visible - navigating (like changing dialects or filters) takes too much time. Issue seems to be `react-router` specific because it keeps unmounting and mounting all the components on each change. I think I'll try to fix this in the next PRs

----
:books: Documentation preview :books:: https://bowtie-json-schema--381.org.readthedocs.build/en/381/


frontend/src/components/FilterSection.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (languages.includes(current)) {
+    const newParams = new URLSearchParams(searchParams);
+    newParams.delete("language");
+    languages
+      .filter((lang) => lang !== current)
+      .forEach((lang) => newParams.append("language", lang));
+    return (
+      <Link key={current} to={{ search: newParams.toString() }}>
+        <Badge pill bg="filter-active">
+          <div className="px-2">{current}</div>
```
Julian, line 42: Slightly nice to titlecase the names here so we see `Python` rather than `python` -- right now in particular for some reason `Kotlin` is the only language title-ized.

Even nicer would probably be mapping these names to arbitrary human readable names -- in particular probably it's nicer to see `.NET` rather tha `Dotnet`.

But OK with at least the former (title casing).

Julian:
> I've decided to add support for filtering by multiple languages. I guess there can be some considerations from UX/UI perspective - and I'm open to suggestions.

The *nicest* UX I think would be if we had a generic table filter widget -- i.e. a fancy `<table>` where you could filter by any of the columns and their values.

But honestly I'm pretty OK with merging something like this as-is and coming back and improving the UX on filtering after, since I'm also pretty sure such a table (like what I mentioned) exists and we can drop it in when we find it.

Back In The Day™ it was some weird jQuery table thing -- I'm sure these days there's some replacement for that.

Julian:
As above I think this is a bit clunky but not bad enough to not merge UXwise (whee double negatives -- basically I think this is good enough for a first version) -- thanks! Feel free to merge after the title casing thing.

harrel56:
Hmm, do you mean something like this https://examples.bootstrap-table.com/#extensions/filter-control.html? What about filters that are not really related to columns, like keywords, vocabularies?

frontend/src/components/FilterSection.tsx
Julian: > Languages returned from harnesses should be somehow standardized so there is no such case that one implementation is in python and other one in Python.

This is regardless indeed a good point, so fixing that (to validate that the returned value is lowercase, and then fixing the Kotlin harness to return it that way)

Julian:
> Hmm, do you mean something like this https://examples.bootstrap-table.com/#extensions/filter-control.html? 

Yeah though for language I think a dropdown (in the header) is nicer than a text box.

> What about filters that are not really related to columns, like keywords, vocabularies?

It's a good question, I'm not 100% sure, though I'm also not 100% sure we shouldn't at that point / at some point show those as columns! (E.g. maybe we have a hidden/unselected column which contains the keywords that a row is interested in, and then filter on that hidden column...)

But yeah I don't disagree that it may not be as simple...

frontend/src/components/FilterSection.tsx
Julian: (Done)

harrel56:
Not sure what exactly do you mean by adding keywords columns to the table, but I guess it's a topic for another discussion

harrel56:
Added language mapping onto the next PR (https://github.com/bowtie-json-schema/bowtie/pull/382) to avoid conflicts. Probably it's best if these 2 PRs are merged together

PR #383, opened 2023-08-09
Change PMD hook and move it to pre-commit config

OptimumCode:
Looks like the previous [hook](https://github.com/gherynos/pre-commit-java/blob/v0.2.9/install.sh#L5) and [the new one](https://github.com/dustinsand/pre-commit-jvm/blob/main/.pre-commit-channel/pmd.json#L7) use same version of PDM.

So, I hope everything should be fine


Julian:
Awesome, looks like it indeed works! Appreciated.

harrel56:
I made a wrong assumption that Coursier won't be accessible in pre-commit, as docker wasn't. I'm glad it got simplified - nice!

PR #384, opened 2023-08-09
Use pnpm in CI and remove the npm lockfile.

Julian:
pnpm is faster and uses less disk space.

I've been using it locally (for this and other projects) with no issue, though we previously had two lockfiles -- `pnpm` essentially has the same interface as npm, but does not use `package-lock.json` even though it uses `package.json` (mostly because the lock file for npm doesn't have enough information in it IIRC). This PR removes the npm lockfile (meaning you should use pnpm locally now, which already has an equivalent lockfile in the repo).

This PR also enables dependency caching, which I don't think was enabled before, so installs should be a bit quicker.

More details: https://pnpm.io/motivation and https://pnpm.io/

----
:books: Documentation preview :books:: https://bowtie-json-schema--384.org.readthedocs.build/en/384/


.github/workflows/build-frontend.yml
```diff
@@ -13,13 +13,35 @@ jobs:
...
+          path: ${{ env.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+
+      - name: Install dependencies
+        run: pnpm install --dir frontend --frozen-lockfile --strict-peer-dependencies
+
+      - name: Build the UI
+        run: pnpm run --dir frontend build -- --base=${{ vars.BOWTIE_BASE_URL || '/bowtie' }}
```
harrel56, line 43: I can see that passing --base argument looks now a little bit different (in action logs), but currently I cannot check if the build is correct - so hopefully it works 🙂

harrel56 (approved):

.github/workflows/build-frontend.yml
Julian: Indeed :D -- test in prod we go...

Issue #24, opened 2022-09-13
Make the default set of bowtie run implementations smarter


Julian:
Right now one has to pass `-i` explicitly for each image.

We should support a number of easier options:

- [ ] Run all known implementations in the registry
- [ ] Run all local images from `implementations/`
- [ ] Run all implementations supporting the current dialect
- [ ] Run all implementations written in language X
- [ ] Run all implementations from `--implementations-file` which contains a list of images

`bowtie smoke` and other commands may also want to default to a wide set of implementations once we can do this (i.e. "smoke test all implementations", if none are otherwise provided)

adwait-godbole:
Indeed #932 has made it easier now. I was planning to tackle this issue after completing `bowtie filter-implementations` support

adwait-godbole:
> Make the default set of bowtie run implementations smarter

You mean `bowtie run` right ?

> Run all known implementations in the registry

Can you explain this one ?

adwait-godbole:
> Make the default set of bowtie run implementations smarter

Hey @Julian, here you mean `bowtie run` command right ?

> Run all known implementations in the registry

Can you explain this one please.

>  Run all local images from implementations/

When you say `from implementations/` that means having a local checkout present but same was the problem with `filter-implementations` that its not necessary the user has a local checkout so now that we have #932 we can do the similar thing like we did there which is `Implementation.known() if sys.stdin.isatty() else sys.stdin`.

> Run all implementations supporting the current dialect
> Run all implementations written in language X

One possible solution for these 2 could have been to first run `bowtie filter-implementations` from #920 and then pass the filtered output to `bowtie run` but the flaw here is that we will essentially have to spin up the containers twice, once for filtration and then again for running the test (if it was filtered). Quite expensive and slow :(. Having #874 would've certainly helped here, but if you could perhaps maybe provide some more context on how to solve that one then this gets a lot easier!

> Run all implementations from --implementations-file which contains a list of images

Is is suppose to accept a `.txt` or a `.json` ? or both ?

Julian:
> Hey @Julian, here you mean bowtie run command right ?

I meant "implementations which are run by Bowtie" I guess, rather than `bowtie run` specifically. In today's implementation, that corresponds essentially to the `implementation_subcommand` helper.

There's two parts to this issue really, one's straightforward, the other is slightly less so but still not impossible:

* Make the `-i` argument in `implementation_subcommand` support these more intelligent options
[...]

PR #397, opened 2023-08-14
feat: add per-implementation page

AgniveshChaubey:
This PR adds `/implementations/<lang>-<implementation-name>` path to the **_Bowtie Report_** (as mentioned in #248). 

At this stage, the newly added page displays the runtime info of all the implementations supported by Bowtie which are compatible with the `draft2020-12`. 

I tried to write the code in TypeScript, but as I'm relatively new to TypeScript, there's room for potential improvements. Any guidance in this area would be greatly appreciated. Also, any suggestions pertaining to the implementation-specific pages are also welcomed.

I have also done a basic setup for `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>`, however, I won't be focusing on this route as part of this PR (I will address this one in the next follow-up PR).

-----------------------
Edit: Following the suggestions below, I am removing `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>` to narrow down the scope of this PR to address one issue at a time.

----
:books: Documentation preview :books:: https://bowtie-json-schema--397.org.readthedocs.build/en/397/


frontend/src/components/Summary/ImplementationRow.jsx
```diff
@@ -4,21 +4,40 @@ import { useState } from "react";
...
 import { mapLanguage } from "../../data/mapLanguage";
+import { NavLink } from "react-router-dom";
 
 const ImplementationRow = ({ cases, implementation }) => {
   const [showDetails, setShowDetails] = useState(false);
   const [showRunTimeInfo, setShowRunTimeInfo] = useState(false);
 
+  const getPath = (language, name) => {
+    const urlPath = {
+      "c++-valijson": "cpp-valijson",
```
Julian, line 15: Cool! I'll take a closer look at the entire PR, but because I see it immediately -- I wouldn't want us hardcoding anything here (or really in the UI in general).

I believe which implementations support draft2020-12 is already available in the data outputted by the 2020 report (but if that or anything else ever wasn't available, we should make it available rather than hard coding it, we control both sides here!)

AgniveshChaubey: Yes, you are absolutely correct. The data is, indeed, available within the report. However, to grab the data related to a specific implementation, we need to verify whether the **name** and **language** of that implementation are present in any of the available image URLs.

In most cases, the Image URL follows the pattern `/<language>-<name>` at the end, for example `go-jsonschema`. However, there are cases where obtaining data for certain implementations becomes challenging due to their names. For example, matching `JsonSchema.Net (dotnet)` with `/dotnet-jsonschema-net` is difficult. Due to this reason, I've hard-coded these specific values.

If we want to make the entire path dynamic, we would need to make slight adjustments to the implementation names. For example, in the metadata, we would have to change the language name from `c++` to `cpp`, implementation name form `JsonSchema.Net` to `jsonschema-net` and so on.

AgniveshChaubey: Yes, you are absolutely correct. The data is, indeed, available within the report. However, to grab the data related to a specific implementation, we need to verify whether the **name** and **language** of that implementation are present in any of the available image URLs.

In most cases, the Image URL follows the pattern `/<language>-<name>` at the end, for example `go-jsonschema`. However, there are cases where obtaining data for certain implementations becomes challenging due to their names. For example, matching `JsonSchema.Net (dotnet)` with `/dotnet-jsonschema-net` `(ghcr.io/bowtie-json-schema/dotnet-jsonschema-net)` is difficult. Due to this reason, I've hard-coded these specific values.

If we want to make the entire path dynamic, we would need to make slight adjustments to the implementation names. For example, in the metadata, we would have to change the language name from `c++` to `cpp`, implementation name form `JsonSchema.Net` to `jsonschema-net` and so on.

Julian: Looking at the report, from what I see, this is all already there. This is the first line for a run I just did with 1 implementation:

```json
{"started": "2023-08-14T17:54:15.968585+00:00", "bowtie_version": "2023.8.2.dev53+g5dc9de3.d20230804", "dialect": "https://json-schema.org/draft/2020-12/schema", "implementations": {"ghcr.io/bowtie-json-schema/python-jsonschema": {"language": "python", "name": "jsonschema", "version": "4.18.4", "homepage": "https://python-jsonschema.readthedocs.io/", "issues": "https://github.com/python-jsonschema/jsonschema/issues", "dialects": ["https://json-schema.org/draft/2020-12/schema", "https://json-schema.org/draft/2019-09/schema", "http://json-schema.org/draft-07/schema#", "http://json-schema.org/draft-06/schema#", "http://json-schema.org/draft-04/schema#", "http://json-schema.org/draft-03/schema#"], "os": "Linux", "os_version": "6.4.7-200.fc38.aarch64", "language_version": "3.11.4", "image": "ghcr.io/bowtie-json-schema/python-jsonschema"}}, "metadata": {"Commit": {"text": "5cc9214e82f1e0a5e9644960b6fe0166afb7b283", "href": "https://github.com/json-schema-org/JSON-Schema-Test-Suite/commit/5cc9214e82f1e0a5e9644960b6fe0166afb7b283"}}}
```

which already contains the mapping between ID (image name) and language and name.

Have a look and point out if you see something specific missing.

Julian: (`http get https://bowtie.report/draft2020-12.json | head -n1 | jq` is the "production" file obviously, and seems to already contain all this data for all implementations we run -- so you want to parse that and create your routes probably)

AgniveshChaubey: To visit an implementation page, one has to click the link on the homepage. Now according to the current code, the link will take them to `base/implementations/${language}-${name}`. 
For example, when someone clicks 
image
then they will be redirected to `base/implementations/go-jsonschema` and when someone clicks 
image
they will be redirected to `base/implementations/dotnet-JsonSchema.Net`.

In the case of `base/implementations/go-jsonschema`, the substring 'go-jsonschema' is present in one of our unique IDs (i.e. - `ghcr.io/bowtie-json-schema/go-jsonschema`), so we can use the value mapped to this image URL and display it on the following page `base/implementations/go-jsonschema`. 

But in the case of `base/implementations/dotnet-JsonSchema.Net`, it is difficult to get the value mapped to `ghcr.io/bowtie-json-schema/dotnet-jsonschema-net` ID because the image URL does not include the substring "dotnet-JsonSchema.Net" (it includes "dotnet-jsonschema-net").

However, if we could somehow visit `base/implementations/dotnet-jsonschema-net` instead of  `base/implementations/dotnet-JsonSchema.Net`, then we could use the **_dotnet-jsonschema-net_** substring to grab this `ghcr.io/bowtie-json-schema/dotnet-jsonschema-net` ID and use the associated value to display on the dotnet specific page.

Only a few of the implementations have the above-mentioned issue while the rest of them are working totally fine with `base/implementations/${language}-${name}`

Do you have any other ideas for achieving the above functionality?

AgniveshChaubey: Hmm, I overlooked that the data is already available in `ImplementationRow.jsx` with a unique ID.
Fixed!

harrel56:
I'm not sure about the goal of this PR/task - what should be presented on this specific implementation page?

Right now e.g. `/implementations/go-jsonschema` shows basically the same info as modal dialog with implementation info. The only additional thing is list of supported dialects. The way I would see this: 

- it should contain a table with each **dialect** as a row, and for each dialect it should show number of passed/errored/skipped cases for this specific implementation. Now, this would require fetching all the `jsonl` reports (so 6 requests to the server), but I believe if do them all at once (concurrently) this should not be a big deal. If it's still clunky, it could be optimized later on :)
- I would expect that if we are on `/dialects/draft2020-12` and then we click on a specific implementation it would navigate to `/dialects/draft2020-12/implementations/go-jsonschema` - which is reversed version of what of what was originally proposed. TBH I haven't thought about contents of this page yet.

Then the question is: how could user navigate to `/implementations/go-jsonschema`? Ideally, I would see it on the main navigation bar, right next to the `Dialects` menu. It would work basically the same as `Dialects` menu, but it would contain all the implementations. But, to render this navigation menu we would need to know all the possible implementations, which would require to fetch all `jsonl` reports. So, I would say that doing it right now is not feasible, but I suppose we could generate simple list of all implementations together with reports?

harrel56:
Moreover, can you limit the scope of this PR to include only changes related to one new path? I mean, each subtask of https://github.com/bowtie-json-schema/bowtie/issues/248 is a lot of work, so why not separate them?

PS please switch to `pnpm` and not commit `npm` lock file.
PS2 IMHO `/doc` path is not really something needed or even something that makes sense. Why don't use navigation link to docs page directly?

Julian:
I agree with basically all the above, just to add slightly more context from PMs (which yeah @AgniveshChaubey let's maybe stick to talking in the channel at least so others can chime in) --

I think part of what's here is meant to be an intermediate step (i.e. probably we have the same final page in mind, what's here was a small refactor on the way to getting there). More specifically:

> It should contain a table with each dialect as a row, and for each dialect it should show number of passed/errored/skipped cases for this specific implementation.

I definitely agree with this, and I think @AgniveshChaubey indeed wants to do so, I think we simply left it to a follow-up PR to make this one easier to do a first version of.

> I would expect that if we are on /dialects/draft2020-12 and then we click on a specific implementation it would navigate to /dialects/draft2020-12/implementations/go-jsonschema

+1, definitely agree with this too, and I think we discussed this @AgniveshChaubey -- though right now we have 3 "ambiguous" places someone can click when on a dialect page:

* clicking the link / name itself takes you to the implementation's issue page (which this PR changes)
* clicking the `i` button shows you metadata about the implementation
* clicking the `details` button shows you dialect-specific implementation-specific information (i.e. test failures)

To me we definitely have 2 distinct things to show, and the question is how / with what kinds of buttons and labels:

* non-dialect specific implementation info (i.e. this PR) -- so replacing the `i` modal with a page, or making the modal and a page have the same content. I think as part of this we should stop linking directly to implementations in the link (as this PR tweaks) because it's more jarring than linking to an internal page with info. If someone wants to continue on externally they should get a link on that page.
* dialect-specific implementation-specific implementation info (so "tell me what tests fail") -- i.e. changing the `details` button to link to or straight up be `/dialects/foo/implementations/bar`

So again I think we agree on the target more or less, this PR perhaps could be an intermediate step where we remove the `i` and instead link to the new page from the name, and then once we have a details page we again tweak things, not sure.

> Then the question is: how could user navigate to /implementations/go-jsonschema? Ideally, I would see it on the main navigation bar, right next to the Dialects menu.

Definitely agree with this (adding a nav bar dropdown for implementations).

> But, to render this navigation menu we would need to know all the possible implementations,

Technically it only requires reading the first *line* of all supported dialects. Even more technically, that could be done lazily after someone hovers on the menu (and even more lazily populating the menu if it turned out that worked well).

But I think even more pertinently we can of course choose to change what we write out to support this kind of thing -- so we could (at report generation time) generate a separate JSON file with extra metadata needed for the UI, like the list of *all* implementations, and then the UI could read that. This kind of thing is anyways useful for implementing #24, so we could think about doing so (it should really be reasonably simple for this specific use case).

> Moreover, can you limit the scope of this PR to include only changes related to one new path?

And yes 1000% agree with this -- @AgniveshChaubey in general in any PR do the minimal amount possible, never do 2 things at once, and certainly not until you're pretty confident about when to break that rule. So one "page" or really feature at a time.

+1 on the `pnpm` thing too though sorry I didn't make sure you'd seen #384 -- but we (me and Harrel at least) are giving a shot towards *only* using `pnpm`. That PR has short rationale, ask if anything's unclear.

> PS2 IMHO /doc path is not really something needed or even something that makes sense. Why don't use navigation link to docs page directly?

And here by the way, this might be news to both of you, but the docs page now has its own subdomain, https://docs.bowtie.report (which just shows the readthedocs docs).

I 100% agree with linking to that somewhere in the UI, though I don't think it should get mounted on some path, we should just have a link somewhere which takes you there. But where exactly should be in a separate PR.

AgniveshChaubey:
As per the feedback from both of you:

* I have removed `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>` to narrow down the scope of this PR.
* I have also removed npm lock files.
* I removed the `RunInfoModal` and shifted all of its content to `/implementations/<lang>-<impl-name>`.
* Additionally, I have added a table on the above-mentioned page which displays test results for each dialect for the specific implementation.

> I definitely agree with this suggestion of adding a nav bar dropdown for implementations.

Currently, we are already providing links to implementation-specific pages through implementation names. So, is there any specific need for adding a dropdown for implementations in the navbar?

image

AgniveshChaubey:
As per the feedback from both of you:

* I have removed `/implementations/<lang>-<implementation-name>/dialects/<dialect-name>` to narrow down the scope of this PR.
* I have also removed npm lock files.
* Removed the `RunInfoModal` and shifted all of its content to `/implementations/<lang>-<impl-name>`.
* Additionally, I have added a table on the above-mentioned page which displays test results for each dialect for the specific implementation.

> I definitely agree with this suggestion of adding a nav bar dropdown for implementations.

Currently, we are already providing links to implementation-specific pages through implementation names. So, is there any specific need for adding a dropdown for implementations in the navbar?

image

frontend/src/components/Per-ImplementationPage/DialectCompliance.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import { Card, Table } from "react-bootstrap";
+import { ReportData } from "../../data/parseReportData";
+
+interface DialectComplianceProps {
+  loaderData: ReportData;
```
harrel56, line 6: This seems to not be of `ReportData` type? it causesa bunch of TS errors below

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+                <th>Name:</th>
+                <td>{implementationDetail.name}</td>
+              </tr>
+              <tr>
+                <th>Language:</th>
+                <td>
+                  {implementationDetail.language}
+                  <span className="text-muted">
+                    {implementationDetail.language &&
+                      ` (${implementationDetail.version || ""})`}
```
harrel56, line 45: This looks like it's language version when in reality it's implementation version. I would put version as a separate row

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <Table>
+            <tbody>
+              <tr>
+                <th>Name:</th>
+                <td>{implementationDetail.name}</td>
+              </tr>
+              <tr>
+                <th>Language:</th>
+                <td>
+                  {implementationDetail.language}
```
harrel56, line 42: Let's map language to a user friendly name - check out `mapLanguage.ts`

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+      ...allImplementations,
+      ...value.runInfo.implementations,
+    };
+  });
+  Object.keys(allImplementations).forEach((key) => {
+    if (key.includes(implementationName)) {
+      implementationDetail = allImplementations[key];
+    }
+  });
+  return implementationDetail ? (
```
harrel56, line 28: This 80+ lines ternary operator is not really readable :P

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
harrel56: BTW you lost language version?

frontend/src/index.jsx
```diff
@@ -39,6 +40,22 @@ const fetchReportData = async (dialect) => {
   return parseReportData(lines);
 };
 
+const fetchAllReportData = async () => {
+  let loaderData = {};
+  const fetchPromises = Object.keys(dialectToName).map(async (draft) => {
```
harrel56, line 45: This code duplicates a lot from the function above. Why not do something like this:
```js
Object.keys(dialectToName).map(dialect => fetchReportData(dialect))
```
?
Probably changing document title should be moved to some different place then

harrel56 (commented):
Left a few more code specific comments. Gonna look into it again when you manage to fix TS compilation errors

AgniveshChaubey:
Hey @harrel56, thank you for taking a look! I will address/respond to your comments tomorrow after my exam!

AgniveshChaubey:
Hey @harrel56, thank you for taking a look! Right now, I am preparing for my exams. I will address/respond to your comments tomorrow after my exam!"

harrel56:
Yes yes, of course, there is no rush :)

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
```diff
@@ -0,0 +1,108 @@
+import { Card } from "react-bootstrap";
+import { useLoaderData, Link } from "react-router-dom";
+import { Table, Container } from "react-bootstrap";
+import { ReportData, ImplementationMetadata } from "../../data/parseReportData";
+// @ts-ignore
+import LoadingAnimation from "../../components/LoadingAnimation";
+import DialectCompliance from "./DialectCompliance";
+
+export const PerImplementationPage = () => {
```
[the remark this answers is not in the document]
AgniveshChaubey: Yes, it seems reasonable!

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
AgniveshChaubey: my bad! Fixed :)

AgniveshChaubey: > This looks like it's language version when in reality it's implementation version.

my bad! corrected now :)

> BTW you lost language version?

Added it in separate row.

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
AgniveshChaubey: Done

frontend/src/components/Per-ImplementationPage/PerImplementationPage.tsx
AgniveshChaubey: Moved those lines into its own component for improved readability! 😄
```js
  return implementationDetail ? (
    <ReportComponent />
  ) : (
    <LoadingAnimation />
  );
```

frontend/src/components/Per-ImplementationPage/DialectCompliance.tsx
```diff
@@ -0,0 +1,73 @@
...
+  implementationsDetail: {
+    [x: string]: any;
```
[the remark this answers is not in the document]
AgniveshChaubey: It contains an array of strings.

Fixed now.

frontend/src/components/Per-ImplementationPage/DialectCompliance.tsx
AgniveshChaubey: Fixed

frontend/src/index.jsx
AgniveshChaubey: Done!

> Probably changing document title should be moved to some different place then

This shouldn't cause any problems since in implementation-specific pages, I've overwritten the `title` with the implementation's name.

AgniveshChaubey:
Hi @harrel56, @Julian I've addressed all the comments mentioned above. Whenever you have time, kindly review it once again and let me know if it's ready to be merged.

Julian:
Perfect! Will have another look in a bit.

harrel56:
Hey, the TS compilation is still failing (the `Frontend UI build / build (pull_request)` job from checks). Could you fix it first? You can also run the build locally via `pnpm run build`

AgniveshChaubey:
Fixed!

AgniveshChaubey:
A slight change--
On the implementation page, the **section-navigations** (`#summary` , `#details`, etc.) were causing an error when clicked. Therefore, I have hidden them on those pages.

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
```diff
@@ -0,0 +1,126 @@
...
+      implementationDetail = allImplementations[key];
+      document.title = `Bowtie - ${implementationDetail.name}`;
+    }
+  });
+
+  const ReportComponent = () => {
+    return (
+      <Container className="p-4">
+        <Card className="mx-auto mb-3 w-75">
+          <Card.Header>Runtime Info</Card.Header>
```
harrel56, line 44: `Implementation info`

frontend/src/components/NavBar.jsx
```diff
@@ -41,21 +51,25 @@ const NavBar = () => {
...
-                <Link className="nav-link" to={{ hash: "summary" }}>
-                  Summary
-                </Link>
-              </li>
-              <li className="nav-item">
-                <Link className="nav-link" to={{ hash: "cases" }}>
-                  Details
-                </Link>
-              </li>
+              {isDialectPage ? (
```
harrel56, line 54: I think more common pattern is to use `isDialectPage && ...`, at least there is no null at the end

frontend/src/components/NavBar.jsx
```diff
@@ -1,21 +1,31 @@
...
-  const { hash, key } = useLocation();
+  const { pathname, hash, key } = useLocation();
 
   useEffect(() => {
     if (hash) {
       const targetElement = document.getElementById(hash.substring(1));
       targetElement?.scrollIntoView({ behavior: "smooth" });
     }
-  }, [key, hash]);
+    const matchPath = () => {
```
harrel56, line 20: You could get rid of all this logic to compute `isDialectPage` and use other router APIs, something lime this (outside of `useEffect` of course):
```js
  const rootMatch = useMatch('/')
  const dialectsMatch = useMatch('/dialects/*')
  const isDialectPage = matches1 || matches2
```

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
```diff
@@ -0,0 +1,126 @@
...
+    dialects: [],
+    homepage: "",
+    name: "",
+    version: "",
+    language: "",
+    issues: "",
+    image: "",
+  };
+
+  let allImplementations: { [key: string]: ImplementationMetadata } = {};
```
harrel56, line 26: This is my biggest concern I guess - you transform data in this component and also in `DialectCompliance`. Ideally, all this work could be done way up in a router loader.

Also please have a look at `calculateTotals` function in `SummaryTable.tsx` (probably should've put it somewhere else :P). I think if in a router loader we could get the data to look like something like this:
```ts
{
  [key: string]: Totals
}
```
plus the implementation metadata, then it would be all we need?

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
```diff
@@ -0,0 +1,126 @@
...
+                          <Link to={dialect}>{dialect}</Link>
+                        </li>
+                      ))}
+                    </ul>
+                  </td>
+                </tr>
+                <tr>
+                  <th>Image:</th>
+                  <td>
+                    <Link to={implementationDetail.image}>
```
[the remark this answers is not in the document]
AgniveshChaubey: removing the link seems to be a better option :P

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <DialectCompliance
+          loaderData={loaderData}
+          implementationsDetail={implementationDetail}
+          implementationName={implementationName}
+        />
+      </Container>
+    );
+  };
+
+  return implementationDetail ? <ReportComponent /> : <LoadingAnimation />;
```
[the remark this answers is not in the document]
AgniveshChaubey: right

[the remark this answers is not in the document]
AgniveshChaubey: yes

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
AgniveshChaubey: Done!

Now, the loader data looks like:
```js
{
    "name": "...",
    "homepage": "...",
    "issues": "...",
    ...
    "results": {
        "drafts": Totals
    }
}
```

AgniveshChaubey:
Hi @harrel56 @Julian, I have addressed all the comments, and as per the suggestions, I have moved the data transformation logic to higher-level component. Kindly have a look and let me know if any further changes are required.

_If possible, kindly review this at your earliest convenience. This would enable me to begin work on the GSoC final submission report after merging the PR. :)_

harrel56:
Nice! I don't know when I'll be able to thoroughly review these new changes - at a first glance I can see some things that could still be improved. But certainly it looks a lot better - so if there is any urgency for merging this I'll be fine with it. Assuming ofc that it builds and works as expected :D

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { Card, Table } from "react-bootstrap";
+import { ImplementationMetadata } from "../../data/parseReportData";
+
+const DialectCompliance: React.FC<{ specificData: ImplementationMetadata }> = ({
+  specificData,
+}) => {
+  const dialectMapping: { [key: string]: string } = {
+    "draft2020-12": "https://json-schema.org/draft/2020-12",
```
Julian, line 9: This seems like an oddly specific place for this mapping.

Does it not exist in the frontend yet? If not, let's move it somewhere less specific to this page.

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
```diff
@@ -0,0 +1,119 @@
...
+import LoadingAnimation from "../LoadingAnimation";
+import DialectCompliance from "./DialectCompliance";
+import { mapLanguage } from "../../data/mapLanguage";
+
+export const ImplementationReportView = () => {
+  const allImplementations = useLoaderData() as {
+    [key: string]: ImplementationMetadata;
+  };
+
+  const pathSegments = window.location.href.split("/");
```
Julian, line 15: I haven't tried to fully understand what this does yet, so it's possible that this comment won't apply, but treating URLs as strings is a code smell. I believe we added a dependency on URI.js already, have a look and see whether it already does what you're doing manually here (which seems like maybe it's just pulling off the last path segment?).

frontend/src/data/parseReportData.ts
```diff
@@ -118,8 +196,9 @@ export interface ImplementationMetadata {
   name: string;
   version?: string;
   dialects: string[];
-  homepage?: string;
+  homepage: string;
```
Julian, line 199: Homepage is left optional in the current schema:

https://github.com/bowtie-json-schema/bowtie/blob/main/bowtie/schemas/io-schema.json#L70

Probably it shouldn't be, so feel free to constrain it there, but be careful when making changes that they match the schema.

frontend/src/index.jsx
```diff
@@ -8,16 +8,20 @@ import ThemeContextProvider from "./context/ThemeContext";
...
+import { ImplementationReportView } from "./components/ImplementationReportView/ImplementationReportView";
 
 const reportHost =
   import.meta.env.MODE === "development"
     ? "https://bowtie.report"
     : window.location.href;
 const reportUri = new URI(reportHost).directory(import.meta.env.BASE_URL);
 
-const dialectToName = {
+export const dialectToName = {
```
Julian, line 24: Given this and the mapping (which I mentioned above I think belongs in some more central place) I'd probably recommend creating a `Dialect` object at some point containing the various things we want to capture about them (human readable names, mappings to and from their URIs, etc.). Feel free to leave that for now though.

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
harrel56: Oh yes, this path parameter should be retrieved from react router api, believe there is a hook for that

frontend/src/index.jsx
```diff
@@ -59,6 +72,11 @@ const router = createHashRouter([
         path: "/local-report",
         Component: DragAndDrop,
       },
+      {
+        path: "/implementations/:langImplementation",
+        Component: ImplementationReportView,
+        loader: async () => fetchAllReportData(),
```
Julian, line 78: This makes a thunk but we have one already no, does `loader: fetchAllReportData` not work in Javascript?

frontend/src/DialectReportView.tsx
```diff
@@ -1,15 +1,19 @@
 // @ts-ignore
-import CasesSection from "./components/Cases/CasesSection";
+import CasesSection from "./components/Cases/CasesSection.jsx";
```
Julian, line 2: What's happening here?

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -0,0 +1,57 @@
...
+            <tr>
+              <th rowSpan={2} scope="col" className="text-center align-middle">
+                Supported Dialects
+              </th>
+              <th colSpan={3} className="text-center">
+                Tests
+              </th>
+            </tr>
+            <tr className="text-center">
+              <th>Unsuccessful</th>
```
Julian, line 32: The terminology here is confusing to me (though it's undoubtedly confusing to everyone, and will require some tooltips to explain).

But in the backend, the term "unsuccessful" is used to refer to tests which *either* fail (i.e. have a result different than the expected one) or error (i.e. the test caused an exception in the harness).

That's distinct from "failures" which are just the former.

Which does this mean here? Combined or just failures?

Julian:
(You'll also have to merge main again, I was poking at trying to fix #411 but have now had to back that change out.)

frontend/src/index.jsx
AgniveshChaubey: Yes, it works. Fixed now!

frontend/src/DialectReportView.tsx
AgniveshChaubey: I don't actually know how the extension got added. 😅
Removed

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
AgniveshChaubey: It should be `Failed`, actually. (Both **failed** and **errored** tests are displayed separately here)

frontend/src/index.jsx
AgniveshChaubey: Done!

AgniveshChaubey:
Hey @Julian, done with the changes!

frontend/src/data/parseReportData.ts
AgniveshChaubey: I believe we only need to include `homepage` in the **required** array. Is there anything else we should modify to make `homepage` a required field?

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -0,0 +1,49 @@
...
+              <th>Errored</th>
+            </tr>
+          </thead>
+          <tbody className="table-group-divider">
+            {Object.entries(specificData.results).map(
+              ([draft, result], index) => {
+                return (
+                  <tr key={index}>
+                    <td>{Dialect.dialectMapping[draft]}</td>
+                    <td className="text-center">{result.unsuccessfulTests}</td>
```
Julian, line 35: This still looks confusing -- the field is `.unsuccessfulTests`, but you're sure it's failure and not unsuccessful?

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { Card, Table } from "react-bootstrap";
+import { ImplementationMetadata } from "../../data/parseReportData";
+import { Dialect } from "../../data/Dialect";
+
+const DialectCompliance: React.FC<{ specificData: ImplementationMetadata }> = ({
```
Julian, line 6: Perhaps `implementation` is a better name than `specificData`.

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -0,0 +1,49 @@
...
+            </tr>
+            <tr className="text-center">
+              <th>Failed</th>
+              <th>Skipped</th>
+              <th>Errored</th>
+            </tr>
+          </thead>
+          <tbody className="table-group-divider">
+            {Object.entries(specificData.results).map(
+              ([draft, result], index) => {
```
Julian, line 31: I think we should avoid `draft` as a variable name (or certainly visible name) -- I tried best I could to avoid it everywhere, but basically the specification doesn't like this term any more. I'd suggest either `dialectName` or `version` or something like that.

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export const ImplementationReportView = () => {
+  const allImplementations = useLoaderData() as {
+    [key: string]: ImplementationMetadata;
+  };
+
+  const { langImplementation } = useParams();
+  const implementationName = langImplementation ?? "";
+
+  function filterImplementation(
```
Julian, line 18: I have to stare at this longer to understand what it does, but it seems a bit like it's doing a somewhat odd kind of pivot (taking data keyed by something and pivoting it to be keyed by something else).

In general it'd be easier to follow some of these if you added a few docstrings I think on what these are trying to do -- the function name could be a bit clearer too -- this is filtering what, implementation metadata? To what? Does `ImplementationMetadata` represent *all* implementations' metadata? And this filters to just one specific implementation? (That sounds a bit odd to me, I'd instead expect not to have an object representing *all* metadata, to instead simply have that be a mapping / JS object from implementation ID to an object representing *one* implementation's metadata. But perhaps that isn't something new to this PR.)

frontend/src/components/Summary/ImplementationRow.jsx
```diff
@@ -1,24 +1,22 @@
...
-import { RunTimeInfoModal } from "../Modals/RunTimeInfoModal";
 import { mapLanguage } from "../../data/mapLanguage";
+import { NavLink } from "react-router-dom";
 
 const ImplementationRow = ({ cases, implementation }) => {
   const [showDetails, setShowDetails] = useState(false);
-  const [showRunTimeInfo, setShowRunTimeInfo] = useState(false);
 
+  const getPath = () => {
+    const pathSegment = implementation.id.split("/");
```
Julian, line 11: Can we encapsulate this? Perhaps with a property / function on `implementation`? What type of object is that? Can we teach it to know what this value is?

frontend/src/data/Dialect.ts
```diff
@@ -0,0 +1,27 @@
+export const Dialect: Dialect = {
+  dialectToName: {
+    "draft2020-12": "Draft 2020-12",
+    "draft2019-09": "Draft 2019-09",
+    draft7: "Draft 7",
+    draft6: "Draft 6",
+    draft4: "Draft 4",
+    draft3: "Draft 3",
+  },
+  dialectMapping: {
```
Julian, line 10: These don't look right. Please double check them. Specifically, some JSON Schema dialect IDs have empty fragments (`#`) and some do not. Make sure you've copied them exactly as-is from the Python file or from the specification page.

frontend/src/data/parseReportData.ts
Julian: > I believe we only need to include homepage in the required array.

That seems right yep.

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
AgniveshChaubey: Yes, I am sure. Okay, I have renamed `unsuccessfulTests` to `failedTests` to avoid confusion.

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
AgniveshChaubey: > This is filtering what, implementation metadata?

Yes, it returns metadata of the page-specific implementation.

> Does ImplementationMetadata represent all implementations' metadata?

Nope. It represents just one specific implementation.

> I'd instead expect not to have an object representing all metadata, but instead, to simply have that be a mapping / JS object from implementation ID to an object representing one implementation's metadata.

The current structure represents the same thing as you mentioned above.

See this:  
image

AgniveshChaubey:
Updated!
_(Apologies for pinging you multiple times today, @julian)_

AgniveshChaubey:
Updated!
_(And apologies for pinging you multiple times today, @julian)_

frontend/src/components/Summary/ImplementationRow.jsx
AgniveshChaubey: I've now converted this file to `tsx` in order to add **Type** annotations.

frontend/src/data/parseReportData.ts
Julian: This is still outstanding right?

frontend/src/components/ImplementationReportView/ImplementationReportView.tsx
Julian: I still don't really understand then what / why we're iterating over `Object.keys` and filtering that, that's a code smell. Why isn't this simply a key-value lookup? Why iterate over a bunch of keys trying to find the implementation we're looking for?

frontend/src/components/Summary/ImplementationRow.jsx
Julian: That wasn't precisely what I meant, though seems like a good idea anyhow. What I meant is "can we put this function on a class or javascript object which represents an `Implementation`"? It seems likely to be an operation we may want to do in a bunch of places in the code base. Specifically I think someone should want to have a JavaScript accessor which lets them write `implementation.imageName` (or some appropriate name) which returns this value, but that doesn't involve the caller/user knowing how to do this parsing.

Julian:
Going to merge this and tidy up a few of the remaining comments, we can address the rest with a follow-up or further commits, but just trying to make sure this doesn't sit here. Thanks again.

PR #404, opened 2023-08-19
feat: add a button in the UI to access documentation page

AgniveshChaubey:
image

----
:books: Documentation preview :books:: https://bowtie-json-schema--404.org.readthedocs.build/en/404/

### DIFF
--- a/frontend/src/components/NavBar.jsx
+++ b/frontend/src/components/NavBar.jsx
@@ -111,6 +111,15 @@ const NavBar = () => {
               </li>
             </ul>
           </div>
+          <button id="docs" className="btn btn-sm btn-secondary border-0 me-1">
+            <Link
+              className="nav-link"
+              to="https://docs.bowtie.report/"
+              target="blank"
+            >
+              Docs
+            </Link>
+          </button>
           <button
             id="theme-toggler"
             className="btn border-0 me-1"


frontend/src/components/NavBar.jsx
```diff
@@ -111,6 +111,15 @@ const NavBar = () => {
               </li>
             </ul>
           </div>
+          <button id="docs" className="btn btn-sm btn-secondary border-0 me-1">
```
harrel56, line 114: Is `id` attribute necessary?

harrel56 (approved):

frontend/src/components/NavBar.jsx
AgniveshChaubey: Not necessary, but it's good practice to keep it.

harrel56: Why is it a good practice? I think I've never heard of that. What are the advantages?

AgniveshChaubey: Although the button text in our code indicates its purpose for documentation, the 'id' attribute can also serve as a form of documentation. It's beneficial for SEO as well, even though we are not currently focusing on that aspect in our case.

I don't see any issues with it, but I'll remove it if you want :)

harrel56: TBH I tried to find something on the web that confirms ids affect SEO, but without much luck - if you have any source on that I'll be happy to have a look.

PS I'm genuinely curious about your motives here and not trying to push anything, so there is no need to remove this id

AgniveshChaubey: These points were discussed in a course (**Full-Stack Web Development with React Specialization**) offered by Hong Kong University on Coursera. TBH, I didn't cross-verify them as SEO wasn't my primary focus. I didn't place significant emphasis on it and simply included the 'id' attribute based on the information from that course. (considering it as a good practice)😃

PR #413, opened 2023-08-22
Fix Hyperjump test harness for file URIs

jdesrosiers:
I noticed that a [fix](https://github.com/bowtie-json-schema/bowtie/pull/325) I made previously to my test harness didn't actually work the way I thought it did and wasn't working for draft-04. This appears to fix the problem.

(I managed to get it to run this time and verify that it works. I needed to use docker instead of podman to build the images. When I built the image with podman like the documentation says, bowtie couldn't find the image. When I built with docker instead, everything worked as expected.)

----
:books: Documentation preview :books:: https://bowtie-json-schema--413.org.readthedocs.build/en/413/

### DIFF
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -116,10 +116,9 @@ const cmds = {
       });
     } else {
       try {
-        const idToken =
-          args.dialect === "http://json-schema.org/draft-04/schema#"
-            ? "id"
-            : "$id";
+        const idToken = dialect === "http://json-schema.org/draft-04/schema#"
+          ? "id"
+          : "$id";
         const host = testCase.schema?.[idToken]?.startsWith("file:")
           ? "file://"
           : "https://example.com";
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -116,9 +116,8 @@ const cmds = {
       });
     } else {
       try {
-        const idToken = dialect === "http://json-schema.org/draft-04/schema#"
-          ? "id"
-          : "$id";
+        const idToken =
+          dialect === "http://json-schema.org/draft-04/schema#" ? "id" : "$id";
         const host = testCase.schema?.[idToken]?.startsWith("file:")
           ? "file://"
           : "https://example.com";


Julian:
Thanks! Can you share which commands weren't working? I've never used docker, I use podman exclusively.

jdesrosiers:
Build with podman

```
> podman build --quiet -f Dockerfile -t localhost/js-hyperjump .
npm notice 
npm notice New patch version of npm available! 9.8.0 -> 9.8.1
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.8.1>
npm notice Run `npm install -g npm@9.8.1` to update!
npm notice 
d3c8df58cfec6ea2ba529251b270b40f661aa450de0859c956de45a4003b8e4a
```

Bowtie doesn't find the image
```
> bowtie run -i localhost/js-hyperjump <<EOF
    {"description": "test case 1", "schema": {}, "tests": [{"description": "a test", "instance": {}}] }
    {"description": "test case 2", "schema": {"const": 37}, "tests": [{"description": "not 37", "instance": {}}, {"description": "is 37", "instance": 37}] }
EOF
2023-08-22 11:57.22 [info     ] Will speak                     dialect=https://json-schema.org/draft/2020-12/schema
2023-08-22 11:57.22 [error    ] Not a known Bowtie implementation. [localhost/js-hyperjump] 
2023-08-22 11:57.22 [error    ] No implementations started successfully!
```

Build with docker
```
> docker build --quiet -f Dockerfile -t localhost/js-hyperjump .
sha256:4250006298fffbf2ef199ce4246b73fd3f894abb0ba783c7f679a2231df80caf
```

Bowtie runs as expected
```
> bowtie run -i localhost/js-hyperjump <<EOF                    
    {"description": "test case 1", "schema": {}, "tests": [{"description": "a test", "instance": {}}] }
    {"description": "test case 2", "schema": {"const": 37}, "tests": [{"description": "not 37", "instance": {}}, {"description": "is 37", "instance": 37}] }
EOF
2023-08-22 12:00.57 [info     ] Will speak                     dialect=https://json-schema.org/draft/2020-12/schema
{"started": "2023-08-22T19:00:57.647839+00:00", "bowtie_version": "2023.8.11.dev13+g56cc16b", "dialect": "https://json-schema.org/draft/2020-12/schema", "implementations": {"localhost/js-hyperjump": {"language": "javascript", "name": "hyperjump-jsv", "version": "1.5.1", "homepage": "https://json-schema.hyperjump.io/", "issues": "https://github.com/hyperjump-io/json-schema-validator/issues", "dialects": ["https://json-schema.org/draft/2020-12/schema", "https://json-schema.org/draft/2019-09/schema", "http://json-schema.org/draft-07/schema#", "http://json-schema.org/draft-06/schema#", "http://json-schema.org/draft-04/schema#"], "os": "linux", "os_version": "6.4.11-200.fc38.x86_64", "language_version": "v20.5.1", "image": "localhost/js-hyperjump"}}, "metadata": {}}
{"case": {"description": "test case 1", "schema": {}, "tests": [{"description": "a test", "instance": {}, "comment": null, "valid": null}], "comment": null, "registry": null}, "seq": 1}
{"implementation": "localhost/js-hyperjump", "seq": 1, "results": [{"valid": true}], "expected": [null]}
{"case": {"description": "test case 2", "schema": {"const": 37}, "tests": [{"description": "not 37", "instance": {}, "comment": null, "valid": null}, {"description": "is 37", "instance": 37, "comment": null, "valid": null}], "comment": null, "registry": null}, "seq": 2}
{"implementation": "localhost/js-hyperjump", "seq": 2, "results": [{"valid": false}, {"valid": true}], "expected": [null, null]}
2023-08-22 12:00.57 [info     ] Finished                       count=2
{"did_fail_fast": false}
```

When I first installed bowtie, I had podman installed, but not docker. I run Fedora 38 where podman is installed by default. When trying to run bowtie, it blew up with the message, "Missing valid docker_host.Either DOCKER_HOST or local sockets are not available". I installed docker (actually moby-engine) and the error went away, but the podman builds were still invisible to bowtie. Since the previous error indicated that bowtie liked docker and podman wasn't a sufficient substitute, I tried building with docker instead and then bowtie found the image and ran as expected.

jdesrosiers:
That worked. I found documentation on how to set DOCKER_HOST at https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md

Julian:
Ah awesome ok will add a note to https://docs.bowtie.report/en/stable/implementers/ -- thanks for the debug info, glad it works now.

jdesrosiers:
The tutorial change looks good. Just one FYI, the docs you link to doesn't have the right solution for linux users (or at least so incomplete that I needed to find other resources to solve my problem). I can't say for the other operating systems. In any case, it didn't take too long to find the solution once I knew what I was looking for.

Julian:
Huh interesting -- thanks for pointing that out. As in `podman info --format '{{.Host.RemoteSocket.Path}}'` doesn't give you the right path? What does it give? The macOS tab at least it appears to give the right thing.

I don't want to link to the one you found, as this doesn't really have to do with socket activation, but there should definitely be one that's appropriate -- if the one I linked to is wrong I'll maybe open a bug on their repo but yeah what didn't work there so I can tell them?

jdesrosiers:
It's the right path, but it needed to be expressed as a `unix:` URI, not a path.

```
export DOCKER_HOST="unix://$(podman info --format '{{.Host.RemoteSocket.Path}}')"
```

Julian:
Ahhhh OK that indeed seems like a doc bug on Podman's part. And one I swear I've even realized previously at some point but forgot until now. Will file it there.

Julian:
Fix sent: containers/podman-desktop#3655.

PR #422, opened 2023-08-27
implementation for openapi-processor json schema validator

hauner:
it runs, finally :-)

this adds an implementation for https://github.com/openapi-processor/openapi-parser/tree/master/json-schema-validator

----
:books: Documentation preview :books:: https://bowtie-json-schema--422.org.readthedocs.build/en/422/

### DIFF
--- a/implementations/dotnet-jsonschema-net/bowtie_json_everything.csproj
+++ b/implementations/dotnet-jsonschema-net/bowtie_json_everything.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="JsonSchema.Net" Version="5.2.0" />
+    <PackageReference Include="JsonSchema.Net" Version="5.2.1" />
   </ItemGroup>
 
   <ItemGroup>


Julian:
Amazing, congrats! I'll try to give this a closer look. Just acking I've seen it.

On CI -- one thing you can ignore (there's a "known" issue I don't know how to fix well quite yet which has to do with GitHub runners losing docker images during a workflow -- specifically, you can ignore the failure for [building your image](https://github.com/bowtie-json-schema/bowtie/actions/runs/5996532180/job/16261226848?pr=422) which has to do with the fact that builds are currently confused when it's the first time they're building an image. I'll check manually that your image builds.)

The other CI failure though you should address (the pre-commit one), it's saying that there's a file missing a newline at the end. You can either fix that manually or install and run `pre-commit run` locally and it will automatically fix it.

More comments to come, and thanks again for the PR.

implementations/java-openapiprocessor/gradlew
```diff
@@ -0,0 +1,244 @@
...
+darwin=false
+nonstop=false
+case "$( uname )" in                #(
+  CYGWIN* )         cygwin=true  ;; #(
+  Darwin* )         darwin=true  ;; #(
+  MSYS* | MINGW* )  msys=true    ;; #(
+  NONSTOP* )        nonstop=true ;;
+esac
+
+CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
```
Julian, line 117: I'd very much like to avoid committing any large binary files to the repo. Do you know how we can avoid adding this JAR?

Also, we don't really need cross-platform support I think, because your image is only going to run in the Linux container we're building it for. If you want to add *slightly* more so you can also run it natively on your host machine I'm probably fine with that, but there's really quite a few files in here so I'd definitely like to remove as much as isn't really needed.

implementations/java-openapiprocessor/src/main/kotlin/commands/Start.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun run(request: StartRequest): Map<String, Any> {
+        assert(request.version == 1)
+
+        return mapOf(
+            "ready" to true,
+            "version" to 1,
+            "implementation" to mapOf(
+                "language" to "java",
+                "language_version" to "11",
```
Julian, line 18: Can we not get this information slightly more specifically (e.g. I assume there's a minor version at least?) and dynamically? This is going to go out of date if we change the image.

implementations/java-openapiprocessor/src/main/kotlin/commands/Start.kt
```diff
@@ -0,0 +1,43 @@
...
+                "name" to "io.openapiprocessor.json-schema-validator",
+                "version" to "2023.3-SNAPSHOT",
```
Julian, line 20: Same here, is this not available dynamically?

implementations/java-openapiprocessor/gradlew
hauner: I can remove it.

It is only required for a local build without docker. The docker image has a pre-installed gradle and doesn't need it.

implementations/java-openapiprocessor/src/main/kotlin/commands/Start.kt
hauner: the java version provides the minimum version required to use it (the version it is compiled with). This is usually given just with the major version.

more correct would be to say something like `"language_version" to "11 or newer"` if it fits.

implementations/java-openapiprocessor/gradlew
Julian: You'll have to force push to actually remove it, your commits delete the file but it will still be in the repo history if this is merged (and will increase the size of clones)

implementations/java-openapiprocessor/src/main/kotlin/commands/Start.kt
Julian: This field is for Bowtie, it's meant to show users what version is the one being run -- e.g., imagine someone observes different behavior from what Bowtie reports than what they see themselves.

In that case they're not interested in the versions your library is saying it supports, they want to know exactly what's running so they can reproduce if needed.

hauner: oh,  *that* version :-)

implementations/java-openapiprocessor/gradlew
hauner: fixed here https://github.com/bowtie-json-schema/bowtie/pull/439

Issue #462, opened 2023-09-11
The Java implementation(s) container images are huge


Julian:
Here's all of the current implementation images, sorted by size:

```
ghcr.io/bowtie-json-schema/lua-jsonschema                    latest         ada98b1dee70  38 hours ago  12.5 MB
ghcr.io/bowtie-json-schema/rust-boon                         latest         537e5358b2e8  38 hours ago  15.5 MB
ghcr.io/bowtie-json-schema/cpp-valijson                      latest         c933b4a17f90  38 hours ago  19.6 MB
ghcr.io/bowtie-json-schema/rust-jsonschema                   latest         e8826839973f  37 hours ago  21.2 MB
ghcr.io/bowtie-json-schema/go-jsonschema                     latest         2c54a892a7bb  38 hours ago  21.4 MB
ghcr.io/bowtie-json-schema/go-gojsonschema                   latest         9d4d4cf81827  38 hours ago  27.1 MB
ghcr.io/bowtie-json-schema/python-fastjsonschema             latest         e9885b29f040  38 hours ago  72.1 MB
ghcr.io/bowtie-json-schema/python-jschon                     latest         e651d5bc9169  38 hours ago  72.8 MB
ghcr.io/bowtie-json-schema/php-opis-json-schema              latest         37832c88e85e  38 hours ago  76.6 MB
ghcr.io/bowtie-json-schema/python-jsonschema                 latest         986e8813992c  38 hours ago  81.3 MB
ghcr.io/bowtie-json-schema/ruby-json_schemer                 latest         e5e850d96c76  38 hours ago  89.8 MB
ghcr.io/bowtie-json-schema/dotnet-jsonschema-net             latest         82dbabb4ba9d  38 hours ago  95.8 MB
ghcr.io/bowtie-json-schema/js-ajv                            latest         c73ebaf87d90  38 hours ago  189 MB
ghcr.io/bowtie-json-schema/js-hyperjump                      latest         4c5168da242d  38 hours ago  190 MB
ghcr.io/bowtie-json-schema/ts-vscode-json-languageservice    latest         9d17bdbe8dd5  38 hours ago  197 MB
ghcr.io/bowtie-json-schema/clojure-json-schema               latest         0b9aa8704bdc  38 hours ago  220 MB
ghcr.io/bowtie-json-schema/kotlin-kmp-json-schema-validator  latest         9e110e4ad454  38 hours ago  694 MB
[...]

PR #464, opened 2023-09-13
Reduce JVM images size

harrel56:
https://github.com/bowtie-json-schema/bowtie/issues/462

Managed to reduce sizes roughly to ~130MB. Certainly the requirement of supporting arm architecture is not helping, but without it maybe I could strip 30 more MBs, so I guess it's not such a big deal.

As suggested in some linked issues with gradle alpine builds - I decided to use bellsoft vendor for the jdk because they provide alpine (and many other varations) images for arm arch. Honestly I never used bellsoft jdks (liberica) but a jdk vendor should not matter at all. The thing that matters the most is the license - in this case it's GPL, hopefully it's fine? I feel like the most common vendor would be temurin (successor of AdoptOpenJDK), if they finally sort out the issue with arm images, then maybe it would be worth to switch.

Also changed all java versions to 17 and this was kind of more risky change - but I generated reports with new images and compared it with originals and it looks the same

image

----
:books: Documentation preview :books:: https://bowtie-json-schema--464.org.readthedocs.build/en/464/


implementations/java-openapiprocessor/Dockerfile
```diff
@@ -1,6 +1,9 @@
...
+FROM gradle:8.3.0-jdk17 AS builder
+WORKDIR /opt/app
 COPY . .
 RUN gradle installDist --no-daemon
 RUN chmod +x ./build/install/bowtie/bin/openapiprocessor
-CMD ["/app/build/install/bowtie/bin/openapiprocessor"]
+
+FROM bellsoft/liberica-openjdk-alpine:17
+COPY --from=builder /opt/app/build /opt/app/build
+CMD ["./opt/app/build/install/bowtie/bin/openapiprocessor"]
```
Julian, line 9: Presumably this is trying to be `/opt/app/etc/...` right? And same below?

implementations/kotlin-kmp-json-schema-validator/build.gradle.kts
```diff
@@ -12,6 +12,11 @@ tasks.withType<KotlinCompile> {
     }
 }
 
+java {
+    sourceCompatibility = JavaVersion.VERSION_11
```
Julian, line 16: (And here)

Julian:
> The thing that matters the most is the license - in this case it's GPL, hopefully it's fine? I feel like the most common vendor would be temurin (successor of AdoptOpenJDK), if they finally sort out the issue with arm images, then maybe it would be worth to switch.

Yeah would definitely be nice if/when temurin works.

But fine with me for now if it works!

implementations/java-openapiprocessor/Dockerfile
harrel56: hmm, I don't know what you mean - `openapiprocessor` is a sh script produces by the build, which basically runs `java -jar` underneath

Julian: Isn't this going to look in `/usr/app/opt/app/build/...`? Is that where the build outputs?

implementations/java-openapiprocessor/build.gradle.kts
```diff
@@ -12,8 +14,15 @@ repositories {
     }
 }
 
-kotlin {
-    jvmToolchain(libs.versions.build.jdk.get().toInt())
+tasks.withType<KotlinCompile> {
+    kotlinOptions {
+        jvmTarget = "11"
```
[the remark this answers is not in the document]
harrel56: Honestly I don't know why they decided to set target for 11 and not 17. I guess it shouldn't matter at all it's just the bytecode version for harness and not for their lib. I think generally any code compiled with older jdk version should work without any issues on newer jdk versions. I could just make it to compile do jdk 17 bytecode, but I don't know :D

implementations/java-openapiprocessor/Dockerfile
harrel56: ahh, now I get it. Not really experienced linux user but I thought that to run any file you need to start with `./` - is it gonna work without?

Julian: No, that's only if you want a path relative to the current working directory.

If the working directory is `/`, then `./foo/bar` is `/foo/bar`.

If the working directory is `/bar/baz`, then `./foo/bar/` is `/bar/baz/foo/bar`, so this works basically "accidentally" because the working directory happens to be `/` -- but you don't want that here, you've explicitly copied the binary to the path `/opt/app/build/whatever`, so that's always where you want to run, so you should just have:

```suggestion
CMD ["/opt/app/build/install/bowtie/bin/openapiprocessor"]
```

essentially.

implementations/java-openapiprocessor/build.gradle.kts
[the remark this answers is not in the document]
Julian: I would assume (again not knowing anything and possibly not being arsed to check -- which means feel free to ignore as well!) that compiling for older JVMs probably means forgoing e.g. optimizations which may exist at the bytecode layer but only work on newer ones.

I.e. I'd expect compiling for older JVMs increases compatibility and potentially decreases performance. And that here since we're running in a single fixed JVM version that compiling for the newest one possible is the smartest.

But yeah I know nothing, I'm just guessing -- it likely doesn't matter. If it works when you change this to 17 I'd personally go with that, but if it doesn't seem to work for even trivial reasons then yeah leave it and ignore.

implementations/java-openapiprocessor/Dockerfile
harrel56: nice, thanks for clarification - changing it now

implementations/java-openapiprocessor/build.gradle.kts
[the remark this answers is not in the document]
harrel56: Yeah, probably your reasoning about performance is correct. At least it would be for plain java, I'm not sure if this also the same for kotlin. Nevertheless I agree that compiling to newest bytecode would be best. Gonna try it out later :)

Julian:
This LGTM assuming you've now checked that 17 works -- feel free to merge when CI passes, and thanks again!

Julian:
(Separately we should address why the heck build times are so long, I'm sure there might be one or two things we could do there, but certainly in a follow-up)

harrel56:
> This LGTM assuming you've now checked that 17 works -- feel free to merge when CI passes, and thanks again!

I don't expect any surprises but still gonna try to generate report with new images on my fork. Probably just testing it locally would be sufficient but I somehow broke my python env and don't have patience to deal with it now :D

> (Separately we should address why the heck build times are so long, I'm sure there might be one or two things we could do there, but certainly in a follow-up)

Yeah, rust build feels like forever - but I personally have no ideas how to approach this

Julian:
> Yeah, rust build feels like forever - but I personally have no ideas how to approach this

Yes, that one is definitely the long horse (and I think involves looking at `cargo build --timings` and seeing whether we can use GitHub actions cache to cache built dependencies) -- but here I meant just for the JVM builds, I have no idea why the Kotlin build is so long, Rust has a reputation for being obscenely long to compile but I never heard Kotlin being in that group (maybe I know nothing about Kotlin though -- probably that!).

harrel56:
Hmm, maybe these kotlin images could be built the same way as java libs (idk if it's feasible) because even building them locally  is time consuming. I mean java builds are lengthy too but still looks faster. One thing I noticed is that arm image builds are significantly slower - I wonder if it's the case for all images or just some quirk of gradle?

Julian:
I'd assume that (ARM) may also have to do with what the host architecture is (i.e. whether stuff is cross-compiling or not) but I haven't looked carefully / recently.

harrel56:
> I'd assume that (ARM) may also have to do with what the host architecture is (i.e. whether stuff is cross-compiling or not) but I haven't looked carefully / recently.

oh right, that's sounds totally reasonable as the images are built on linux based runner. Didn't think about it

harrel56:
Alright, eventually did test it locally and it looks fine. Merging then

PR #469, opened 2023-09-15
Fix failing smoke tests for java images

harrel56:
Seems that `java.vendor.version` is optional property and it's not provided in liberica. No idea how all images managed to build on my fork - must have mixed something up I guess

----
:books: Documentation preview :books:: https://bowtie-json-schema--469.org.readthedocs.build/en/469/


harrel56:
Right, now the smoke tests won't pass on PR because it's testing images from the release :D I think images that didn't pass smoke tests shouldn't be published at all. But I guess it's a problem you already mentioned few times

Julian:
> I think images that didn't pass smoke tests shouldn't be published at all. 

Indeed, if I knew how to get GH actions to behave I'd have fixed it by now.

Going to merge.

PR #602, opened 2023-11-07
update harness (dev.harrel) with support for draft 2019-09

harrel56:
----
:books: Documentation preview :books:: https://bowtie-json-schema--602.org.readthedocs.build/en/602/

### DIFF
--- a/bowtie/schemas/io.json
+++ b/bowtie/schemas/io.json
@@ -77,7 +77,7 @@
                       "description": "The name of the implementation itself",
 
                       "type": "string",
-                      "pattern": "^[A-Za-z][\\w\\-.]*[A-Za-z0-9]$"
+                      "pattern": "^[A-Za-z][\\w\\-:.]*[A-Za-z0-9]$"
                     },
                     "version": {
                       "description": "The implementation version",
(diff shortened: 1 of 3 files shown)


harrel56:
Wanted to be a bit more explicit with naming my implementation but the IHOP schema is pretty strict. `dev.harrel.json-schema` is fine but it would probably make more sense to use java/gradle notation like `dev.harrel:json-schema`. Maybe you could consider allowing `:` in the implementation name :P

Julian:
Fine with that! Feel free to modify.

harrel56:
hmm, seems like my change did not take effect - should I change it somewhere else?

Julian:
Run `pre-commit run` locally and it will fix the other place (which is [here](https://github.com/bowtie-json-schema/bowtie/blob/main/tests/fauxmplementations/lintsonschema/schemas/io.json#L80), unfortunately there's some file duplication to support one of the test cases and Docker not being able to include files outside of a given root directory.

Julian:
I'm not sure honestly why pre-commit.ci didn't fix that automatically, maybe it doesn't understand that's a changed file, so there may be a small bug in the pre-commit hook which does this check, not sure.

harrel56:
Didn't I change that file? https://github.com/bowtie-json-schema/bowtie/pull/602/files#diff-ec66c7052f7e2f942fb00efe8feea3f56e09d3b190adebc4e79bf648ca5ebd00

Julian:
Oh indeed.. weird. Will look after lunch.

Julian:
Ah the issue is because the job runs with the prod version of Bowtie (and it needs to for other reasons). Not sure if there's an obvious fix for this kind of change then other than doing it in two PRs. But will ignore.

harrel56:
nice, thanks!

PR #619, opened 2023-11-17
Ruby json schemer 2.1.0

davishmcclurg:
With some cleanup and a missing registry fix.

I didn't look too deeply at the missing registry thing—maybe it's unintentional that some test cases don't have a `registry` key?

----
:books: Documentation preview :books:: https://bowtie-json-schema--619.org.readthedocs.build/en/619/


Julian:
Thanks! (And congrats on the release).

> maybe it's unintentional that some test cases don't have a registry key?

It's intentional (the schema is [here](https://github.com/bowtie-json-schema/bowtie/blob/main/bowtie/schemas/io.json#L313) and it's not in the `required` array) -- it being missing is just equivalent to "there are no extra schemas you need for this test".

Merging, will cut a release as soon as I see why the website's crapped out today.

PR #679, opened 2023-12-08
Add support for the Scala MJS implementation

skles:
----
:books: Documentation preview :books:: https://bowtie-json-schema--679.org.readthedocs.build/en/679/

### DIFF
--- a/implementations/scala-mjs-validator/Dockerfile
+++ b/implementations/scala-mjs-validator/Dockerfile
@@ -1,17 +1,11 @@
 FROM sbtscala/scala-sbt:eclipse-temurin-17.0.4_1.7.2_2.13.10 AS builder
 
-RUN git clone https://gitlab.lip6.fr/jsonschema/modernjsonschemavalidator /opt/mjs
-WORKDIR /opt/mjs
-RUN git checkout v0.1.0
-RUN sbt assembly
-
 WORKDIR /opt/harness
 COPY Harness.scala /opt/harness
 COPY build.sbt /opt/harness
 COPY project /opt/harness/project
-RUN mkdir lib && cp /opt/mjs/target/scala-*/jschemavalidator.jar /opt/harness/lib/mjs.jar
 RUN sbt assembly
 
-FROM bellsoft/liberica-openjdk-alpine:17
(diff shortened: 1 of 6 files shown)


implementations/scala-mjs-validator/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM sbtscala/scala-sbt:eclipse-temurin-17.0.4_1.7.2_2.13.10 AS builder
+
+RUN git clone https://gitlab.lip6.fr/jsonschema/modernjsonschemavalidator /opt/mjs
```
Julian, line 3: Is there a reason you're building this yourself here rather than letting sbt do it? I'll post a suggestion (in the build.sbt) just in case you haven't considered that.

implementations/scala-mjs-validator/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN sbt assembly
+
+WORKDIR /opt/harness
+COPY Harness.scala /opt/harness
+COPY build.sbt /opt/harness
+COPY project /opt/harness/project
+RUN mkdir lib && cp /opt/mjs/target/scala-*/jschemavalidator.jar /opt/harness/lib/mjs.jar
+RUN sbt assembly
+
+FROM bellsoft/liberica-openjdk-alpine:17
```
Julian, line 15:
```suggestion
FROM bellsoft/liberica-openjdk-alpine:21
```

Julian (changes requested):
Fantastic, thanks @skles this looks great!

I left some comments, mostly on (what I suspect) further simplifies the build, and makes us less likely to need to touch a ton of places when upgrading `mjs` from `0.1.0` to whatever next version in the future.

The harness code itself seems perfect I think.

implementations/scala-mjs-validator/Dockerfile
skles: I actually had not considered that. Thank you for the suggestion!

Julian:
Perfect, I think this looks great now, will merge and you should see it show up in an hour or so! Thanks again, and congrats!

albertoblue87-netizen (commented):
Ok

Issue #27, opened 2022-09-13
Support case groups


Julian:
Or generalize the notion itself of cases to groups.

Right now in the test suite we group cases by file (often by keyword).

We should reflect this structure in the report emitted by bowtie.

Backing off should perhaps take into account groups.

`schema` and `registry` (#14) are also probably in need of consideration after this is done -- right now they're per-case, but probably can be per-case-group, especially if the notions merge.

Julian:
The linked commit adds a schema for this at least.

To be considered now is that we likely should refrain from sending *harnesses* anything but "leaf" test groups (i.e. ones that match the current structure), and preserve the nesting only in reports / for the UI. That should be mostly feasible with the committed schema, just requires some modification of the `TestCase` class.

PR #708, opened 2023-12-16
Show compliance summaries across all implementations for a specific JSON Schema Vocabulary

AgniveshChaubey:
# Work in Progress

**Note:** This PR is submitted early to get some feedback and suggestions for the ongoing work.

This pull request addresses [Issue #248](https://github.com/bowtie-json-schema/bowtie/issues/248). The primary goal is to create a cross-implementation page that displays compliance summaries across all implementations for a specific JSON Schema Vocabulary.

### Implemented Features

- Implemented the function to extract the names of vocabularies.
- Completed the placement of the dropdown menu (open to suggestions for alternative placements).

image

Search Query Logic

- Logic to add the search query for vocabulary has been implemented successfully.

In order to get more clarity I'm currently looking for some feedback/suggestions on how to proceed with filtering the `reportData` based on the selected vocabulary. Any suggestion would be really appreciated. :)

----
📚 Documentation preview 📚: https://bowtie-json-schema--708.org.readthedocs.build/en/708/

### DIFF
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -45,7 +45,7 @@ export const DialectReportView = ({
     return Array.from(vocabSet);
   }, [reportData]);
 
-  console.log(vocabularies);
+  // console.log(vocabularies);
   const filteredData = useMemo(() => {
     let filteredData = reportData;
     if (params.getAll("language").length) {
@@ -63,7 +63,7 @@ export const DialectReportView = ({
     <div>
       <div className="container p-4">
         <RunInfoSection runInfo={filteredData.runInfo} />
-        <FilterSection languages={languages} />
+        <FilterSection languages={languages} vocabularies={vocabularies} />
         <SummarySection reportData={filteredData} />
         <CasesSection reportData={filteredData} />
       </div>
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -18,7 +18,7 @@ export const DialectReportView = ({
 
   const languages = useMemo(() => {
     const langs = Array.from(reportData.implementations.values()).map(
-      (impl) => impl.metadata.language
+      (impl) => impl.metadata.language,
     );
     return Array.from(new Set(langs).values());
   }, [reportData]);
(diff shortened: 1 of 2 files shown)


Julian:
Nice! I haven't taken a close look at this yet, though I think part of what you're asking depends on #27 to happen (specifically the data needed to filter on vocabularies isn't present yet in report output).

But I've actually made a small bit of progress on that this past week, I can have another look at what's needed when I look at this PR.

And welcome back!

AgniveshChaubey:
>  I've actually made a small bit of progress on that this past week ...

Any updates on your progress?

I have tried to extract the vocabulary list from the current data itself, but that's a brute-force approach. For now, consider not reviewing this PR. Ping me when you are done with the updates, and I'll make changes in this PR accordingly.

Julian:
Hey! 

> Any updates on your progress?

I made some more, but not entirely done actually unfortunately, I got distracted by a few other things that also were in the way.

> Ping me when you are done with the updates, and I'll make changes in this PR accordingly.

I will and thanks again! Hopefully I'll spend a few more days here in the next week or so and then be able to do so. Really appreciated!

Julian:
One specific thing we *do* need by the way, which I noticed as part of the refactoring, is integration tests between the UI and backend particularly for the report processing -- i.e. we need tests that call `bowtie foo` and then parse the output of that with `parseReportData` from the UI, and which complain if that integration doesn't work (#706 is partially that).

I have to try to write one -- we have now either a place to do so in the backend test suite or also in the frontend one, and it really doesn't matter where, as long as one place has at least one, because otherwise it's quite easy to break the frontend.

AgniveshChaubey:
> One specific thing we do need, by the way, which I noticed as part of the refactoring, is integration tests....

I only have a very basic understanding of testing, but this task seems like a good start to getting started with testing as well. However, I am a bit messed up with my schedule these days, so I'll give it a shot in the upcoming weeks and will let you know if I require any help.

Julian:
Slow progress here, but I just added a first test finally, it lives here now: https://github.com/bowtie-json-schema/bowtie/blob/main/frontend/src/data/parseReportData.test.ts so we can finally make changes to the report without the UI completely breaking.

More to be done there but we're on the way finally.

AgniveshChaubey:
Closing this PR as it is too outdated!

PR #753, opened 2024-01-04
Update @hyperjump/json-schema for v1.7.0

jdesrosiers:
I just released a v1.7.0 and needed to make an update to the skipped tests. I made a few other improvements while I was at it.

Change log:
* I no longer allow schemas to self-identify using a `file:` URI scheme for security reasons. I arranged to skip those tests and simplified the retrievalURI construction that no longer needs to take into account `file:` URIs.
* Remove some tests from skipped that got moved to "optional"
* Fixed an issue where I was skipping tests I didn't need to. Oops.
* Renamed the implementation. Since v1.0.0, I haven't been using "JSV" because I consolidated all the components into one package. It's no longer just a validator, it's also a bundler, annotator, etc. Ideally I'd use the npm handle `@hyperjump/json-schema`, but for some reason the allowed characters are pretty limited.
* I switched to using the `registerSchema` and `unregisterSchema` functions over the now deprecated `addSchema` function.
* Improved remote registration logic to skip schemas for irrelevant dialects rather than letting them throw and skipping with try/catch.
* Some other cleanup. Check if the test should be skipped before registering remotes.

I was able to successfully run `bowtie smoke`, but I couldn't manage to get `bowtie suite` to run, so I'm not 100% sure these changes work as expected.

----
📚 Documentation preview 📚: https://bowtie-json-schema--753.org.readthedocs.build/en/753/

### DIFF
--- a/implementations/js-hyperjump/bowtie_hyperjump.js
+++ b/implementations/js-hyperjump/bowtie_hyperjump.js
@@ -1,7 +1,7 @@
 import readline from "readline/promises";
 import os from "os";
 import process from "process";
-import { addSchema, validate } from "@hyperjump/json-schema/draft-2020-12";
+import { registerSchema, unregisterSchema, validate } from "@hyperjump/json-schema/draft-2020-12";
 import "@hyperjump/json-schema/draft-2019-09";
 import "@hyperjump/json-schema/draft-07";
 import "@hyperjump/json-schema/draft-06";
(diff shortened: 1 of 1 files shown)


Julian:
Nice, thanks!

> I was able to successfully run bowtie smoke, but I couldn't manage to get bowtie suite to run, so I'm not 100% sure these changes work as expected.

This is thankfully trivial with GitPod. You can see a [new workspace with your branch](https://gitpod.io/new#https://github.com/jdesrosiers/bowtie/tree/hyperjump17) shows (after building your image, which probably we should do automatically for PRs):

```
gitpod /workspace/bowtie (hyperjump17) $ bowtie suite -i js-hyperjump 2020-12 | bowtie summary --show failures
2024-01-04 19:39.20 [info     ] Finished                       count=359
                              Bowtie                              
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━━━┓
┃ Implementation                     ┃ Skips ┃ Errors ┃ Failures ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━╇━━━━━━━━━━┩
│ hyperjump-json-schema (javascript) │ 4     │ 0      │ 0        │
└────────────────────────────────────┴───────┴────────┴──────────┘
```

Julian:
>  but for some reason the allowed characters are pretty limited

Names are used for files on the filesystem, so they have to be pretty reasonable as filesystem characters essentially.

PR #769, opened 2024-01-12
Update unsupported tests of MJS

skles:
Two tests containing Unicode code points were renamed in the draft 2020-12 Test Suite. MJS does currently not support Unicode. This adds the new names to the set of unsupported tests.

For backwards-compatibility with older version of the draft 2020-12 Test Suite, we keep the previous names of the tests.

----
📚 Documentation preview 📚: https://bowtie-json-schema--769.org.readthedocs.build/en/769/


Julian:
Thanks!

Issue #543, opened 2023-10-11
Highlight the selected test case


gregsdennis:
When a test case has been clicked, the associated instance is populated into the top grid.  However, there should be some visual indication as to which test case is active.

![image](https://github.com/bowtie-json-schema/bowtie/assets/2676804/fc883d63-3207-44ee-a1b7-f69e4f4b9820)

In the screenshot, it's impossible to tell which case is shown in the "instance" column.

Julian:
(Definitely seems like a good idea to me!)

sudo-jarvis:
Hi @gregsdennis, @Julian 
I have implemented the enhancement and have created a PR for the same, kindly check.

PR #827, opened 2024-02-05
Highlight the selected test case

sudo-jarvis:
**This pull request addresses #543.**

The goal is to highlight the test case which has been selected as currently it's impossible to tell which case is shown in the "instance" column.

**Implemented Enhancement:** 

Highlighted the selected test case in a greyish gradient so that it is distinguishable from the other test cases.

Screenshot 2024-02-05 at 8 12 14 PM

----
📚 Documentation preview 📚: https://bowtie-json-schema--827.org.readthedocs.build/en/827/


Julian:
Looks great, well done, appreciated!

PR #832, opened 2024-02-07
Update skipped tests for MJS validator

skles:
Some tests were skipped as a leftover artifact due to restrictions in an older version that are already resolved in the current MJS version (v0.1.0).

----
📚 Documentation preview 📚: https://bowtie-json-schema--832.org.readthedocs.build/en/832/


Julian:
Thanks!

Issue #831, opened 2024-02-06
Add `--format markdown` to commands


Julian:
It's useful to have a way to generate markdown-formatted output from e.g. `bowtie summary` for interoperability with other human-reader-targeted tooling.

If we had that, we could pipe that markdown to e.g. [GitHub Actions step summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) so that we see summary output when regenerating reports for the website.

We should add this for any command with `--format json` and `--format pretty` options. The desired output should match `--format pretty` as closely as possible, but with markdown-compatible syntax, leaving rendering up to the markdown engine.

We should also use an existing library to generate the markdown (I forget if rich can do this or if it can only *render* markdown, probably the latter, so we should pick an existing library capable of emitting markdown).

sudo-jarvis:
@Julian , I was looking onto this, markdown doesnt seem to support showing nested tables as we use in `bowtie summary`, so some other way needs to be figured out to show the same data, it wouldnt be exactly like `--format pretty`

Julian:
Ah -- indeed, that's a good point.

Maybe a good first step then is to only do this for `--show failures`, where there's no nested table.

And then we'd think of what to do for cases where we have to flatten the table for markdown...

sudo-jarvis:
Hi @Julian 
I have implemented this and have created a PR for the same, kindly check.

PR #839, opened 2024-02-08
add --format markdown to commands

sudo-jarvis:
**This pull request addresses #831.**

The goal is to add markdown formatter to commands for interoperability with other human-reader-targeted tooling.

### Implemented Enhancement:

Added --format option: markdown to be able to give output in markdown format for commands such as bowtie summary, bowtie info, bowtie smoke, etc.

### Output images:

**1. bowtie summary with show failures**
Screenshot 2024-02-08 at 10 54 09 AM

**2. bowtie summary with results**
Screenshot 2024-02-08 at 10 49 04 AM
**3. bowtie info**
Screenshot 2024-02-08 at 10 53 10 AM

**4. bowtie smoke** 
Screenshot 2024-02-08 at 10 52 35 AM

----
📚 Documentation preview 📚: https://bowtie-json-schema--839.org.readthedocs.build/en/839/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -21,7 +21,6 @@ from rich.table import Column, Table
 from rich.text import Text
 from trogon import tui  # type: ignore[reportMissingTypeStubs]
 import click
-import markdown
 import referencing_loaders
 import rich
 import structlog
@@ -450,9 +449,9 @@ def _failure_table_in_markdown(
 
     markdown_table = _convert_table_to_markdown(columns, rows)
     return (
-        markdown.markdown("# Bowtie Failures Summary")
+        "# Bowtie Failures Summary"
         + markdown_table
-        + markdown.markdown(f"{report.total_tests} {test} ran\n")
+        + f"**{report.total_tests} {test} ran**\n"
     )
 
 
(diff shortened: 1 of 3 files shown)


Julian:
Nice work. I haven't looked closely yet at using this but two initial comments:

* Have a try at adding tests -- you should see two or three that you can reference, which test the other format modes
* I don't think we need to depend on pandas just for this functionality, but I'll have a look at what you're using it for and try to suggest an alternative perhaps
Will leave some more comments today or tomorrow after looking more carefully.

sudo-jarvis:
It was convenient to convert tables to markdown using pandas so I used it, ig the same can be done using tabulate too.

Julian:
`tabulate` seems like a reasonable option from a quick look -- basically I'd like to not depend on 3 separate packages just for this functionality, I suspect it should be reasonably straightforward I'd hope, but if you need help figuring out which one I can try to look more closely.

Julian:
Thanks, can you give a shot to the tests? That makes it easier to see this works at a glance without needing to run it yet

sudo-jarvis:
Added the tests @Julian

Julian:
Cool! The tests look like you're emitting HTML though, not markdown, no?

sudo-jarvis:
My bad, just figured out markdown.markdown function was rendering html and not markdown, not sure why isnt it named markdown.html then. Since markdown supports html too it worked fine.

Fixed it now.

Julian:
Great! This looks very nice! Going to merge -- if we do find a library which can generate the table without doing it manually we can always remove some code.

Thanks again!

P.S. if you care for a follow-up, a good one would now be making the `bowtie smoke` output show up in GitHub actions by emitting it to `GITHUB_STEP_OUTPUT` when we run the `images.yml` workflow, and similarly the `bowtie summary` output when we run the `report.yml` workflow!

sudo-jarvis:
Okay @Julian, I'll look into it and start working on it once I get some idea.

PR #849, opened 2024-02-10
Fix rerouting back to Local Report after upload

### LINKED ISSUE #331: Uploading a second local report is hard
It isn't easy to upload a second local report after doing so once.

To reproduce:

* Click on "Local report" in the dialect menu
* Upload a local report
* Click on "Local report" again
* The page does not change / the upload dialog does not come up

Refreshing/using the back button works of course, but this shouldn't be necessary.

adwait-godbole:
Fixes #331 

----
📚 Documentation preview 📚: https://bowtie-json-schema--849.org.readthedocs.build/en/849/


Julian:
Nice! Thanks again!

Issue #448, opened 2023-09-06
Highlight new implementations in the UI


Julian:
When we add a new implementation to the UI we likely should somehow highlight it for a period of time, both to showcase a new line in the report as well as to give a bit of "props" to the implementation.

This could be by highlighting it in the summary table somehow, via an icon, via a small card in the side, etc.

sudo-jarvis:
Hi @Julian , I would like to work on this. 

I dont think we already have the date when some implementation has been added being stored somewhere. We can add that along with other metadata that we already have for implementations. Then in the frontend we could simply highlight the implementation if say it is not older than a month or so.

Julian:
I think a good way to handle this one without manual bookkeeping -- and in a way that will benefit other potential use cases -- is to always save *yesterday*'s report when rebuilding the site.

This will let us show day-over-day deltas (changes to test results), but it should also basically give you what you need for this -- you can highlight any implementation which appears today and not yesterday essentially.

Doing that *should* just be a download + copy/rename in the workflow.

Does that make sense to you?

sudo-jarvis:
I have done the frontend changes for the same, but am not sure in which workflow can I access the older reports and upload them to the artifact in some folder named say `prev-reports`. Could you please help me @Julian

Julian:
It'd be in the [report workflow](https://github.com/bowtie-json-schema/bowtie/blob/main/.github/workflows/report.yml) -- before replacing the file, download the current one and include it alongside the new one -- let's call the directory `previous`, so the URLs should be https://bowtie.report/draft2020-12.json for today's / the current report, and https://bowtie.report/previous/draft2020-12.json for the previous one.

PR #851, opened 2024-02-11
Save older build reports separately to track changes

sudo-jarvis:
This pull request addresses #448.

Incorporated the changes in workflow to store the previous reports in a separate folder

----
📚 Documentation preview 📚: https://bowtie-json-schema--851.org.readthedocs.build/en/851/

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -26,7 +26,7 @@ jobs:
         uses: dawidd6/action-download-artifact@v3
         with:
           workflow: report.yml
-          branch: test-workflow
+          branch: main
           name: report
           path: report/previous
 
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -19,7 +19,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  store-prev-reports:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download previous test report from local artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: report
+          path: previous
+
+      - name: Upload previous test report to previous folder in artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: report
+          path: |
+            previous/
+
   test:
+    needs: store-prev-reports
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


sudo-jarvis:
Hi @Julian , can you please look at the changes ? Also, is there any way of testing what the artifact contains ?

Julian:
If you enable workflows on your fork of this repo so that it builds your own version of the site I *think* you should be able to test this out!

sudo-jarvis:
Hi @Julian, I have tested the changes. If you could merge the PR and run the `Regenerate Test Reports` workflow then the older reports would also be available. Then the frontend PR can be merged

.github/workflows/report.yml
```diff
@@ -19,7 +19,26 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  store-prev-reports:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download previous latest test report
```
Julian, line 25: Let's do this in two parts -- do the download here, but do the upload after the next job runs, otherwise we may overwrite the last report but then today's may fail.

sudo-jarvis: Yeah, I agree its a good point. Made the changes now.

.github/workflows/report.yml
```diff
@@ -19,7 +19,29 @@ concurrency:
...
 
 jobs:
+  download-prev-reports:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download previous test report from local artifact
+        uses: dawidd6/action-download-artifact@v3
+        with:
+          workflow: report.yml
+          branch: test-workflow
```
Julian, line 29: This needs removing now.

sudo-jarvis: Ah my bad, was using it for testing, forgot to change

.github/workflows/report.yml
```diff
@@ -19,7 +19,29 @@ concurrency:
...
 
 jobs:
+  download-prev-reports:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download previous test report from local artifact
+        uses: dawidd6/action-download-artifact@v3
+        with:
+          workflow: report.yml
+          branch: main
```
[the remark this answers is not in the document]
sudo-jarvis: Oh yeah you are right, I didnt think this way. I didnt think about the testing usecase. Will remove it , so that it uses the default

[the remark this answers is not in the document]
sudo-jarvis: @Julian, removed the option now. Sorry missed it earlier

PR #852, opened 2024-02-11
Highlight new implementations in summary

### LINKED ISSUE #448: Highlight new implementations in the UI

sudo-jarvis:
Closes #448.

Implemented a way to fetch older reports as well and check which implementations are new by comparing both the reports and highlight such implementations in the UI in summary table.

**Implementation :** 
Showing a small + sign in front of implementations that are new i.e. not present in the previous build report and on hovering over the + sign, the user sees a `Newly Added Implementation Tooltip` 

**Sample :**

https://github.com/bowtie-json-schema/bowtie/assets/75855708/7575674b-383f-4dfc-8db1-e5ac152ee318

----
📚 Documentation preview 📚: https://bowtie-json-schema--852.org.readthedocs.build/en/852/


sudo-jarvis:
@Julian , does the change in the UI look good ?

Julian:
Cool! Let's maybe use some bright color for "new implementation" rather than the same color as what we use for "active row" -- can you perhaps check to see whether we have some light pink or green maybe in the CSS theming we're using, I suspect so since we're using Bootstrap.

Julian:
There's also a small problem I suppose with my suggestion which I'm just thinking of -- essentially this will only show an implementation as new for 24 hours, when ideally we'd show it for some longer period, maybe a week or something.

But I don't have a good idea for how to do that yet, and I think the work you've done is anyways useful for day-over-day information, so maybe the right thing for now is to keep going and we can change/improve this when we have another idea.

sudo-jarvis:
@Julian , does this look good ?

Screenshot 2024-02-12 at 5 49 51 PM

frontend/src/components/Summary/ImplementationRow.tsx
```diff
@@ -56,6 +69,19 @@ const ImplementationRow = ({
         >
           Details
         </button>
+        &nbsp;&nbsp;
```
Julian, line 72: Are the spaces intentional?

frontend/src/components/Summary/ImplementationRow.tsx
```diff
@@ -56,6 +69,19 @@ const ImplementationRow = ({
...
+        {implementation.isNew && (
+          <OverlayTrigger
+            placement="right"
+            overlay={
+              <Tooltip id="tooltip">
+                <strong>Newly added Implementation</strong>
```
Julian, line 78:
```suggestion
                <strong>Newly added implementation</strong>
```

frontend/src/data/parseReportData.ts
```diff
@@ -1,3 +1,15 @@
+export const parseReportWithDiff = (
```
Julian, line 1: It will require a small bit of additional refactoring, but you shouldn't need to download the whole previous report to do this -- all you need is the first line of it! So you should be able to just retrieve the first line to get only the implentations from yesterday. On the backend this corresponds to the `RunMetadata` class, so we may need a version of that in the UI.

Julian (commented):
I haven't looked at this fully yet, so here's just a first pass with one comment that I expect may take a little refactoring.

frontend/src/data/parseReportData.ts
sudo-jarvis: @Julian But in the case of frontend, the report would be present remotely and I am not sure if we can just fetch the first line for this file, the whole file would have to be downloaded first then we can discard the contents except the first line, no ?

Julian: No certainly not :) -- HTTP is built precisely for these kinds of things. It supports even streaming, so you open requests to somewhere that may produce infinite data and just read some of it. Or if it's helpful conceptually, the mental model you probably should have is that when you transfer some data from one place to the other that's like someone who has a copy of it reading you all of its contents over the phone from beginning to end -- you can of course now not get all of them, just hang up the phone after you hear what you want.

Now which Javascript APIs support this that I don't know offhand :), so would require some looking. From a first google, it seems possible that the API might be the same one we're using (`fetch()`) but instead of awaiting it which will wait for the whole body, maybe [`response.body`](https://developer.mozilla.org/en-US/docs/Web/API/Request/body) which returns a stream (that then you can read from until you see a newline). There may though be a middle-level API in between the two which gives you something that already does a bit more for you, but I'm not sure -- I'd try to use `response.body` and see if it looks like you end up having to write more than 3-4 lines of code.

sudo-jarvis: I was able to implement a stream reader which can read the stream until it encounters a newline in the text.

sudo-jarvis:
@Julian , I have used http streaming now using fetch itself to just fetch the metadata of the report for finding about new implementations. Please have a look

sudo-jarvis:
@Julian ig you missed this

Julian:
Yes, thanks for the ping, looking again.

frontend/src/data/Dialect.ts
```diff
@@ -32,14 +31,70 @@ export default class Dialect {
...
-    const lines = jsonl
+    const curReportLines = jsonl
       .trim()
       .split(/\r?\n/)
       .map((line) => JSON.parse(line) as Record<string, unknown>);
-    return parseReportData(lines);
+
+    let curReport = parseReportData(curReportLines);
+
+    const prevVersionUrl = baseURI
```
Julian, line 48: It strikes me as odd that a function meant for retrieving a report is also retrieving the previous report. That seems like something that should be happening outside of here.

You seem to be doing it because of the new added `isNew` property -- but maybe that's indeed worth *not* putting on `Implementation`, it doesn't seem like it does much except create the possibility to be wrong.

Why not just have this function be more or less as it was, then have a *separate* one which pulls the `RunInfo` from yesterday and gives you a `Set` of yesterday's implementations? Then your `isNew` is just checking for set membership.

frontend/src/data/Dialect.ts
```diff
@@ -32,14 +31,70 @@ export default class Dialect {
...
+      .href();
+    const prevReportLines = await this.fetchReportMetadata(prevVersionUrl);
+    if (prevReportLines) {
+      const prevReportMetaData = JSON.parse(prevReportLines) as RunInfo;
+      curReport = this.compareReportToOld(curReport, prevReportMetaData);
+    }
+    return curReport;
+  }
+
+  async fetchReportMetadata(url: string) {
```
Julian, line 62: This looks like what I'd expect, but also looks like the sort of thing I would expect to already *exist* somewhere -- I believe you if you looked and didn't see it, but let's still put it somewhere where we can replace it when/if we find something which does it.

So rather than doing it inline in a function called `fetchReportMetadata`, do it in some helper function which for any `fetch()` call streams the lines of the response (as a JS coroutine I think?). That likely exists somewhere, so at some point we can replace it when we find it, and until then you can use it to do this here.

Julian:
It also occurs to me that with @adwait-godbole's change to add an `implementations.json` file, containing the output of Bowtie info, that we have a second way of doing this by instead looking at that file instead of at the report -- but we can also continue down this way and always decide to change later, especially considering that PR isn't merged yet and your previous one which stores yesterday's report already is.

frontend/src/data/Dialect.ts
sudo-jarvis: So if I am understanding it correctly you are saying that the isNew Field is not needed and instead we could pass that data as a prop maybe ?

Julian: Yeah basically you can calculate it when it's needed from the combination of one implementation and a list of yesterdays -- so if you pass that list of yesterday's implementations to the page which wants to highlight,.you can highlight all those implementations not appearing in the yesterday list.

Does that make sense?

sudo-jarvis: Yeahh, got it

Julian:
> I just checked ig the regenerate test reports workflow hasnt run succesfully after merging the previous report changes. 

I'm having a look to address the disk space issue which has been annoying, I'll get you a working run.

sudo-jarvis:
@Julian, I have taken into consideration the review comments and have made the changes. Please review

Julian:
It looks good from the first glance but I'm playing still with the disk space issue.

frontend/src/data/Dialect.ts
```diff
@@ -37,6 +36,52 @@ export default class Dialect {
...
+      });
+    if (prevReportLines) {
+      const prevReportMetaData = JSON.parse(prevReportLines) as RunInfo;
+      return prevReportMetaData.implementations;
+    }
+    return null;
+  }
+
+  // Helper function to stream and fetch the first `num_lines`
+  async fetchResponseStream(response: Response, num_lines: number) {
```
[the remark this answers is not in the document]
sudo-jarvis: I had done the same however @Julian suggested that it would be better to stream the response and fetch the first line only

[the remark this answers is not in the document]
Julian: > Does it really helps so much performance-wise? Couldn't you just read it all to string and then get the 1st line?

Mostly I was surprised this isn't trivial (which I didn't know at the time when suggesting it) -- in Python this is a 2 second thing to do -- in retrospect I wouldn't have suggested it certainly, knowing that it appears it's not in JS, but it seemed unnecessary to tell you to *back it out* after you figured it out.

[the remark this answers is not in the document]
sudo-jarvis: I can roll it back to just fetching the first line from the whole response. I also think that the code is getting complicated as compared to the small time advantage that it would provide.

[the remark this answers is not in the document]
harrel56: Ideally the server could just return exactly what is needed but It's probably even more troublesome with gh static pages :D

Julian:
(There's a working run with previous implementations FWIW now)

sudo-jarvis:
@harrel56 Yeah currently it would vanish after 24 hrs when the workflow runs again. Ig to ensure that it stays for some time we would need the date when a particular implementation was added stored somewhere.

harrel56:
Hmm, actually having previous report from one month ago in a current setup is a bit hard to achieve. We use https://github.com/dawidd6/action-download-artifact action for downloading artifacts and unfortunately it doesn't have option to somehow filter by creation date. But it uses GH API (https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-workflow-run-artifacts) and here each artifact has its creation date, so it's totally doable. But i don't know if it's worth to implement on our own

Julian:
I think doing fancier things there will be easier if we control more of the site generation from inside Bowtie rather than in the workflow -- specifically getting us to having a `bowtie site` CLI command (essentially specifically tailored to "generate the data the UI wants") where we can evolve that as needed, including by generating more of these kinds of "materialized views". There's a very first step there now [here](https://github.com/bowtie-json-schema/bowtie/blob/badges/.github/workflows/report.yml) (on a poorly named `badges` branch because it started as a way to fix #909) only inasmuch as the sequence of generated data makes slightly more sense, and there's now a `site/` directory instead of splatting everything into the repo checkout.

Obviously I agree a day is too short, and a month sounds good there -- trying to also unblock our real current issue (the disk size nonsense) which is more pressing, but I'm certainly open to any kind of idea for this -- whether we delay a bit, merge this as is and improve after, or yeah anything.

Julian:
Sorry this petered out, let's come back to it properly when we have a solution that won't get overwritten the next day (as I say probably once/if we have a `bowtie site` command).

PR #855, opened 2024-02-11
Make `bowtie info` include GitHub-retrieved repository information (stars, watchers, etc.)

### LINKED ISSUE #854: Extend bowtie info to fetch latest stats of implementation repos
update `bowtie info` command to fetch the latest stats of implementation (harness)'s repos along with the option of `--no-remote-data` to disable the fetching so that `bowtie info` is still runnable without an internet connection. First steps towards fixing #576

adwait-godbole:
fixes #854 

hey @Julian these are some of the queries / help needed I have with this PR:

1. How can I fix the Build UI workflow fails. It is failing on running the `bowtie info $(ls implementations/ | sed 's/^/-i /') --format json | tee implementations.json` command saying some images failed to start.

2. I am not sure if freeing up the disk space was necessary but still added it seeing the comment that it had generated errors during report generation workflow.

3. GitHub API rate limits sometimes with the following error message: `API rate limit exceeded for xxx.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) `

4. Fetching GitHub API data works for most of the repos but sometimes returns 404 Not Found for some repos even though the org name and repo name was correct. For e.g. for this repo it returns 404 --> https://github.com/microsoft/vscode-json-languageservice.

Sorry for asking so many questions 😅. Thanks for your help so far!

### DIFF
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -19,7 +19,7 @@ jobs:
 
       - name: Generate all implementations (harness)'s metadata
         run: |
-          bowtie info $(ls implementations/ | sed 's/^/-i /') --format file  | tee implementations.json
+          bowtie info $(ls implementations/ | sed 's/^/-i /') --format file > implementations.json
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           node-version: 18.x
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v3
         with:
           version: 8
           run_install: false
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -8,18 +8,10 @@ on:
       - "frontend/**"
       - ".github/workflows/build-frontend.yml"
 
-permissions:
-  contents: read
-
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      # Freed up the disk space just in case if it gives error later
-      - name: Free up disk space
-        run: |
-          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/.ghcup /usr/local/lib/android /usr/local/share/boost "$AGENT_TOOLSDIRECTORY"
-
       - uses: actions/checkout@v4
 
       - name: Install Bowtie
@@ -27,7 +19,7 @@ jobs:
 
       - name: Generate all implementations (harness)'s metadata
         run: |
-          `bowtie info $(ls implementations/ | sed 's/^/-i /') --format json  | tee implementations.json`
+          bowtie info $(ls implementations/ | sed 's/^/-i /') --format jsonFile  | tee implementations.json
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Generate all implementations (harness)'s metadata
         run: |
-          `bowtie info $(ls implementations/ | sed 's/^/-i /')  | tee implementationsMetadata.json`
+          `bowtie info $(ls implementations/ | sed 's/^/-i /') --format json  | tee implementations.json`
 
       - uses: actions/setup-node@v4
         with:
@@ -62,4 +62,4 @@ jobs:
           name: frontend-build
           path: |
             frontend/build
-            implementationsMetadata.json
+            implementations.json
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -19,7 +19,7 @@ jobs:
 
       - name: Generate all implementations (harness)'s metadata
         run: |
-          bowtie info $(ls implementations/ | sed 's/^/-i /') --format jsonFile  | tee implementations.json
+          bowtie info $(ls implementations/ | sed 's/^/-i /') --format file  | tee implementations.json
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -8,11 +8,26 @@ on:
       - "frontend/**"
       - ".github/workflows/build-frontend.yml"
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      # Freed up the disk space just in case if it gives error later
+      - name: Free up disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/.ghcup /usr/local/lib/android /usr/local/share/boost "$AGENT_TOOLSDIRECTORY"
+
       - uses: actions/checkout@v4
+
+      - name: Install Bowtie
+        uses: ./
+
+      - name: Generate all implementations (harness)'s metadata
+        run: |
+          `bowtie info $(ls implementations/ | sed 's/^/-i /')  | tee implementationsMetadata.json`
 
       - uses: actions/setup-node@v4
         with:
@@ -45,4 +60,6 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           name: frontend-build
-          path: frontend/build
+          path: |
+            frontend/build
+            implementationsMetadata.json
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -731,7 +731,7 @@ def validate(
     "--no-remote-data",
     is_flag=True,
     default=False,
-    help="Fetch implementation repo stats from GitHub"
+    help="Disable retrieving implementation (harness)'s metadata from GitHub"
 )
 async def info(implementations: Iterable[Implementation], format: _F, no_remote_data: bool):
     """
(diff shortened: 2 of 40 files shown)


adwait-godbole:
hey @Julian I am not sure if freeing up the disk space was necessary but still added it seeing the comment that it had generated errors during report generation workflow

adwait-godbole:
Hey @Julian do you have any suggestions?

Julian:
Hi there!

> How can I fix the Build UI workflow fails.

You have an extra set of backticks around the command in the workflow. Remove them and then it should run correctly.

> I am not sure if freeing up the disk space was necessary but still added it seeing the comment that it had generated errors during report generation workflow.

It shouldn't be (nor should the `permissions` block I think).

> GitHub API rate limits sometimes with the following error message: API rate limit exceeded for xxx.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) 

Locally or in GitHub actions? I think in GHA I thought I had fixed that (to use the automatically provisioned token for any GHA run) but possibly not -- whether it happens or not is somewhat random, based on whether any other jobs that were assigned that IP had accessed the API.

> Fetching GitHub API data works for most of the repos but sometimes returns 404 Not Found for some repos even though the org name and repo name was correct. For e.g. for this repo it returns 404

I'll have a look, and leave some other comments in a second.

bowtie/_cli.py
```diff
@@ -727,12 +727,66 @@ def validate(
...
+                    NotFoundError,
+                )
+
+                gh = GitHub(token=os.environ.get("GITHUB_TOKEN", ""))
+                pattern = r"https://github\.com/([^/]+)/([^/]+)"
+                metadata_dict = dict(metadata)
+                if "source" in metadata_dict:
+                    match = re.search(pattern, metadata_dict.get("source"))  # type: ignore[reportUnknownMemberType,reportUnknownArgumentType]
+                    if match:
+                        org = match.group(1)  # type: ignore[reportUnknownMemberType,reportUnknownVariableType]
```
Julian, line 762: This looks too long to do inline here in this function where it's taking up a significant portion of space even though it doesn't really have to do with the core of what this command does. Try to think about splitting it into a helper function.

tests/test_integration.py
```diff
@@ -1013,12 +1017,85 @@ async def test_info_links(links):
...
+    )
+
+    open_issues_count = 0
+    open_prs_count = 0
+    stars_count = 0
+    watchers_count = 0
+    last_commit_date = ""
+    last_release_date = ""
+
+    try:
```
Julian, line 1038: This has a few issues --

It essentially reproduces the code being tested, which is always a bad sign. The point of a test is to ensure some other code is correct. If the test reproduces that code, that's 2 places it could be equally wrong, and that just means when someone figures out it's wrong, they have to change 2 places instead of 1. Somehow a test has to be simpler to see it is correct than the code it tests.

It's certainly good and right to want to test the code you wrote. The real way to do so is generally to be able to inject in a fake HTTP client to the `GitHub` object you're using, so that you can make it *think* it spoke to GitHub but in fact have it simply be given a response from GitHub that you have already retrieved ahead of time. How easy that is I don't recall, but I can have a look and try to give a suggestion in the next few days.

bowtie/_cli.py
adwait-godbole: I think this will get resolved once I figure out how to make no_remote_data as a callable.

tests/test_integration.py
Julian: > How is it possible to know these values without doing a real API call ?

It's not possible to know the *real* values, but it *is* possible to simply have dummy values, but still in the right "shape".

We don't care whether the watcher count is 3427 to write a test -- we just care that it's a number inside a property called `"watchers_count"` or whatever, and same for all the properties. So if you retrieve one "real response" and simply store it you now have a fake response you can use. I again can spend some time trying to give you an even more specific suggestion if that doesn't help yet, but probably not till the weekend.

adwait-godbole: Ohh okayy. I got your point. Thanks!. I will just retrieve one real response and make the changes.

adwait-godbole:
Hey @Julian implemented all of your suggestions. Please have a look. Also extremely sorry for triggering so many workflows I had no idea that these workflows could be run on my fork as well. Won't happen in the next PRs.

Julian:
That's no problem at all! Feel free to do whatever you find helpful for your development there. And will have a look.

adwait-godbole:
Forgot to put --file instead of --jsonFile. Will change.

Julian:
I don't follow why you're adding `--format file`, does it have something to do with parsing difficulties reading the output?

If so the fix should be only use `indent=2` if we only have one implementation and/or if `sys.stdout.isatty()` (probably both of those), otherwise don't use indent. And then you can just use `--format json`.

adwait-godbole:
so just these two lines would work ig ?

`case "json":
                if sys.stdout.isatty():
                    click.echo(json.dumps(dict(metadata), indent=2))
                else:
                    click.echo(json.dumps(dict(metadata)))`

Julian:
With `implementations` to check how many implementations we're outputting so from memory something like:

```
case "json":
    kwargs = {"indent": 2} if sys.stdout.isatty() and len(implementations) == 1 else {}
    click.echo(json.dumps(dict(metadata), **kwargs))
...
```

That will remove indentation if we are printing 2 or more implementations or if we're not printing to a terminal.

adwait-godbole:
Okay! I will make these changes. Any other things to change ?

bowtie/_cli.py
```diff
@@ -571,6 +571,53 @@ def do_not_validate(*ignored: SchemaResource) -> Callable[..., None]:
     return lambda *args, **kwargs: None
 
 
+def make_remote_data_getter():
+
+    def get_remote_data(source_url: str) -> list[tuple[str, str | int]]:
```
Julian, line 576: The extra level of callable here should be unnecessary -- you don't need both `make_remote_data_getter` and `get_remote_data`, just `get_remote_data` is fine (same for `no_remote_data` and the `lambda`, you just need one level of callable.

bowtie/_cli.py
```diff
@@ -755,14 +809,45 @@ def validate(
...
+    "format",
+    help="What format to use for the output",
+    default=lambda: "pretty" if sys.stdout.isatty() else "json",
+    show_default="pretty if stdout is a tty, otherwise JSON",
+    type=click.Choice(["json", "file", "pretty", "markdown"]),
+)
+@NOREMOTEDATA
+async def info(
+    implementations: Iterable[Implementation],
+    format: Literal["json", "file", "pretty", "markdown"],
```
Julian, line 831: (You still have `file` here)

bowtie/_cli.py
```diff
@@ -755,14 +809,45 @@ def validate(
...
+class _MakeRemoteDataGetter(Protocol):
+    def __call__(
+        self,
+    ) -> Callable[[str], None | list[tuple[str, str | int]]]: ...
+
+
 @implementation_subcommand()  # type: ignore[reportArgumentType]
-@FORMAT
-async def info(implementations: Iterable[Implementation], format: _F):
+@click.option(
```
Julian, line 819: And here -- just use `@FORMAT`, not redefine it.

tests/test_integration.py
```diff
@@ -1055,12 +1059,59 @@ async def test_info_links(links):
     assert stderr == ""
 
 
+@pytest.mark.asyncio
+async def test_info_remote_data(envsonschema):
+    jsonout, stderr = await bowtie(
```
Julian, line 1064: This isn't what I had in mind, as it still will hit the internet -- you can see that yourself by trying to run this test with your computer not connected to the internet -- as I say, in order to not do that, you need to be able to inject in the fake response below *without* actually making a call to the internet. I'll try to look more closely and show you how to do so this weekend.

bowtie/_cli.py
adwait-godbole: Yeah sorry I was waiting for your response to make other changes along with this.

.github/workflows/build-frontend.yml
```diff
@@ -14,6 +14,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install Bowtie
+        uses: ./
+
+      - name: Generate all implementations (harness)'s metadata
+        run: |
+          bowtie info $(ls implementations/ | sed 's/^/-i /') --format file  | tee implementations.json
```
Julian, line 22: `tee` is used for writing to multiple places, as in the example you copied from where we also want the output to go to `bowtie badges`. Here you have only one place, and can just use `>` to write directly to the file you want it to go to.

adwait-godbole:
Hey @Julian I made the other changes you mentioned and also made an attempt at mocking the GitHub client. Didn't work though :(
I went through some articles which mentioned using pytest-mock that allows making fixtures. So I tried patching the `bowtie._cli.GitHub` import with `Mock(spec=github3.GitHub)` . I thought when the tests run they will use this patched GitHub client in _cli but it didn't work.

Julian:
Great!

> Is there some other way around of injecting this mocked_github_client in the def info function call ?

You'll want to have a read through https://github3.readthedocs.io/en/latest/contributing/testing.html#integration-tests and/or the other pages in the docs for testing, which is what I was mentioning earlier but haven't yet looked at myself.

Julian:
(I don't want to be using a mock, we should be using a fake object, but that may not make a ton of sense to you as the terminology is very nuanced. But those pages are the first place to look!)

adwait-godbole:
I will have a look.

adwait-godbole:
@Julian Apologies for reaching out again. Despite my efforts to understand these docs, I am having a hard time figuring out how to inject the fake client into the info call. 😅 I understand if your time is limited, but I'd be extremely grateful for some additional guidance on this matter.

Julian:
I don't have time to look at this yet but you can split this PR in two to unblock yourself. If you first do the bit which simply creates the `implementations.json` file you can finish #877, and then that will magically also work for this once the output also contains the remote GitHub information.

adwait-godbole:
Hey @Julian . I see a "Merged" tag in here as well but this PR shouldn't have been merged until the injecting of GitHub client for testing purposes was fixed (which unfortunately I wasn't able to figure out)

Julian:
It looks like there was a strange merge commit in there, which is what is confusing git (and making GitHub think this is merged), but I didn't look closely at what happened, I'm on a call.
The changes from this are not merged though (even if somehow the commits were) -- so you should be able to work on top of what's merged.
I can have a closer look in a bit if need be.

adwait-godbole:
Yes, indeed the changes of this PR aren't merged.

PR #865, opened 2024-02-13
Don't merge until PR #855 !!  Render Implementation Activity on UI

adwait-godbole:
Hey @Julian do these changes look good ?

Screenshot (39)

**Disclaimer:** Don't merge this PR right away after PR #855 because I will have to change the --format json option to --format jsonFile so that there are no indentations in the output json file otherwise frontend throws parsing error meaning I will change:
`json.dumps(dict(metadata), indent=2)` to `json.dumps(dict(metadata))` in the --format jsonFile option so no indentations in the output file and frontend will be able to parse it.

### DIFF
--- a/frontend/src/utils/formatDate.ts
+++ b/frontend/src/utils/formatDate.ts
@@ -1,0 +1,10 @@
+export const formatDate = (date: string) => {
+  const dateObj = new Date(date);
+  const options: Intl.DateTimeFormatOptions = {
+    year: "numeric",
+    month: "long",
+    day: "numeric",
+  };
+  const formattedDate = dateObj.toLocaleDateString("en-US", options);
+  return formattedDate;
+};
--- a/frontend/src/components/DragAndDrop/DragAndDrop.tsx
+++ b/frontend/src/components/DragAndDrop/DragAndDrop.tsx
@@ -17,7 +17,7 @@ export const DragAndDrop = () => {
     setLines(null);
     setDragActive(false);
     setInvalidFile(false);
-  }, [location]);
+  }, [location.key]);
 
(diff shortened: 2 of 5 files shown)


frontend/src/utils/formatDate.ts
```diff
@@ -0,0 +1,10 @@
+export const formatDate = (date: string) => {
```
Julian, line 1: I haven't looked at any of the rest of this PR and may not do so until we get the other one hashed out, but just because I see this now too -- we shouldn't make localization choices for users, the whole point of localization is that users should see the right thing for where they're from :) -- no one should have to see US-formatted dates if they don't generally use US-formatted dates, so we should be using the locale-aware ways to render dates.

Julian: Something like

```
const options = {
  weekday: 'long',
  year: 'numeric',
  month: 'long',
  day: 'numeric',
};
yourDate.toLocaleDateString(undefined, options);
```

according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString.

adwait-godbole: Fixed!

harrel56:
I would suggest using original GH icons, IMO this would look most natural. But this would require using different icons library (probably this https://github.com/react-icons/react-icons, as the bootstrap ones are packaged as well). So this might be bigger change - if you decide to go for it, I think this should be a separate PR

adwait-godbole:
Definitely I can look into it in future. Actually the title of this PR says that don't merge until PR #855. That PR is still not merged actually! for some odd reason it has been tagged as Merged from GitHub although the code never made it to the repo

Issue #867, opened 2024-02-13
Hide the details button on reports when there are no failures


Julian:
On dialect pages we show details buttons next to implementations:

![Screenshot 2024-02-13 at 10 38 37 AM](https://github.com/bowtie-json-schema/bowtie/assets/329822/02dd78e9-bf1b-43d8-92e9-1357c244283a)

The modals these open are intended to show information about unsuccessful tests.

It's confusing however that for any passing implementation (all 0's) the modal shows a blank page.

We should likely not show the button at all for any implementation where it would be empty.

sudo-jarvis:
Hi @Julian , solved this in #869.

PR #869, opened 2024-02-13
Hide the details button on reports when there are no failures

sudo-jarvis:
**This pull request solves #867.**

**Implementation:**
 Dont show the details button if the implementations has no errored, no skipped and no failed tests.

Screenshot:
Screenshot 2024-02-13 at 9 59 55 PM

----
📚 Documentation preview 📚: https://bowtie-json-schema--869.org.readthedocs.build/en/869/

### DIFF
--- a/frontend/src/components/Summary/ImplementationRow.tsx
+++ b/frontend/src/components/Summary/ImplementationRow.tsx
@@ -49,13 +49,13 @@ const ImplementationRow = ({
       </td>
 
       <td>
-        <button
+        {implementation.failedTests + implementation.erroredTests + implementation.skippedTests > 0 && <button
           type="button"
           className="btn btn-sm btn-primary"
           onClick={() => setShowDetails(true)}
         >
           Details
-        </button>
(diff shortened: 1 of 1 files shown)


frontend/src/components/Summary/ImplementationRow.tsx
```diff
@@ -49,13 +49,18 @@ const ImplementationRow = ({
...
 
       <td>
-        <button
-          type="button"
-          className="btn btn-sm btn-primary"
-          onClick={() => setShowDetails(true)}
-        >
-          Details
-        </button>
+        {implementation.failedTests +
```
Julian, line 52: Great. Can you wrap this in a `.hasUnsuccessful` property on implementations, this same addition is used in a few places I think.

sudo-jarvis: But in the unsuccessful column we just add the `failed tests + errored tests` but here we are adding skippedTests as well.

PR #876, opened 2024-02-14
Fixed #873 issue. UI problem

### LINKED ISSUE #873: Rows without details buttons are not the same height as others on the UI report
After #869 we now have rows in the table which are different heights than others:

Screenshot 2024-02-13 at 2 05 35 PM

which is visually inconsistent.

We should make the row height the same even if the button isn't present.

The text is also not vertically centered in rows *with* the button, so we should fix that as well.

XDreamist:
This pull request is a fix for https://github.com/bowtie-json-schema/bowtie/issues/873.

Screenshot (82)

* The size of the rows are fixed as every row now have the same size even if it has the button or not. 
* The text has also been aligned to the center making it a clean look.

----
📚 Documentation preview 📚: https://bowtie-json-schema--876.org.readthedocs.build/en/876/

### DIFF
--- a/frontend/src/components/Summary/ImplementationRow.css
+++ b/frontend/src/components/Summary/ImplementationRow.css
@@ -1,4 +1,4 @@
-td{
+td {
   height: 40px;
 }
 
--- a/frontend/src/components/Summary/ImplementationRow.css
+++ b/frontend/src/components/Summary/ImplementationRow.css
@@ -1,3 +1,7 @@
+td{
+  height: 40px;
+}
+
 .details-required {
   position: relative;
 }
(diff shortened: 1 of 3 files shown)


frontend/src/components/Summary/ImplementationRow.css
```diff
@@ -1,4 +1,4 @@
-td{
+td {
   height: 40px;
```
Julian, line 2: I don't think px is the right unit to use here, AFAIK (which isn't much) one should use `rem`s for spacing, but we use bootstrap, so we should presumably be using bootstrap's mechanisms for spacing.

XDreamist: Ok! I'll check that out and will let you know!

PR #877, opened 2024-02-14
Render unsupported implementations

### LINKED ISSUE #510: Consider some way of indicating implementations which don't support a dialect on the dialect report page
Even though we obviously cannot report on dialects that an implementation doesn't support, it still may be nice to show a hint that other implementations exist for a dialect.

In other words, if someone goes to the draft 7 page and looks at Python implementations and sees 2 of them, it may be nice if there was a subtle indication that "hey there's a third implementation, though it doesn't support this version", perhaps hidden behind a collapsable element saying "Other Implementations" which appears expandable below the summary table.

adwait-godbole:
Fixes #510 

Screenshot (53)
Screenshot (54)

Will have to wait until other PRs get merged

Also the UI build is failing because the drag and drop element also renders the Report page so do we want to show "Other Implementations" there as well @Julian  ?


adwait-godbole:
Ohh actually this PR also depends on #855 because it uses the `implementations.json`  file. That file is acting like a lookup file to get all implementations across drafts as we had discussed earlier in the issue.

Julian:
Got it -- but just #855 then? Or also #865?

Julian:
Yeah should be minor I hope!

adwait-godbole:
@Julian But I see that the changes you made makes the code output the JSON file always with `indent=2`. Previously we had discussed the following condition:

```python
if sys.stdout.isatty() and implementations_len == 1:
         kwargs["indent"] = 2
```

should i make these changes as well ?

Julian:
No we don't need that anymore essentially -- the reason we needed that was to be able to parse our output -- previously that was hard to do because it was concatenated JSON -- but now it's easy to do, because it's always valid JSON (it's always a JSON object).

Lemme know if that makes sense!

adwait-godbole:
Ohh! Right! I got it now!

adwait-godbole:
Hey @Julian. I finally have this sorted out now. The UI will look the same as in my screenshot. Before merging / deploying, we will have to make sure that indeed the "implementations.json" file is now getting successfully created otherwise the UI will break.

Julian:
Cool! The file is there already I believe right? So should be working?

adwait-godbole:
Ohh right it is working!

adwait-godbole:
> I haven't actually looked at your design until now, I'll have a look at that in a bit, so may have some other comments soon.

Sure

adwait-godbole:
Hey @Julian morning! Was waiting for your feedback yesterday

Julian:
I probably won't look in depth this weekend, will likely disconnect the next 2 days, but I think my first reaction is probably to make this a bit *less* prominent. Specifically I think this would be a bit less "in the way" if it was an "other implementations" as a small button or link or something *inside* the summary box, with just a tooltip that showed these implementations, rather than an accordion.

Basically we want someone to notice this but not *so* prominently. But that's just a first thought, not sure exactly what the right design is honestly yet.

Also no need to include versions for these implementations, just the name, and a link to that implementation's page I think.

Julian:
Or even some small link text saying "There are other implementations besides the above", where if you hover over it shows the tooltip?

frontend/src/index.tsx
```diff
@@ -39,6 +43,20 @@ const fetchAllReportData = async (langImplementation: string) => {
   return parseImplementationData(loaderData);
 };
 
+const fetchImplementationStatsData = async () => {
```
Julian, line 46: I think sometime soon I'd like to move to some `Client` object which encapsulates this logic that we keep repeating -- i.e. we should have `Client.implementations()`, `Client.report(dialect)`, (and then any future things we add), rather than sprinkling this logic each time we add a file / change the data layout.

But of course we can handle that in a future PR, just mentioning it for awareness.

adwait-godbole:
Well I went with the Accordion because that's what you had mentioned earlier on the issue description. But yeah open to these remarks as well.

adwait-godbole:
I will have a look with the tooltip / link kind of design decisions if they fit well  so we keep it less prominent.

Julian:
Got it yeah feel free to experiment a bit then, I may not have fully thought it through until now seeing it "real"!

adwait-godbole:
Hey @Julian what are your thoughts about the above screenshots ?

adwait-godbole:
Also @Julian a note here, currently the behaviour is such that if for example we are on draft 7 and some versions of python support draft 7 but some don't then the unsupported ones of python show up in this tooltip / popover as requested by you in the issue but let's say if there were no versions of Clojure that supported draft 7 however Clojure does support some other drafts but then still Clojure will not show up in the unsupported tooltip because it wasn't a part of draft 7 langs. Its slight difficult for me to explain this in words but in short we only show the unsupported versions for languages for the current selected draft and not for any other programming langs if they aren't a part of draft 7. Does that make sense 😅 or should i elaborate more ?

adwait-godbole:
In my view this approach (as was expected by you) makes more sense to see the other implementations if language filters are applied because if someone applied the python language filter then he / she will only see the unsupported versions of python and won't see Clojure come up in there

Julian:
> In my view this approach (as was expected by you) makes more sense to see the other implementations if language filters are also applied because if someone applied the python language filter then he / she will only see the unsupported versions for python filtered as well and won't see Clojure or any other lang there irrespective if it was a part of draft 7 or not. This makes it easier for user to see the unsupported lang (if any) according to his / her selected filter as well.

I agree with this I think.

In other words, we would show the unsupported implementations hover *only* if you filter a language, and only if there are then other implementations besides the ones you see. That's what you were suggesting/asking as well right?

adwait-godbole:
Yeah the filtering part was just as an example because that truly highlights the use case of having this feature but even if no filters are applied we can still see the unsupported langs for the current draft 7 langs

adwait-godbole:
> Hey @Julian what are your thoughts about the above screenshots ? Which alignment do you prefer ? Left or Top ?

Let me know about this. Also its not really hover right now, if you click it then it shows up and if you click away somewhere else then it closes. I tried the hovering approach but for some reason the hover does not really let you click the link inside the popup, even before you click it vanishes.

adwait-godbole:
I will make the changes in this PR so you are able to try it out yourself.

Julian:
Yeah that sounds good! I think even more conspicuous (e.g. using `text-body-secondary` and maybe a separate button intead of link text) might look good, but it's hard to suggest without playing around.

adwait-godbole:
I have pushed the changes. You can try out!

adwait-godbole:
> Yeah that sounds good! I think even more conspicuous (e.g. using `text-body-secondary` and maybe a separate button intead of link text) might look good, but it's hard to suggest without playing around.

text-body-secondary on what exactly ? and what will the button say ?

Julian:
> text-body-secondary on what exactly

On a sentence saying "Other implementations are available which do not support the current dialect and filters."

And then a button with just. the `i` icon next to that to show your hover.

Not sure if that makes sense describing in text without a mockup -- or how good that will look without playing a bit.

adwait-godbole:
Got it! I will make the changes and push again.

adwait-godbole:
Hey @Julian how about now ?

adwait-godbole:
Sorry forgot the text you gave. Once again made the changes!

adwait-godbole:
@Julian have a look now I made some styling changes!

adwait-godbole:
Also hover effect achieved!

frontend/src/DialectReportView.tsx
```diff
@@ -22,27 +28,82 @@ export const DialectReportView = ({
...
+    if (selectedLanguages.length > 0) {
+      const filteredReportArray = Array.from(
         reportData.implementations.entries(),
       ).filter(([, data]) =>
-        params.getAll("language").includes(data.metadata.language),
+        selectedLanguages.includes(data.metadata.language),
+      );
+      const filteredReportImplementationsMap = new Map(filteredReportArray);
+      filteredReportData.implementations = filteredReportImplementationsMap;
+      if (allImplementationsData) {
```
harrel56, line 43: Hard to examine all the possible scenarios - can `allImplementationsData` be falsy?

frontend/src/DialectReportView.tsx
```diff
@@ -22,27 +28,82 @@ export const DialectReportView = ({
...
+        filteredOtherImplementationsData = filterOtherImplementations(
+          allImplementationsData,
+          languages,
+          filteredReportImplementationsMap,
+        );
+      }
     }
-    return filteredData;
-  }, [reportData, params]);
+    return { filteredReportData, filteredOtherImplementationsData };
```
harrel56, line 63: Maybe it would be easier to read if this memo was split into two? It's pretty complex now

frontend/src/DialectReportView.tsx
```diff
@@ -22,27 +28,82 @@ export const DialectReportView = ({
...
+  langs: string[],
+  filteredReportImplementationsMap: Map<string, ImplementationData>,
+): Record<string, Implementation> => {
+  const filteredOtherImplementationsArray: [string, Implementation][] =
+    Object.entries(allImplementationsData).filter(
+      ([, impl]: [string, Implementation]) => langs.includes(impl.language),
+    );
+
+  const filteredOtherImplementationsData: Record<string, Implementation> = {};
+  filteredOtherImplementationsArray.forEach(([key, impl]) => {
```
harrel56, line 95: This forEach basically transforms this collection - using `map()` should be more clear

frontend/src/DialectReportView.tsx
```diff
@@ -22,27 +28,82 @@ export const DialectReportView = ({
...
+    let isFiltered = false;
+    for (const val of filteredReportImplementationsMap.values()) {
+      if (impl.source === val.metadata.source) {
```
harrel56, line 98: It looks like both collections have keys - cannot we just compare the keys?

harrel56: Additionaly, `filteredReportImplementationsMap` is a map, can't we just make use of `get()` method?

frontend/src/components/DragAndDrop/DragAndDrop.tsx
```diff
@@ -17,7 +17,7 @@ export const DragAndDrop = () => {
     setLines(null);
     setDragActive(false);
     setInvalidFile(false);
-  }, [location]);
```
harrel56, line 20: I'm not following - is it related to the PR topic? In any case could you explain what it does?

frontend/src/components/OtherImplementations.tsx
```diff
@@ -0,0 +1,79 @@
...
+          transition={false}
+        >
+          {(props) => (
+            <Popover id="popover-basic" {...props}>
+              <Popover.Body>
+                {Object.entries(otherImplementationsData).map(
+                  ([id, impl], index) => {
+                    const implementationPath = getImplementationPath(id);
+                    return (
+                      <p key={index}>
```
harrel56, line 45: Using index as a key is generally discouraged. I think you can just use `id`? It uniquely identifies a row, right?

frontend/src/index.tsx
```diff
@@ -48,13 +66,24 @@ const router = createHashRouter([
...
       {
         path: "/dialects/:draftName",
         Component: ReportDataHandler,
-        loader: async ({ params }) =>
-          fetchReportData(Dialect.forPath(params.draftName!)),
+        loader: async ({ params }) => {
+          const reportData = await fetchReportData(
+            Dialect.forPath(params.draftName!),
+          );
+          const allImplementationsData = await fetchImplementationStatsData();
```
harrel56, line 84: These API calls are sequential but could be run in parallel. Also the 2 loader functions are almost exactly the same - as this is only 5 lines long I don't really care, but if you want you can extract the common code

frontend/src/components/OtherImplementations.tsx
```diff
@@ -0,0 +1,79 @@
...
+            </Popover>
+          )}
+        </Overlay>
+        <div className="d-flex align-items-center text-body-secondary">
+          <InfoCircle />
+        </div>
+      </div>
+      <div className="text-body-secondary ps-2">
+        {
+          "Other implementations are available which do not support the current dialect and filters."
```
harrel56, line 69: This is just a plain text, you can write it directly as a part of HTML - without braces and quotations

frontend/src/DialectReportView.tsx
adwait-godbole: Yes @harrel56  actually on the `DragAndDrop` component when the `DialectReportView` gets rendered we don't have the `allImplementationsData`  over there so I made it optional.

harrel56: Oh right, now I see it. So maybe it's better to make it required and pass empty object in `DragAndDrop`? That way you can get rid of the ifs and it would simplify it a bit. But that's just a suggestion

adwait-godbole: No problem I will do that

frontend/src/components/DragAndDrop/DragAndDrop.tsx
adwait-godbole: Yeah actually it isn't related to this PR. During the PR #849  I had done this fix, although it worked but later I realized in the useEffect dependency I had directly passed `location` which is a bad practice because its an object. It should have been `location.key`. This was just a 1 one line fix so I included it over here.

harrel56: Makes sense but they actually still use it with object in their docs :): 
image

But it's fine, current solution looks more reasonable to me

adwait-godbole: Okay then I will remove this!

adwait-godbole: Okay!

adwait-godbole:
Hey @harrel56 Thanks for the constructive feedbacks! I have made all the changes you suggested above. Please have a look!. Also @Julian let me know about the hover effect if that's what you had in mind!

adwait-godbole:
Hey @harrel56 Thanks for the constructive feedbacks! I have made all the changes you suggested above. Please have a look!. Also @Julian let me know about the hover effect if that's what you had in mind!

Julian:
What you did is exactly the sort of thing I had in mind, and looks great! I suspect it'd be helpful to have one other extra piece of info in the hover, maybe the latest dialect that implementation supports, as a link to that dialect's report? But it may also be perfect as-is.

I trust @harrel56 completely :), so Harrel maybe if you're gonna have a look feel free to approve/merge whatever.

@adwait-godbole also probably need to fix the merge conflicts, which might have come from the other 2 PRs I just merged. Hopefully shouldn't be too difficult.

frontend/src/DialectReportView.tsx
```diff
@@ -84,23 +89,16 @@ export const DialectReportView = ({
...
-  filteredOtherImplementationsArray.forEach(([key, impl]) => {
-    let isFiltered = false;
-    for (const val of filteredReportImplementationsMap.values()) {
-      if (impl.source === val.metadata.source) {
-        isFiltered = true;
-        break;
-      }
-    }
-    if (!isFiltered) {
+  filteredOtherImplementationsArray.map(([key, impl]) => {
```
harrel56, line 100: Maybe I mislead you a bit with `map()` - but you are using it here incorrectly, mapping logic should be side effect free and should return transformed/mapped value for each element.

Actually, just filtering is enough:
```ts
  const filteredOtherImplementationsArray = Object.entries(allImplementationsData)
    .filter(([, impl]) => langs.includes(impl.language),)
    .filter(([key]) => !filteredReportImplementationsMap.has(key));
  return Object.fromEntries(filteredOtherImplementationsArray);
```
I believe this should do the same thing and is more concise. But please ensure if I didn't miss anything :)

PS: IMO annotating everything with TS types is just making things less readable. As a rule of thumb you shouldn't write any types as long as type inference works correctly. In the case above I just get rid of all the annotations and the type information is not lost and you're getting all the benefits the same way you were getting before

frontend/src/DialectReportView.tsx
```diff
@@ -16,66 +16,71 @@ export const DialectReportView = ({
...
+    }
+
+    return filteredData;
+  }, [reportData, params]);
+
+  const filteredOtherImplementationsData = useMemo(() => {
+    let filteredData: Record<string, Implementation> = {};
+    const selectedLanguages = params.getAll("language");
+
+    if (selectedLanguages.length > 0) {
```
harrel56, line 51: That just does the same work twice. You did the work while calculating `filteredReportData` and you can reuse it here. It simplifies thing greatly. I believe the whole useMemo logic can just become:
```ts
    const availableLanguages = selectedLanguages.length ? selectedLanguages : languages
    return filterOtherImplementations(
      allImplementationsData,
      availableLanguages,
      filteredReportData.implementations,
    );
```
But again, please check if I didn't miss anything and if it will work correctly :)

adwait-godbole: Yes! I should've catched this! Sorry!

frontend/src/DialectReportView.tsx
```diff
@@ -16,66 +16,71 @@ export const DialectReportView = ({
...
 
     if (selectedLanguages.length > 0) {
       const filteredReportArray = Array.from(
-        reportData.implementations.entries(),
+        reportData.implementations.entries()
       ).filter(([, data]) =>
-        selectedLanguages.includes(data.metadata.language),
+        selectedLanguages.includes(data.metadata.language)
       );
       const filteredReportImplementationsMap = new Map(filteredReportArray);
```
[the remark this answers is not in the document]
adwait-godbole: I don't quite get this one we do need to set `filteredData.implementations` that map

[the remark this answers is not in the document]
harrel56: this is enough: `filteredData.implementations = new Map(filteredReportArray);`. No need for defining `const` here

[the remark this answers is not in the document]
adwait-godbole: Ohh! Okay!

adwait-godbole:
@harrel56 can you check now ?

frontend/src/components/OtherImplementations.tsx
```diff
@@ -38,22 +41,44 @@ export const OtherImplementations = ({ otherImplementationsData }: Props) => {
...
+                              to={`/implementations/${implementationPath}`}
+                            >
+                              {impl.name}
+                            </NavLink>
+                            <span className="ps-1 text-body-secondary fw-bold">
+                              {mapLanguage(impl.language)}
+                            </span>
+                          </div>
+                          <span className="text-body-secondary">
+                            (LTS Dialect:{" "}
```
harrel56, line 71: 1. is LTS really an abbreviation of "latest"?
2. Please dons' use spaces for positioning elements - use CSS classes instead

frontend/src/components/OtherImplementations.tsx
```diff
@@ -38,22 +41,44 @@ export const OtherImplementations = ({ otherImplementationsData }: Props) => {
...
+                  <Row className="flex-column">
+                    {otherImplementationsDataArray.map(([id, impl], index) => {
+                      const implementationPath = getImplementationPath(id);
+                      const latestSupportedDialect =
+                        getLatestSupportedDialect(impl);
+                      return (
+                        <Col
+                          key={id}
+                          className={
+                            index === otherImplementationsDataArray.length - 1
```
harrel56, line 54: I kind of understand what are you doing here, but for such cases flexbox comes in handy. I believe `<Col>` component already has `display: flex` so you only need to set gap - always prefer spacing elements in a list with a `gap` instead of `margin`

frontend/src/components/OtherImplementations.tsx
```diff
@@ -74,3 +99,12 @@ const getImplementationPath = (id: string): string => {
   const pathSegment = id.split("/");
   return pathSegment[pathSegment.length - 1];
 };
+
+const getLatestSupportedDialect = (impl: Implementation): string => {
+  const latestSupportedDialect = impl.dialects.sort((a, b) =>
+    b.localeCompare(a)
```
harrel56: You sort it by name which not necessarily might be a correct way for finding the latest dialect. You can first map the `dialects` array elements to instances of Dialect (`Dialect.foPath()`) and then sort it. Or even better, because you actually don't need to sort the whole array, just find min element (by publish date). `reduce()` can do that stuff

harrel56: Actually, `Dialect.foPath()` won't work, so bigger changes are required :) Why not change the type of `Implementation.dialects` to be `Dialect[]` and assign it when parsing implementation data?

In `Dialect.ts` you can provide an additional method `forUri()` - don't really see a very clean way of implementing this, but you can figure it out

harrel56 (commented):
TBH I don't know why you added more stuff - but you can address my comments or postpone these changes for the followup PR (I mean either get rid of the new changes or correct them). Otherwise, everything looks fine and ready to merge

frontend/src/components/OtherImplementations.tsx
adwait-godbole: should I change LTS to "Latest Supported Dialect" then ?

frontend/src/data/Dialect.ts
```diff
@@ -55,6 +55,16 @@ export default class Dialect {
     }
     return dialect;
   }
+
+  static forURI(baseURI: URI): Dialect {
```
harrel56, line 59: Ok, I see you already correct it. But still, splitting URI path is a fragile logic. In `dialects.json` there is already an association with dialect URI - it's our source of truth so it'd be better if you somehow make use of this value instead

frontend/src/components/OtherImplementations.tsx
harrel56: Not really feeling strong in english writing :) but I'd say yes - and I don't know if capitalization is correct here? It's not some kind of title or anything

adwait-godbole: > actually as it's text next to other text, it's better the way it is

So do I leave it as it is ?

harrel56: > So do I leave it as it is ?

Yep

frontend/src/components/OtherImplementations.tsx
adwait-godbole: Which approach do you suggest I take here ? I have already added `forURI()`  as you pointed out below. I can make the changes you suggested below but overall what are you inclined towards ?

frontend/src/components/OtherImplementations.tsx
adwait-godbole: What is the exact keyword for `gap` here ? I tried doing `gap-2` and also `flex-gap-2` both didn't work

frontend/src/components/OtherImplementations.tsx
adwait-godbole: We could use `find()` as well right ?

frontend/src/components/OtherImplementations.tsx
adwait-godbole: Ohh gap works on the grid ig then. Will do the changes 👍

adwait-godbole:
@harrel56 Thanks again for all these suggestions I have tried doing as many of the changes you asked for but not sure if I was able to crack the `URI` and `Dialect` logic completely but it does work using `find()` since we are only looking for a single match.

adwait-godbole:
@harrel56 Pushed the changes!

adwait-godbole:
@harrel56 made the changes!

adwait-godbole:
Thanks a lot @harrel56 for all your suggestions. Had ton of learnings in just a single PR!

PR #878, opened 2024-02-14
Fix #873 issue. UI problem

### LINKED ISSUE #873: Rows without details buttons are not the same height as others on the UI report

XDreamist:
This pull request is a fix for https://github.com/bowtie-json-schema/bowtie/issues/873.

Screenshot (82)

Fixed the size problem without any css modification.

----
📚 Documentation preview 📚: https://bowtie-json-schema--878.org.readthedocs.build/en/878/

### DIFF
--- a/frontend/src/components/Summary/ImplementationRow.css
+++ b/frontend/src/components/Summary/ImplementationRow.css
@@ -1,4 +1,4 @@
-td{
+td {
   height: 40px;
 }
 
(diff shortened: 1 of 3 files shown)


XDreamist:
I hope you are satisfied with this solution.

Julian:
Looks great I think, thanks!

XDreamist:
No problem! ☺️

PR #879, opened 2024-02-15
fix-#333: add extra info on UI

Siddharth-Singh-2004:
fix #333: Added code to Implementation Info page to display Additional Links (URLs) provided by JSON schema implementers.

----
📚 Documentation preview 📚: https://bowtie-json-schema--879.org.readthedocs.build/en/879/

PR #882, opened 2024-02-15
code optimisation introduced an UI error #879

Siddharth-Singh-2004:
@harrel56, shortened version of code introduced a minor UI error, hence reverted to verbose code from previous version. See screenshots below.

Summarising from #879 here:
The shortened code, implementation.links?.length, leads to a UI bug (see the 0 displayed) when no additional Links are present:

screen1

The long form of code, implementation.links && !!implementation.links.length, meets the expected behavior:

screen2

The shortened version seems to be printing the value of length when there is no link to display. I'll revert to the long form of code.

----
📚 Documentation preview 📚: https://bowtie-json-schema--882.org.readthedocs.build/en/882/


harrel56 (approved):
Yeah, sorry about that, conditional rendering in react still surprises me :D Good thing you noticed it

Issue #368, opened 2023-08-02
Show a description of what Bowtie is on the homepage


Julian:
It'd be nice to somewhere show some short description of what someone is looking at when visiting Bowtie's report.

Min2who:
> It'd be nice to somewhere show some short description of what someone is looking at when visiting Bowtie's report.

can you explain a bit about what you want ?

cc @Julian

Siddharth-Singh-2004:
@Julian , I am working on this issue. Kindly assign this issue to me.

Siddharth-Singh-2004:
There are many ways of implementing this requirement:
1.  Leveraging tooltip (really short piece of text)
2. Showing text on hover (++tooltip with additional information)
3. Implementing cards (for more visual context, such as images )

As I see it, this action would be used only by somehow who is new to the UI and the first few times they use the UI until it all feels familiar. 

Here's an implementation going with option 2, to display text on hover:
Before the change:

<img width="1664" alt="Bowtie-Before" src="https://github.com/bowtie-json-schema/bowtie/assets/112953828/785c7581-a612-4370-9311-30be34bf9752">

After implementing hover for the docs button:
Clicking on Docs on the hover takes user to Bowtie documentation as before.

<img width="1664" alt="Bowtie-After-ThemeDark" src="https://github.com/bowtie-json-schema/bowtie/assets/112953828/9a2b303e-3f2b-4e99-bb88-f353352dd32c">

<img width="1663" alt="Bowtie-After-ThemeLight" src="https://github.com/bowtie-json-schema/bowtie/assets/112953828/dae641dc-67d3-4f72-90cf-bb40017b14f4">

Julian:
Bowtie's docs are more about how you *use* Bowtie, particularly its CLI.

This issue is more about the report / the web page you're looking at.

If I asked you what you were looking at at https://bowtie.report -- what is it? What is it for?

That's the kind of information this issue is looking to add somewhere on the page.

Siddharth-Singh-2004:
Just to clarify: Is it just on the navbar elements or any other section of what is visible? As a suggestion: Clicking the bowtie icon on the top left of the page would be the right place for that kind of info. However, we can consider adding such info to various "#" sections on the navbar as well. My implementation was to highlight the choice of UI elements and I could implement the same for any of the elements of the navbar.

Julian:
This isn't about any of the navbar elements. It's to add detached information about what is in the report, and why someone may care!

[...]

PR #885, opened 2024-02-17
Bowtie Info on Homepage

Siddharth-Singh-2004:
fix-#368 to show a description of what Bowtie is, on the homepage

----
📚 Documentation preview 📚: https://bowtie-json-schema--885.org.readthedocs.build/en/885/

### DIFF
--- a/frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
+++ b/frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
@@ -7,15 +7,37 @@ const BowtieInfoSection = () => {
           <thead>
             <tr>
               <td>
-              Bowtie is tool for understanding and comparing implementations of the &nbsp; 
-               <a href={"https://json-schema.org/"} target={"_blank"} rel={"noreferrer"}>{"JSON Schema specification"}</a>  &nbsp;
-               across all programming languages. 
-              The report below uses the &nbsp;
-              <a href={"https://github.com/json-schema-org/JSON-Schema-Test-Suite"} target={"_blank"} rel={"noreferrer"}>{"official JSON Schema Test Suite"}</a> &nbsp; 
-               to display bugs or functionality gaps in implementations. 
-              You can also use Bowtie to get this information about your own schemas. 
-              You can find out how in its &nbsp;
-              <a href={"https://docs.bowtie.report/"} target={"_blank"} rel={"noreferrer"}>{"documentation"}</a>.
+                Bowtie is tool for understanding and comparing implementations
+                of the &nbsp;
+                <a
+                  href={"https://json-schema.org/"}
+                  target={"_blank"}
+                  rel={"noreferrer"}
+                >
+                  {"JSON Schema specification"}
+                </a>{" "}
+                &nbsp; across all programming languages. The report below uses
+                the &nbsp;
+                <a
+                  href={
+                    "https://github.com/json-schema-org/JSON-Schema-Test-Suite"
+                  }
+                  target={"_blank"}
+                  rel={"noreferrer"}
+                >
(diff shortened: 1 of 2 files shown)


frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <table className="table table-sm table-hover">
+          <thead>
+            <tr>
+              <td>
+                Bowtie is tool for understanding and comparing implementations
+                of the &nbsp;
+                <a
+                  href={"https://json-schema.org/"}
+                  target={"_blank"}
+                  rel={"noopener noreferrer"}
```
Julian, line 15: It's fine to include `referrer` here, linking to the JSON Schema site is essentially "internal" inasmuch as we probably would be interested to know if lots of traffic was being generated from Bowtie's site.

```suggestion
                  rel={"noopener"}
```

frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+                  {"JSON Schema specification"}
+                </a>{" "}
+                &nbsp; across all programming languages. The report below uses
+                the &nbsp;
+                <a
+                  href={
+                    "https://github.com/json-schema-org/JSON-Schema-Test-Suite"
+                  }
+                  target={"_blank"}
+                  rel={"noopener noreferrer"}
```
Julian, line 26: And here, same.
```suggestion
                  rel={"noopener"}
```

frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+                >
+                  {"official JSON Schema Test Suite"}
+                </a>{" "}
+                &nbsp; to display bugs or functionality gaps in implementations.
+                You can also use Bowtie to get this information about your own
+                schemas. You can find out how in its &nbsp;
+                <a
+                  href={"https://docs.bowtie.report/"}
+                  target={"_blank"}
+                  rel={"noopener noreferrer"}
```
Julian, line 36:
```suggestion
                  rel={"noopener"}
```

And certainly in Bowtie's own docs.

Julian:
Thanks! Just some notes about not scrubbing referrers, but this looks good afterwards, appreciated!

Siddharth-Singh-2004:
@Julian,

I presume we want links to json schema.org, test suite etc., **to open in a new tab/window** based on user's browser preferences. If that is not a requirement, then I just remove the "target=_blank" and you don't have to read any further. Else, there are some nuances and I need your concurrence to make the changes.

If I use "a href" tag with target={"_blank"}, then nox's  "ui(style)" checker throws up errors when rel="noopener" is used without rel="noreferrer" (don't understand why it is so).

Here's a screenshot of the error when code is such..
 <a
                  href={"https://docs.bowtie.report/"}
                  target={"_blank"}
                  rel={"noopener"}
 >

Errors:
Screenshot 2024-02-17 at 9 38 20 PM

Ask from you:
I have two choices, either fix nox checks for "ui(style)" to not mandate "rel=noreferrer" (when target is used, not investigated if this is supported in nox, but will in due course) or remove target option from <a href>. 

Please let me know how you want me to proceed.

Julian:
We may not care about such old browsers but I'm less inclined to do the research right this second -- it's fine to just remove the `target: _blank` entirely and have it load in the same tab then!

Siddharth-Singh-2004:
Thanks for the quick revert. I'll have the fix in shortly.

Julian:
Great! Think this looks good! Merging, thanks again!

PR #894, opened 2024-02-18
Navigate to failed test from Modal

### LINKED ISSUE #871: Details modals for failed tests in the UI should also show the schema + instance for the relevant test
When an implementation fails a test, we show the failed test in a modal which is opened when clicking the details buttons on the dialect report:

Screenshot 2024-02-13 at 11 56 04

which shows:

Screenshot 2024-02-13 at 11 56 08

Someone looking at that doesn't have any way of knowing what the contents of that test case are, even though they could go back and find it on the details part of the report. In the screenshotted case, that's:

Screenshot 2024-02-13 at 11 57 33

We should make it extremely easy to get from the modal to that information.

The ideal way to handle this would be to have fragments (IDs) associated with each test+implementation, and to have the modal contain a link that jumps directly to the fragment for that test.

An interim solution if that's tricky is to instead simply have an expandable accordion in the modal which shows the same information again.

adwait-godbole:
Fixes #871 

https://github.com/bowtie-json-schema/bowtie/assets/68469605/ca3dc7cf-a109-424b-81c5-0dac1d10b323

### DIFF
--- a/frontend/src/components/Cases/CaseItem.tsx
+++ b/frontend/src/components/Cases/CaseItem.tsx
@@ -116,8 +116,8 @@ const CaseItem = memo(function CaseItem({
           caseData={caseData}
           implementations={implementations}
           schemaDisplayRef={schemaDisplayRef}
-        />
-      )
+        />,
+      ),
     );
   }, [seq, caseData, implementations]);
   return (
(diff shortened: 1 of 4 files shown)


Siddharth-Singh-2004:
@adwait-godbole, I was working on this issue #871 as indicated in the comments. We should not be duplicating effort. 

cc @Julian

adwait-godbole:
Okay @Siddharth-Singh-2004 apologies for the misunderstanding. I have closed the PR.

Issue #19, opened 2022-09-13
Generalize implementations beyond containers


Julian:
I.e. be able to harness into a container running perhaps locally, or over HTTP, or over some general interface, rather than via a managed docker container.

PR #895, opened 2024-02-18
Creating `implementations.json` file with no remote data to fix #510

### LINKED ISSUE #510: Consider some way of indicating implementations which don't support a dialect on the dialect report page

adwait-godbole:
Hey @Julian as you suggested I broke the PR #855 in this part so we have no remote data in the `implementations.json` file for now. 

Also a note I have added a field called `"image"` in the `implementations.json` file so that when we render unsupported implementations for a dragt in the UI we have a clickable link on them because currently the implementations report view  page depends on the last string of the image name to navigate to it. 
For e.g. in this url  `https://bowtie.report/#/implementations/js-hyperjump` , `js-hyperjump` comes from its image name i.e.
`"ghcr.io/bowtie-json-schema/js-hyperjump"` 

Also can you help me out here
`Build the UI failed because ghcr.io/bowtie-json-schema/lua-jsonschema' failed to start.` ?

Is it the `no space left on device` error ? As mentioned by you in PR #855 I removed the freeing up disk space logic.


adwait-godbole:
PS. Build the UI failed because ghcr.io/bowtie-json-schema/lua-jsonschema' failed to start. ?

Julian:
> Also can you help me out here Build the UI failed because ghcr.io/bowtie-json-schema/lua-jsonschema' failed to start. ?

Running the command is failing, but it's hard to tell why. You should diagnose by re-running it in simpler ways until you see the real error, and then possibly we're missing some error handling.

> Also a note I have added a field called "image" in the implementations.json file

I don't want to add that field -- it eventually will be an implementation detail that an implementation has a container image (#19) -- you can look through the existing UI code, right now it generates that field from others (language and name, and a fixed prefix). But we should not rely on it being present.

adwait-godbole:
@Julian in the UI code I see no mention of it being generated from name and language. Correct me if I am wrong. You can see in `ImplementationRow.tsx` file the clickable link to every implementation comes from its `id` which is basically is the image string. Also the image string is present in the output of `draft{{version}}.json` from `bowtie suite` command file in the key `"implementation"`. The id is getting derived from this key `"implementation"`.

adwait-godbole:
> I don't want to add that field -- it eventually will be an implementation detail that an implementation has a container image (#19) -- you can look through the existing UI code, right now it generates that field from others (language and name, and a fixed prefix). But we should not rely on it being present.

I tried freeing up the disk space but still got `System.IO.IOException: No space left on device` error

Julian:
> I added freeing up the disk space logic but still got System.IO.IOException: No space left on device error

Where are you seeing this error?

Julian:
Where? Here I see https://github.com/bowtie-json-schema/bowtie/actions/runs/7953442753/job/21709444279?pr=895#step:5:5

Julian:
Oh I see, it's only on the summary page

adwait-godbole:
> Where? Here I see https://github.com/bowtie-json-schema/bowtie/actions/runs/7953442753/job/21709444279?pr=895#step:5:5

Yes its pretty confusing this page shows `ghcr.io/bowtie-json-schema/dotnet-jsonschema-net' failed to start.` whereas the checks tab shows `System.IO.IOException: No space left on device`.

adwait-godbole:
Hey @Julian I came across this thread https://github.com/orgs/community/discussions/26351 which discuess the no space left issue.
I will try hacking around if any solutions mentioned in here works.

adwait-godbole:
@Julian any luck on the space issue ? 😅

Julian:
Not yet unfortunately, other than to convince myself it's not the host machine filling up, as it has 30GB+ of free space when we run a report.

Still working on it...

Julian:
OK I know what the issue is now finally. Should have it resolved today I hope.

Julian:
OK this is fixed now in GH Actions. It's not fixed elsewhere yet (in that basically it will still happen when Bowtie pulls the image, but not if you run `docker pull`...) I still don't 100% understand what in aiodocker is causing that behavior. That will be the next thing I fix -- but anyways, you should be unblocked here now -- you'll notice the workflow has changed, hopefully it's straightforward now to emit this data. Will circle back and give some more tips.

Julian:
Oh, one specific one -- do this in the `report.yml` workflow rather than the build-frontend one. You should see an obvious spot for that to happen now.

I'll likely end up combining these two workflows, or at least trying to remind myself why we have two.

adwait-godbole:
@Julian it finally worked! Please have a look overall and then we can hash out PR #877

adwait-godbole:
Also @Julian let me know your thoughts about this one, so its finally hashed out

Julian:
I am tweaking it slightly because of the `image` thing I mentioned earlier, but glad it's working finally! Will merge in a few hours.

Julian:
OK I tweaked the format slightly here so that the image is the key of a dictionary in the case we have multiple implementations.

That makes it a bit easier to avoid some of what I was trying to get at above (that we won't always have an image, even though we of course always need an ID for the implementation -- there's probably more to clean up there, you're right I'm sure about what you said is happening in the UI, so I'll have to look there again).

But for now I *think* the tweak I made shouldn't complicate things too much for you in the other issue. Have a look and lemme know -- and thanks for your persistence!

PR #905, opened 2024-02-20
Details modals for failed tests in the UI should also show the schema + instance #871

Siddharth-Singh-2004:
Details modals for failed tests in the UI should also show the schema + instance #871

----
📚 Documentation preview 📚: https://bowtie-json-schema--905.org.readthedocs.build/en/905/


frontend/src/components/Modals/DetailsButtonModal.tsx
```diff
@@ -47,6 +49,9 @@ export const DetailsButtonModal = ({
     <Modal show={show} onHide={handleClose} fullscreen={true}>
       <Modal.Header closeButton>
         <Modal.Title>
+          <small className="text-muted ps-2">
+            JSON Schema test cases that need attention : &nbsp;
```
Julian, line 53: What's with the nonbreaking spaces? A few of your PRs have had them.

Julian, line 53:
```suggestion
            Unsuccessful Tests:
```

frontend/src/components/Modals/DetailsButtonModal.tsx
```diff
@@ -69,21 +74,35 @@ const DetailItem = ({
...
 }) => {
+  const schemaFormatted = JSON.stringify(schema, null, 2);
+  const instanceFormatted = JSON.stringify(instance, null, 2);
   return (
     <div className="col">
       <div className={`card mb-3 ${borderClass}`}>
         <div className="card-body">
-          <h5 className="card-title">{title}</h5>
-          <p className="card-text">{description}</p>
+          <h5 className="card-title">Case : {title}</h5>
```
Julian, line 94: These should probably use `<label>`, but I don't really want to spend much time here -- I wouldn't really have been comfortable merging this PR as-is rather than simply doing the work to use the schema/instance widget which I thought I was clear about in https://github.com/bowtie-json-schema/bowtie/issues/871#issuecomment-1954413824 -- I said:

> I think at very least we should reuse/improve the schema/instance widget

I'll try to be even clearer next time.

Please fix the minor inconsistencies (which I'll leave suggestions on in a second) and then I'll merge this one time, but really this is a bit too underpolished for comfort -- the existing UI is pretty unpolished but I prefer not churning.

Julian, line 94:
```suggestion
          <h5 className="card-title">Case: {title}</h5>
```

frontend/src/components/Modals/DetailsButtonModal.tsx
```diff
@@ -69,21 +74,35 @@ const DetailItem = ({
...
+          <p className="card-text">Test : {description}</p>
```
Julian, line 95:
```suggestion
          <p className="card-text">Test: {description}</p>
```

frontend/src/components/Modals/DetailsButtonModal.tsx
```diff
@@ -69,21 +74,35 @@ const DetailItem = ({
...
+          <div className="card-body">
+            <pre id="schema-code">Schema: {schemaFormatted}</pre>
+          </div>
+          <div className="card-body">
+            <pre id="schema-code">Instance: {instanceFormatted}</pre>
+          </div>
+        </div>
+        <div className="card-footer text-muted text-center">
+          Result : {message}
```
Julian, line 104:
```suggestion
          Result: {message}
```

frontend/src/components/Modals/DetailsButtonModal.tsx
Siddharth-Singh-2004: I use &nbsp to have a space between text and value, when there is text preceding a value. I don't want the value to stick to the colon, like :Json-Java-Schema. Instead, I would like to have it show like : Json-Java-Schema.

frontend/src/components/Modals/DetailsButtonModal.tsx
Siddharth-Singh-2004: I approached this issue as a bug fix doing only as much as is needed to bringing in new data without changing the surrounding code. It was tagged as good-first-issue and I assumed it to mean no refactoring or no changing more than what is needed. Now that I spent a few days working with the community and mentors, I understand a bit more about how to work with  feedback and comments.

frontend/src/components/Modals/DetailsButtonModal.tsx
Siddharth-Singh-2004: The changed code needs to be &nbsp ; follow Unsuccessful Tests: 

Before the inclusion of &nbsp ;

Screenshot 2024-02-21 at 10 44 01 AM

After  the inclusion of &nbsp ;

Screenshot 2024-02-21 at 10 44 33 AM


Note the space between : and Jsonschema

frontend/src/components/Modals/DetailsButtonModal.tsx
Julian: Note that it was tagged [enhancement](https://github.com/bowtie-json-schema/bowtie/issues/871).

"Don't change more than what's needed" is always a good rule (for any kind of change), but in general for enhancements that means "don't invent a halfway thing unless it's clearly better, and definitely not if we immediately switch to the real one right after, in that case, just do the real work".

frontend/src/components/Modals/DetailsButtonModal.tsx
Julian: A nonbreaking space is very weird to use for that -- I'm no frontend developer so I don't know whether there are any norms there -- but you sound like you just want "a space". The nonbreaking behavior has nothing to do with anything, and worse is probably wrong -- if need be we probably *would* want to break at that character.

If having a space isn't working that (to me, again not a frontend expert) sounds like a CSS bug, since one (i.e. exactly one) should be being preserved -- presumably using `<label>` or some other proper tag to mark the label would have fixed this as well.

Siddharth-Singh-2004: I certainly can use labels with padding properties. The project has not used label tag on any screen so far and I tried to keep my fixes in synch with the existing coding style. I will in the future with your permission do what is right for my fixes even if it looks slightly out of place. Sorry about the back and forth.

Siddharth-Singh-2004: I certainly can use labels with padding properties. The project has not used label tag on any screen so far and I tried to keep my fixes in synch with the existing coding style. I will in the future with your permission do what is right for my fixes even if it looks slightly out of place. Sorry about the back and forth.

Siddharth-Singh-2004: I'll include labels and check in code.

frontend/src/components/Modals/DetailsButtonModal.tsx
Siddharth-Singh-2004: Thanks for the feedback.

Julian:
Nice! This looks mergeable -- one further improvement that I don't really understand is that I see schemas having quite a bit of padding on the right, and instances then get squished.

Not sure what's causing that, but might be a good next thing to look at. Here's what I see:

Screenshot 2024-02-22 at 8 57 21 AM

where the left sides obviously have lots more room than the right.

PR #911, opened 2024-02-20
`--max-fail` and `--max-error` on `bowtie suite` and `bowtie run` commands

### LINKED ISSUE #3: Add --max-fail / --max-error / --max-disagreement
(For stopping short after the corresponding situations).

adwait-godbole:
Fixes #3 

@Julian plz have a look and let me know.

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1027,11 +1027,9 @@ async def _run(
                         # Stop after this case, since we still have futures out
                         should_stop = unsucessful.causes_stop
                     else:
-                        if (max_fail is not None and
-                            unsucessful.failed >= max_fail):
+                        if max_fail and unsucessful.failed == max_fail:
                             should_stop = True
-                        if (max_error is not None and
-                            unsucessful.errored >= max_error):
+                        if max_error and unsucessful.errored == max_error:
                             should_stop = True
 
                 if should_stop:
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -658,6 +658,18 @@ FAIL_FAST = click.option(
     default=False,
     help="Fail immediately after the first error or disagreement.",
 )
+MAX_FAIL = click.option(
+    "-mf",
+    "--max-fail",
+    type=click.IntRange(min=1),
+    help="Fail immediately if N tests fail in total across implementations"
+)
+MAX_ERROR = click.option(
+    "-me",
+    "--max-error",
+    type=click.IntRange(min=1),
+    help="Fail immediately if N errors occur in total across implementations"
+)
 SET_SCHEMA = click.option(
     "--set-schema",
     "-S",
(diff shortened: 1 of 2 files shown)


bowtie/_cli.py
```diff
@@ -658,6 +658,18 @@ def _do_nothing(*args: Any, **kwargs: Any) -> CaseTransform:
     default=False,
     help="Fail immediately after the first error or disagreement.",
 )
+MAX_FAIL = click.option(
```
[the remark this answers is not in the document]
adwait-godbole: @Julian . Can you have a look once again now. Let me know.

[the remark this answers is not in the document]
adwait-godbole: Hey @Julian . Can you have a look once again now and let me know.

bowtie/_cli.py
```diff
@@ -656,8 +690,23 @@ def _do_nothing(*args: Any, **kwargs: Any) -> CaseTransform:
...
+    "-mf",
+    "--max-fail",
+    type=click.IntRange(min=1),
+    callback=_check_fail_fast_provided,
+    help="Fail immediately if N tests fail in total across implementations",
+)
+MAX_ERROR = click.option(
+    "-me",
```
Julian, line 704: I'd be surprised if this works -- short options usually have to be one letter, so for most CLIs `-ab` should be equivalent to `-a -b` -- if this does work, it certainly would break if we added a `-m` option (or a `-f` one) I suspect.

Anyways, I think this probably is rare enough that it's fine to only have the long option, you can just remove these.

adwait-godbole: Will do. Any other changes ?

Julian: The rest I think looks good to me!

adwait-godbole: I had a doubt though, currently running `bowtie suite -i lua-jsonschema https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft7/type.json --max-fail 2` command also outputs `"did_fail_fast": true` in the output json file. Is that correct or do we explicitly keep it only for --fail-fast ?

Julian: Good question! No we should also call that failing fast! That marker should signal to someone "the report might be truncated -- there were more cases, but we stopped because we reached a maximum number of failing tests" -- if you see a good place to either document that more explicitly or ensure it with a test that would be good too!

adwait-godbole: Do you mean that I should just keep it as `"failing_fast": true` for both --fail-fast and --max-fail 2  😅?

Julian: We should output `did_fail_fast: true` any time `--max-fail`, `--max-error` *or* `--fail-fast` ended up being triggered (i.e. we exceeded their number).

And no need to rename the field I think if that's what you're asking?

adwait-godbole: I removed the short flags as you suggested. Its mergable now!

Julian:
Thanks again!

PR #914, opened 2024-02-21
Fix possibility of misleading terminal output when smoke testing multiple implementations

### LINKED ISSUE #913: Possibility of wrong exit code message when smoke testing multiple implementations
Identified a bug in which when we smoke test multiple implementations, if some prior implementation gives an error then for all next implementations: `❌ some failures` would be printed since the `exit_code` variable isnt being reinitialized to 0 after each implementation.

```
    exit_code = 0

    for implementation in implementations:
        echo(f"Testing {implementation.name!r}...\n", file=sys.stderr)
        serializable: list[dict[str, Any]] = []

        async for _, results in implementation.smoke():
            async for case, result in results:
                if result.unsuccessful():
                    exit_code |= _EX_DATAERR

                match format:
                    case "json":
                        serializable.append(
                            dict(
                                case=case.without_expected_results(),
                                result=asdict(result.result),
                            ),
                        )

                    case "pretty":
                        echo(f"  · {case.description}: {result.dots()}")

                    case "markdown":
                        echo(f"* {case.description}: {result.dots()}")

        match format:
            case "json":
                echo(json.dumps(serializable, indent=2))

            case "pretty":
                message = "❌ some failures" if exit_code else "✅ all passed"
                echo(f"\n{message}", file=sys.stderr)

            case "markdown":
                message = (
                    "**❌ some failures**"
                    if exit_code
                    else "**✅ all passed**"
                )
                echo(f"\n{message}", file=sys.stderr)

    return exit_code
```

sudo-jarvis:
Closes #913

Fixed the bug where wrong terminal output stating `❌ some failures` would be printed for all implementations irrespective of their results while smoke testing multiple implementations after one fails.

----
📚 Documentation preview 📚: https://bowtie-json-schema--914.org.readthedocs.build/en/914/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -847,11 +847,12 @@ async def smoke(
     for implementation in implementations:
         echo(f"Testing {implementation.name!r}...\n", file=sys.stderr)
         serializable: list[dict[str, Any]] = []
-
+        implementation_exit_code = 0
+        
         async for _, results in implementation.smoke():
             async for case, result in results:
                 if result.unsuccessful():
-                    exit_code |= _EX_DATAERR
+                    implementation_exit_code |= _EX_DATAERR
 
                 match format:
                     case "json":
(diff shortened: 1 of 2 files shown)


Julian:
Thanks, can you add a test?

sudo-jarvis:
I am trying to write a test using bowtie smoke for two implementations - envsonschema which is a fake implementation and python json schema.... so one must return failures while one must suceed. However, on running the smoke test multiple times the order of results is changing. I am not sure why this is happening. Should this be happening @Julian ?

Julian:
Don't use a real implementation for tests, if you need a second one there are some string implementations higher up in the file. That might not help for the order thing, if the order is changing possibly some things are being put in a set, which makes the order arbitrary, in which case just asserting against either order is one easy way around it.

sudo-jarvis:
Any fake implementation in your information @Julian that returns all passed when smoke tested? Tried always_valid but even that didnt return all passed. Will need 1 implementation returning all passed to test this.

sudo-jarvis:
@Julian , added the test using a new `passes_smoke` shell implementation.

Julian:
Great! I just meant to use `printf` like the other implementations there, no need for the full "power" of awk -- but merging, made that small tweak. Thanks again!

PR #920, opened 2024-02-22
`bowtie filter-implementations` support

### LINKED ISSUE #191: Add `bowtie filter-implementations`
Running:

`bowtie filter-implementations` should filter the list of supported implementations by some provided option.

As a first such option, `bowtie filter-implementations --dialect http://json-schema.org/draft-07/schema#` should produce the names of all implementations supporting the Draft 7 dialect (or obviously any other provided dialect ID).

`--language foo` is a second "obvious" option for filtering out implementations supporting a given language.

#24 is related (though this command still needs to solve "what is the list of *all* implementations if that list isn't provided on the CLI).

adwait-godbole:
Fixes #191 

Screenshot (50)

Screenshot (51)


bowtie/_cli.py
```diff
@@ -49,6 +49,8 @@
     from bowtie._commands import AnyTestResult
     from bowtie._core import DialectRunner, ImplementationInfo, MakeValidator
 
+IMPLEMENTATIONS = Path(__file__).parent.parent / "implementations"
```
Julian, line 52: This is a good thought, but won't work -- in particular it relies on having a checkout of Bowtie present! It won't work once Bowtie is installed. It's unfortunate CI doesn't tell you that, the reason for that is that it obviously runs with a checkout present. Really we should have a separate job that would tell you that this would all be very broken if someone ran `pip install bowtie-json-schema` and didn't also have a checkout.

The way I'd hoped to do this was to use the GitHub packages API to list known implementations -- but unfortunately, it requires authentication (meaning everyone who uses Bowtie would need to login to  use it).

I'm still tied up with some other things so can't give more specific guidance quite yet, but if you want to tackle this issue (which also would be great!) the thing you need to do is learn how to have [hatchling](https://hatch.pypa.io/latest/) compile this data at package build time.

In other words -- when we build Bowtie as part of a release, at *that* time we should essentially compile a list of then-known implementations as a JSON file, and put that within the package.

Alternatively/in addition, we can make use of your `implementations.json` file as well once we have it -- but I think *both* having a local copy and then falling back to retrieve the remote one would be a good long term behavior.

adwait-godbole: Okay @Julian I got your reasoning. I will look into `hatchling` so we have the JSON file during bowtie build process and yes we certainly have `implementations.json` as well (once it is available) to fall back on so I will code that logic as well in case the build JSON file was absent for some reason.

adwait-godbole: Also if we manage to build this JSON file during `hatchling build` process do we still need to start all the individual Docker containers just to get to know the output of `bowtie filter-implementations` ? Seems quite expensive given that we are not running any tests and already have a JSON file from build.

Julian: We indeed wouldn't! We could put all we need in there.

adwait-godbole: Hey @Julian is it possible to start a bowtie subprocess like how we do in `test_integration.py` to run this command 
`bowtie info $(ls implementations/ | sed 's/^/-i /') --format json > implementations.json` (programmatically obviously)  in the `noxfile.py`'s below session function to guarantee having an `implementations.json` file in the packaged bowtie app? or do we have to force include the file in `pyproject.toml` ?

```python
@session(tags=["build"])
def build(session):
...
```

Julian: No the `noxfile` isn't the right place, nor is `pyproject.toml` as that will rely on you having a static copy of the file -- I believe the thing you want to look at is [this](https://hatch.pypa.io/1.9/config/build/#build-hooks) -- a build hook plugin which will do this work while hatch(ling) is building the package. That hook should *generate* that file and include it.

adwait-godbole: Okay @Julian I will look into it and revert back if I face some other query.

Julian:
OK. Looking back at this PR I remember why you were trying to have all the implementation data in there and not just the names -- I guess you were trying to speed up this filter.

That makes sense obviously, but I think there's other ways we could do that, or we could indeed go back and add more stuff to the build plugin, but I kind of feel icky doing it honestly.

I think as a first version we should be OK simply starting the containers and not pre-caching -- we can always cache the output to speed up a second run.

adwait-godbole:
> OK. Looking back at this PR I remember why you were trying to have all the implementation data in there and not just the names -- I guess you were trying to speed up this filter.

Indeed @Julian that's what I had in mind initially 😅 as running this command is quite an expensive task! Thanks for your changes on the PR #932 I have made the changes in here to use the `known()` method you wrote on `Implementation` class to get those names. 

So the command is now up and running! You can try it out! Just the thing is that it takes quite some time to get the results as it involves spinning off all the containers!.

adwait-godbole:
@Julian now that #877 is hashed out let me know your thoughts on this one!

bowtie/_cli.py
```diff
@@ -146,53 +146,55 @@ def implementation_subcommand(reporter: _report.Reporter = SILENT):
     """
 
     def wrapper(fn: ImplementationSubcommand):
-        async def run(
```
Julian, line 149: I don't understand immediately why you refactored this function. It seems like it should do exactly what you need, and your job is to just provide it with the right `image_names`, which is how I intended it to be used. Will have to try to understand why you changed it.

bowtie/_cli.py
```diff
@@ -146,53 +146,55 @@ def implementation_subcommand(reporter: _report.Reporter = SILENT):
...
-                    exit_code |= await fn(implementations=running, **kw) or 0
-                else:
-                    exit_code |= _EX_CONFIG
+        return cmd
+
+    return wrapper
+
 
-            return exit_code
+def all_implementations_subcommand(reporter: _report.Reporter = SILENT):
```
Julian, line 168: (And relatedly then why you introduced this function).

bowtie/_cli.py
```diff
@@ -714,19 +716,69 @@ def _check_fail_fast_provided(
...
+                "Possible shortnames include: "
+                f"{', '.join(sorted(Dialect.by_alias()))}."
+            ),
+        )
+
+    return wrapper
+
+
+DIALECT = create_dialect_option()
+LANGUAGE = click.option(
```
Julian, line 742: Please never hardcode things, that should set off red alarm bells.

Here a simple way to get what you need here presumably is to use `Implementation.known` and then grab the stuff before the hyphen, unless we think of an even better idea, but that seems pretty ok to me for now. You could also do some sort of aliasing similar to what we do in the frontend, so that someone can type `typescript` instead of `ts`, but yeah, no need to add something which will so easily go stale.

bowtie/_cli.py
```diff
@@ -880,7 +932,32 @@ def validate(
...
+async def filter_implementations(
+    implementations: Iterable[Implementation],
+    dialect: Dialect | None,
+    language: _L | None,
+):
+    supporting_implementations: list[str] = []
+    for each in implementations:
+        metadata = each.info.serializable()
+        impl_id = each.info.id
+        impl_name = impl_id.split("/")[-1]
```
Julian, line 950: This too should set off alarm bells that you're parsing something you probably shouldn't. I'm not sure why you're doing the parsing here, this will be completely wrong if someone provides a full image name -- in other words, we treat `foo-bar` as if someone wrote `ghcr.io/bowtie-json-schema/foo-bar`, but someone can provide `foo/bar/baz` and then we start the image from `foo/bar/baz` and make no assumptions about it, whereas you here would print `baz` and therefore not be able to find the image at all.

docs/spelling-wordlist.txt
```diff
@@ -25,3 +25,10 @@ subcommands
 # this "word" (tui) appears nowhere but oh god as usual who wants to debug Sphinx
 tui
 validator
+clojure
```
Julian, line 28: All of these should go away if you don't hardcode the language names presumably.

Julian (commented):
I still haven't gotten to looking at this fully, but I left a few initial comments.

Julian:
This also seems like it should be testable -- specifically, if you make it take `-i` implementation options, and default that to "all known implementations" (kind of like what we want to do in the future for all commands) then in your test case you can test this by using `-i envsonschema -i whateverothertestingharness` and make sure it filters those correctly.

bowtie/_cli.py
adwait-godbole: My reply on the other comment you gave will help you understand my side of why I introduced this function.

bowtie/_cli.py
adwait-godbole: Yes @Julian you are indeed correct but if you look closely in my code I am still providing the `run` function with the right image names as you pointed out. The only refactoring I did was to take that function(`run`) out of its `wrapper` function and rename it as `_run_implementations` and make it a separate one so that I can call it again inside the `all_implementations_subcommand` function and don't have to rewrite it in there. 

Now why I took it out separately is because in the `all_implementations_subcommand` function (which I introduced in my code), the `wrapper` function inside it isn't really expecting the `@IMPLEMENTATION` option /decorator on it unlike the `implementation_subcommand` function's `wrapper` function which does have that decorator on it.

 The user when using the `bowtie filter-implementations` command only provides either the `dialect` or the `language` as you had mentioned in the issue description and in this case even if we use `@IMPLEMENTATION` on the wrapper function it doesn't really make sense because we aren't expecting any implementations option from the user side and it will throw error because it is set as required = True. When the user hits this command we simply start off all the containers (which is why its slow) and filter out based on either the provided dialect or the language or both.
I also thought of doing `@IMPLEMENTATION(required=False)` but then when the user does `bowtie filter-implementations --help` he / she will still see that the command expects image names.

bowtie/_cli.py
adwait-godbole: Will do these changes. Thanks!

bowtie/_cli.py
adwait-godbole: @Julian, as explained by me in the reply to the first comment, the user here isn't really providing the image names.  If you look closely, the `impl_id.split("/")[-1]` I am doing is on the `each.info.id` which is already the full path of `ghcr.io` included image name.

docs/spelling-wordlist.txt
adwait-godbole: Yes! removing this

adwait-godbole:
> This also seems like it should be testable -- specifically, if you make it take `-i` implementation options, and default that to "all known implementations" (kind of like what we want to do in the future for all commands) then in your test case you can test this by using `-i envsonschema -i whateverothertestingharness` and make sure it filters those correctly.

@Julian are we expecting this command to take the `-i` option ? If so then all my above replies make no sense. In the issue description, the command was suppose to filter out implementations based on `dialect` or `language` or both right ? So what exactly is the behaviour here when someone provides the `-i` option ? What are we filtering on in that case and then what's the purpose of `dialect` and / or `language` ?

adwait-godbole:
@Julian I have removed the hard coded langs in the latest commit.

bowtie/_cli.py
Julian: No need to be sorry! You definitely don't *offend* or anything :D, I'm mentioning these things in the hope they're useful to learn! But apologies are never needed for not knowing something!

Julian:
> @Julian are we expecting this command to take the -i option ? There was no explicit mention of it in the issue description. Did I miss something over there ? 

I'll "apologize" here myself, you didn't miss anything explicit! -- as you can tell, sometimes the right design for a feature isn't obvious until you start, or even until you see code changes! So yes I'm sure the issues aren't as clear as they can be (you'll extra forgive me as I'm one person trying to spec these out, on a normal product we'd have product managers and designers and et cetera to help share the load on making sure issues are clearly written :), but alas!).

So, for the actual question: I didn't expect to need it before, but I think it does make sense (as a sidebar, there's a philosophy in testing which I believe is attributed to Michael Feathers but I forget exactly who champions it so I may be misremembering -- but it basically is that anytime you notice something is hard to test, you should consider whether that means it's missing some functionality or has the wrong design -- in other words, being hard to test is a good indicator for being deficient in some *other* non-testing related way. So here for example, we'd have difficulty testing this command, so we might be missing something!)

Now all that philosophizing aside, the thing that might be useful is for someone to be able to filter a *provided* set of implementations. Here's a fuller idea of a design:

Print out all Python implementations:

```
bowtie filter-implementations --language python
```

which should print:

```
ghcr.io/bowtie-json-schema/python-fastjsonschema
ghcr.io/bowtie-json-schema/python-jsonschema
ghcr.io/bowtie-json-schema/python-jschon
```

Print out all Python implementations *from the specified ones on the command line*:

```
bowtie filter-implementations -i python-jsonschema -i python-jschon -i js-hyperjump
```

which should print:

```
ghcr.io/bowtie-json-schema/python-jsonschema
ghcr.io/bowtie-json-schema/python-jschon
```

We may as well also do stdin, so:

```
ls implementations/ | bowtie filter-implementations --language python
```

should print

```
ghcr.io/bowtie-json-schema/python-fastjsonschema
ghcr.io/bowtie-json-schema/python-jsonschema
ghcr.io/bowtie-json-schema/python-jschon
```

(note: implementing that will mean using `sys.stdin.istty` to differentiate -- if it's a tty, you use `Implementation.known` as your default, otherwise you read stdin and use that as your implementations as if they were provided via `-i`).

Hopefully the above makes some sense (and hopefully you see `--dialect` should work similar -- though I don't think we should call it `--dialect` actually -- it has a very different meaning from the `--dialect` option to other commands. There it means "please speak this dialect". Here it means "support a given dialect". So I think we actually should call the option `--supports-dialect / -D for short option`. You should allow it to be repeated multiple times, so `-D 7 -D2020` should mean "supports *both* draft 7 and 2020".

Ask if the above looks unclear! And on the code side hopefully this will mean *less* refactoring, so if it doesn't seem that way ask!

adwait-godbole:
Thanks @Julian for giving me this clear picture now! I should've just asked about the exact details on the issue itself before directly raising a PR. But yeah, learnings for the future!

adwait-godbole:
Hey @Julian I tried to achieve all the requirements you specified above have a look at these two screenshots:

Screenshot (56)

Screenshot (57)

The refactoring part was still required and I haven't really done much refactoring if you look closely I have only renamed the `run` function to `_run_implementations` and extracted it out so I can also call it in `all_implementations_subcommand()`. The reason for refactoring is `implementation_subcommand()` is also used by the `bowtie info` and some other commands as well so trying to default there to all known implementations if no image names are provided didn't make sense and later having `all_implementation_subcommand()` did turn out to be useful to also have checks for the `sys.stdin.tty` logic.

I haven't added the test yet. Will do once all these changes are approved by you :)

Julian:
I haven't fully gotten through looking back at the comments here. I meant to leave some notes on what I was after, I'll perhaps try to leave a few in a bit -- but I did "revert" a bunch of the refactoring here to try to show what I meant. Have a look, and tell me if something really doesn't make sense. It's certainly possible I misled or have missed something! But what I pushed at least seems to have the desired behavior.

Personally I always recommend you write your tests either first or else immediately after (which one is personal preference, but certainly before I review, you should have some -- not having them makes it harder to refactor!). Adding some now for the various cases is likely the right immediate step besides the above.

(Thanks as usual, you're definitely doing good work here!)

Julian:
Oh one general thing I forgot to mention and also fix, but which *does* point to some needed refactoring is that a reason this is slow is that unlike some other commands, here we don't want to *keep* the implementations running, so it's a waste to wait for all them to start, we instead want to immediately do our query once each one has started up. But what I was trying to stress earlier about the refactoring business is that there's still some in progress, so doing more is very careful work essentially to make sure it handles all the cases.

adwait-godbole:
Thanks @Julian for the feedbacks and your changes in the code. The changes definitely make it more simpler than my longer approach of refactoring and unnecessary click option wrapping.

> Personally I always recommend you write your tests either first or else immediately after (which one is personal preference, but certainly before I review, you should have some -- not having them makes it harder to refactor!). Adding some now for the various cases is likely the right immediate step besides the above.

I will add the tests asap!

> Oh one general thing I forgot to mention and also fix, but which _does_ point to some needed refactoring is that a reason this is slow is that unlike some other commands, here we don't want to _keep_ the implementations running, so it's a waste to wait for all them to start, we instead want to immediately do our query once each one has started up. But what I was trying to stress earlier about the refactoring business is that there's still some in progress, so doing more is very careful work essentially to make sure it handles all the cases.

So if I understand this correctly do you mean we want something like

start one container -> query it -> stop it -> start another container -> query it -> stop it -> …..

And do this until all specified containers / known implementations (if no impls were mentioned) were queried ?

adwait-godbole:
> Personally I always recommend you write your tests either first or else immediately after (which one is personal preference, but certainly before I review, you should have some -- not having them makes it harder to refactor!). Adding some now for the various cases is likely the right immediate step besides the above.

I will add the tests asap!

> Oh one general thing I forgot to mention and also fix, but which _does_ point to some needed refactoring is that a reason this is slow is that unlike some other commands, here we don't want to _keep_ the implementations running, so it's a waste to wait for all them to start, we instead want to immediately do our query once each one has started up. But what I was trying to stress earlier about the refactoring business is that there's still some in progress, so doing more is very careful work essentially to make sure it handles all the cases.

The command is only slow when no implementations are provided in the command because that is the only case we default to **all known implementations** which makes it slow. So if I understand this correctly, do you mean we want something like

start one container -> query it -> stop it -> start another container -> query it -> stop it -> …..

And do this until all the known implementations were queried ? If so then won't it take more or less the same amount of time for the command to complete as if we were starting all the containers at once and then querying them 😅? Is it like a trade-off between 2 for loops vs 1 for loop ? Just an initial thought.

Julian:
> However your changes in the below block of code are not creating the full IMAGE_REPOSITORY url (the one with ghcr.io) when no implementations are passed in the command😅. It does work when implementations are passed.

Yeah I know I originally showed examples that contained it but I think it's fine if we leave it out. The thing is we want something that will work if you pass it to `bowtie validate -i <that>`. And either way works -- the reason I left it out is just because the user is going to be more used to typing `bowtie validate -i lua-jsonschema` than `bowtie validate -i ghcr.io/bowtie-json-schema/lua-jsonschema`, even though both work.

>  If so then won't it take more or less the same amount of time for the command to complete as if we were starting all the containers at once and then querying them 

If you wait for each one then yes! But the point is you want to start *all* of them immediately, query all of them as soon as they are ready, etc. -- but right now I believe (but haven't fully verified) that we do start all at once, but we wait for *all* to be started up before we start querying the first one, when we should essentially be moving "as fast as possible" independently.

adwait-godbole:
> Yeah I know I originally showed examples that contained it but I think it's fine if we leave it out. The thing is we want something that will work if you pass it to `bowtie validate -i <that>`. And either way works -- the reason I left it out is just because the user is going to be more used to typing `bowtie validate -i lua-jsonschema` than `bowtie validate -i ghcr.io/bowtie-json-schema/lua-jsonschema`, even though both work.

Ohh! I think I didn't frame it correctly. What I meant was running  `bowtie filter-implementations --supports-language python` is not working since your changes. It is not getting the correct image url. However passing implementations is working!

adwait-godbole:
Also @Julian I added the tests for cases where implementations are given but wasn't sure how to write the tests for known implementations ? One approach I thought was of doing this:

```python
@pytest.mark.asyncio
async def test_filter_known_implementations(tmp_path):
    implementations_dir = tmp_path / "data"
    implementations_dir.mkdir()
    
    known_impls = implementations_dir.joinpath("known_implementations.json")
    known_impls.write_text(
        "["python-jsonschema","go-jsonschema","js-jsonschema"]"
    )

    stdout, stderr = await bowtie(
        "filter-implementations",
        "--supports-language",
        "python",
    )
    output = {
        "ghcr.io/bowtie-json-schema/python-jsonschema",
    }
    assert set(stdout.splitlines()) == output
    assert stderr == ""
```
Is this the correct way ?

Julian:
> Ohh! I think I didn't frame it correctly. What I meant was running bowtie filter-implementations --supports-language python is not working since your changes.

Isn't it? Here I get:

```
⊙  bowtie filter-implementations --supports-language python                                                                                                                                                                                                                                                       julian@Airm ●
python-jschon
python-fastjsonschema
python-jsonschema
```

> Is this the correct way ?

It's tricky actually. What you did there is one way (basically stub out the data). But the thing is we don't actually want to spin up real implementations in the test suite (which is what would happen if that test runs). Doing so makes the test suite a lot slower and also flakier when it fails, which could be for many unrelated reasons. I think for now it's fine to not test that case actually. We can come back later to test it if/when we have an even better set of testing helpers. But good that you thought of it and tried.

tests/test_integration.py
```diff
@@ -1249,6 +1256,83 @@ async def test_info_unsuccessful_start(succeed_immediately):
...
+        "-i",
+        fakejsimpl,
+        "--supports-language",
+        "python",
+    )
+    output = {
+        "bowtie-integration-tests/envsonschema",
+        "bowtie-integration-tests/lintsonschema",
+    }
+    assert set(stdout.splitlines()) == output
```
Julian, line 1280: Don't use `set`, that could hide some strange bug where we print things out twice. Use `sorted` if you're just trying to avoid ordering problems.

tests/test_integration.py
```diff
@@ -1249,6 +1256,83 @@ async def test_info_unsuccessful_start(succeed_immediately):
...
+        envsonschema,
+        "-i",
+        lintsonschema,
+        "-i",
+        fakejsimpl,
+        "--supports-language",
+        "python",
+    )
+    output = {
+        "bowtie-integration-tests/envsonschema",
```
Julian, line 1277: Use `tag("envsonschema")` (and everywhere else here) to save a bit of duplication.

Julian:
> Shouldn't the command output the entire ghcr.io included implementation name / id like this ghcr.io/bowtie-json-schema/python-fastjsonschema as in your earlier guidance in the comments you said about this output format right ?

That's what I was responding to with

> Yeah I know I originally showed examples that contained it but I think it's fine if we leave it out. The thing is we want something that will work if you pass it to bowtie validate -i <that>. And either way works -- the reason I left it out is just because the user is going to be more used to typing bowtie validate -i lua-jsonschema than bowtie validate -i ghcr.io/bowtie-json-schema/lua-jsonschema, even though both work.

The two are equivalent, so I think I simply changed my mind on which one to output :) -- when you pass `python-jsonschema` to `bowtie validate -i python-jsonschema` it gets automatically interpreted as `bowtie validate -i ghcr.io/bowtie-json-schema/python-jsonschema`.

adwait-godbole:
Umm well, I just tried once again by copy pasting the entire `_cli.py` file on Codespaces but it throws error.

Julian:
What did you try and what happened?

adwait-godbole:
I tried `bowtie filter-implementations --supports-language python` and got this set of errors. I think all of them are 400 DockerError 400

Screenshot (61)

adwait-godbole:
> Don't use `set`, that could hide some strange bug where we print things out twice. Use `sorted` if you're just trying to avoid ordering problems.

Resolved!

> Use `tag("envsonschema")` (and everywhere else here) to save a bit of duplication.

Resolved!

Julian:
> PS. @Julian I have copied the _cli.py file from the last commit you made on this PR and none of my changes on that (there aren't any per say)

Wait as in you get those errors if you copy paste the file rather than simply checkout the branch? I mean I wouldn't be surprised :) -- just checkout the branch? The tests seem to pass and it works locally. Also make sure you wait till codespaces actually pulls the images and that that isn't happening in the background? I'm not sure we avoid that race condition in our codespaces setup, you can check with `ps aux | grep pull` maybe.

adwait-godbole:
No but even on local I am already on this branch and have pulled all your changes on my machine (the branch is up to date on my local with the remote) still the same error is getting thrown on my machine as well :(. Extremely sorry for troubling you with the same thing again and again but I have confirmed multiple times that I am on this branch and locally as well its up to date. Also the tests are passing because we aren't really testing the `all known implementations` logic :(

adwait-godbole:
@Julian I recorded my local machine :) Please have a look


https://github.com/bowtie-json-schema/bowtie/assets/68469605/10899e28-3279-4391-8fd8-b4971f412fec

Julian:
I'll have a look in a bit. But if you run `bowtie validate -i python-jsonschema <(printf '{}') <(printf 37)` that works?

adwait-godbole:
> bowtie validate -i python-jsonschema <(printf '{}') <(printf 37

Yes that works because you have explicitly passed `-i` unlike in `bowtie filter-implementations --supports-language python` where we don't pass `-i` so the `type=_Image()` is not getting triggered is what my guess is.

Julian:
Ah, I see, thanks! I was misunderstanding what was working and what wasn't. OK, yeah, that should be easy to fix.

Julian:
> Perhaps we could try it in another one ?

Yes definitely definitely! I don't even mean to / want to do so right now, there's a small chance we'll eventually change our async IO framework entirely sometime.

adwait-godbole:
Can we merge this then ? Or do you want to do some additional changes ?

Julian:
I want to be able to reproduce your error -- you changed the behavior (to now print the `ghcr.io` fully qualified names), which I mentioned I decided we don't want to do, it's unnecessary and at this point not something end-users see so it's less confusing to not show it.

But given everything works here I can't tell yet what part of your fix is necessary yet. I guess I'll have to try in codespaces, because yeah, everything works here.

adwait-godbole:
Yeah sorry for this trouble but it really isn't working for me when `-i` is not passed neither on my local nor in Codespaces so I had to do those changes.  I think you should be able to see that error in Codespaces

adwait-godbole:
> I want to be able to reproduce your error -- you changed the behavior (to now print the `ghcr.io` fully qualified names), which I mentioned I decided we don't want to do, it's unnecessary and at this point not something end-users see so it's less confusing to not show it.

But @Julian I think even without the changes I did, the command still outputs the `ghcr.io` fully qualified names both on my local and in Codespaces!

Julian:
OK great, definitely can reproduce it now! Will make a small tweak and then we should be good.

Julian:
OK, should be good to go now, if CI passes will merge.

adwait-godbole:
Thanks @Julian! I know I hardly contributed to this one :( but learnt a lot specifically thanks for that simple `issuperset` hack!

Julian:
That's wrong!!
Contribution is moving the project forward, not lines of code!
Without you the PR wouldn't have been merged!

Issue #544, opened 2023-10-11
Long schemas in the UI should not cause arbitrarily large Schema widgets, they should have a max-width + scrollbar


gregsdennis:
There are cases where it's very difficult or impossible to see the entire test scenario in the test suite report UI details.  A couple examples of the variations I see:

- **multiple dynamic paths to the $dynamicRef keyword** - This one has a huge schema, and (unless you have a portrait-oriented monitor) you can't see both the schema and the test cases.
- **email format** - This one just has a lot of test cases.  Scrolling down and clicking one of the cases populates the "instance" column, but you can't see that it's done anything because it's out of frame.

Julian:
Here (even on a landscape display) I see both of these fine, e.g. I see:

![Screenshot 2024-02-18 at 3 18 31 PM](https://github.com/bowtie-json-schema/bowtie/assets/329822/349ea1e9-b79a-4a86-8012-0a7014b6b42d)

and:

![Screenshot 2024-02-18 at 3 18 19 PM](https://github.com/bowtie-json-schema/bowtie/assets/329822/cd30a938-3c86-4e9e-a09b-f5094778c8f8)

For the second one we've now made clicking on tests automatically scroll back up to the instance widget, but for the second one I don't think anything's changed from when this was filed, and probably this is really about your screen resolution. So probably the fix for that is actually that the `Schema`  and `Instance` widgets should have min + max heights (one which at least shows them themselves along with some of the tests below) and then use a scrollbar to scroll long schemas.

Going to add this to `good first issue` in order to try to do ^ for anyone who picks this up.

gregsdennis:
@Julian in your first screenshot, you capture the entire schema, yes, but you can't also see the test cases.

PR #923, opened 2024-02-23
Add max height to schema and instance displays

SimonDMC:
Resolves https://github.com/bowtie-json-schema/bowtie/issues/544.

Here's how it looks:
image

I wasn't able to find a way to do this with just bootstrap so I added CSS to it. I'm not sure if it's better to keep it inline or extract it into a separate file (since the codebase seems to contain a mix of both) but it made sense to me to keep it inline for just one line of CSS. Happy to adjust the height if desired.

----
📚 Documentation preview 📚: https://bowtie-json-schema--923.org.readthedocs.build/en/923/

### DIFF
--- a/frontend/src/components/Cases/SchemaDisplay.tsx
+++ b/frontend/src/components/Cases/SchemaDisplay.tsx
@@ -1,5 +1,9 @@
 import CopyToClipboard from "../CopyToClipboard";
 
+const schemaStyle = {
+  maxHeight: "30em",
+};
+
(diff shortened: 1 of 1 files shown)


Julian:
This looks great, well done! Appreciated.

> I'm not sure if it's better to keep it inline or extract it into a separate file (since the codebase seems to contain a mix of both)

I'm not sure what typical style is myself! Certainly happy if we do trend towards one or the other though.

Issue #925, opened 2024-02-23
Fix the mobile UI to not horizontally scroll / bleed elements out of their containers


Julian:
The below screenshots are on Firefox Android, but I suspect reproduce across mobile browsers:

![Screenshot_20240223-091343](https://github.com/bowtie-json-schema/bowtie/assets/329822/9924ee72-7f09-4f3a-8e76-5ac3c22bf26f)
![Screenshot_20240223-091352](https://github.com/bowtie-json-schema/bowtie/assets/329822/b8e1b04a-77ed-4b4d-a673-9311d1ed20ee)

We should fix our components to render correctly on mobile browsers without leaking from their containers or horizontally scrolling the page.

skushagra9:
can I work on this

Julian:
Great, go for it!

PR #927, opened 2024-02-23
Fixed mobile responsive issue

skushagra9:
#925 
Screenshot from 2024-02-23 21-42-28
Screenshot from 2024-02-23 21-42-33

### DIFF
--- a/frontend/src/components/RunInfo/RunInfoSection.tsx
+++ b/frontend/src/components/RunInfo/RunInfoSection.tsx
@@ -50,7 +50,7 @@ const RunInfoSection = ({ runInfo }: { runInfo: RunInfo }) => {
                     return (
                       <tr key={label}>
                         <td>{label}</td>
-                        <td>
+                        <td style={{ maxWidth: "200px" }}>
                           <a href={value.href}>{value.text}</a>
                         </td>
                       </tr>
--- a/frontend/src/components/Summary/SummaryTable.tsx
+++ b/frontend/src/components/Summary/SummaryTable.tsx
@@ -5,6 +5,7 @@ import { ReportData, calculateTotals } from "../../data/parseReportData";
 const SummaryTable = ({ reportData }: { reportData: ReportData }) => {
   const totals = useMemo(() => calculateTotals(reportData), [reportData]);
   return (
+    <div className="table-responsive">
     <table className="table table-hover">
       <thead>
         <tr>
@@ -98,6 +99,7 @@ const SummaryTable = ({ reportData }: { reportData: ReportData }) => {
         </tr>
       </tfoot>
     </table>
+    </div>
   );
 };
 


Julian:
Thanks. Are those screenshots before or after the change? I presume obviously after.

Can you try to briefly explain what you've done and why when submitting changes?

In particular, here are two immediate things from your screenshots:

* it looks like you made labels be vertically justified to the bottom. Probably top makes more sense for this, as they're labels, and someone shouldn't see data before they know what they're lookinga t
* What did you do to fix the table? Does it horizontally scroll, or where are the extra columns?

skushagra9:
These are the pictures after the change
So , I made changes at two places 
1) commit link was overflown, image1- now its not flowing out anymore
2) and the summary table which was overflown

Both of them works perfectly fine,
and the scroll works horizontally and vertically both

frontend/src/components/RunInfo/RunInfoSection.tsx
```diff
@@ -50,7 +50,7 @@ const RunInfoSection = ({ runInfo }: { runInfo: RunInfo }) => {
                     return (
                       <tr key={label}>
                         <td>{label}</td>
-                        <td>
+                        <td style={{ maxWidth: "200px" }}>
```
Julian, line 53: I'm not sure pixels are the right unit to use here. Do you know that they are?

skushagra9: yes, I researched on it

skushagra9: you can try it once in the browser, by minimizing and maximinzing the window

skushagra9:
I have added the corrected images in the description

skushagra9:
@Julian , my latest commit made it more generic

Julian:
What is "it", what do you mean by "corrected images", corrected how?

@skushagra9 fast/sloppy work isn't what I want! Be slow, explain what you're doing clearly, and read what I'm asking you.

skushagra9:
I have added an commit , which fixes the maxwidh:200px to 50vw which is more generic
By what I meant from corrected images was , that I added the images before and after solving the issue

frontend/src/components/RunInfo/RunInfoSection.tsx
skushagra9: okay, I understand , what I did in my latest commit was to make it more generic, otherwise pixels also work ,
Pixels are a more absolute way of defining, what I now did in my latest commit is a reletive way

Julian:
@skushagra9 I'm giving you direct feedback that you're not doing yourself any favors here. I'm not trying to be hard on you, but responding immediately with a 2 sentence message in 2 minutes right after I tell you "be slow and careful" is not a good look. Your message makes very little sense, you just said "I researched it" on whether pixels are the right unit and then immediately change your mind when I ask -- so did you or did you not research what the right answer is? If you didn't clearly, then the right thing to do is to go re-read something to learn what the right thing to do is -- and then ideally share that so I learn it too.

You also haven't addressed what I asked about vertical justification -- that's more sloppiness, since even though I should have put that somewhere distinguishable, if you were reading slower and not rushing you would have remembered there was something else to change or respond to.

I'll have a look at your work in a bit and see about merging it, but please understand clearly that sloppiness and rushing are essentially the worst things you could be doing for your chances!

skushagra9:
Hey @Julian , I don't understand why did you got confused, As a token of respect I changed my approach , but there is no problem , This was a easy fix, so I made the fix quick

PR #928, opened 2024-02-23
Add codespace support

### LINKED ISSUE #919: Adding support for Github Codespaces
As discussed in #917, it would probably be good to have support for Github codespaces like we do for Gitpod workspaces. It would allow users to use bowtie on their virtual codespaces.

sudo-jarvis:
Closes #919 .

The goal was to add support for GitHub Codespaces in a similar manner to Gitpod support.

**Changes made:**

1. Added the Open in Github Codespace badge in README file.
2. Created the Dockerfile needed for Codespace config with python3.11 and docker.
3. Created the devcontainer.json file which installs bowtie after successful creating of the container. 
4. Configured the MOTD message similar to what is there in Gitpod so that it opens up by default on starting the codespace.

**First start screenshot:**
Screenshot 2024-02-23 at 10 30 24 PM

----
📚 Documentation preview 📚: https://bowtie-json-schema--928.org.readthedocs.build/en/928/


.devcontainer/.codespace.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.11
+
+# Install Docker CLI
+RUN apt-get update \
+  && apt-get install -y docker.io
```
Julian, line 5: Does podman also work? If so can we use that instead by default here?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "Bowtie",
+  
+  "dockerFile": ".codespace.Dockerfile",
+
+  "runArgs": ["-v", "/var/run/docker.sock:/var/run/docker.sock"],
+
+  "onCreateCommand": "sudo chown -R vscode /var/run/docker.sock; pip install -r requirements.txt -e .",
+  
+  "postCreateCommand": "for each in $(ls implementations); do docker pull ghcr.io/bowtie-json-schema/$each; sudo cp /workspaces/bowtie/docs/gitpod-motd.txt /usr/local/etc/vscode-dev-containers/first-run-notice.txt; done;"
```
Julian, line 10:
```suggestion
  "postCreateCommand": "for each in $(ls implementations); do docker pull "ghcr.io/bowtie-json-schema/$each:latest"; sudo cp /workspaces/bowtie/docs/gitpod-motd.txt /usr/local/etc/vscode-dev-containers/first-run-notice.txt; done;"
```

Also probably you can rename `gitpod-motd` to just be `motd` or something more generic, so that it's clear it's used for both.

Julian (commented):
Nice! Two minor comments.

.devcontainer/.codespace.Dockerfile
sudo-jarvis: But would the cli work with podman ? The cli too requires docker to be present right ?

sudo-jarvis: Oh! I am not very well aware of podman, I'll learn about it and get back to you.

sudo-jarvis: @Julian , seems like its not that easy to run podman in codespace, facing lots of permission errors

sudo-jarvis: @Julian , just wanted to update you that I am trying this but initially was facing some permission issues in the codespace while running podman. Thats resolved now but somehow bowtie is still not able to run the implementations, they are failing to start.

Julian: Cool -- could also give up, it's not the biggest deal if we go with this as is certainly, I just asked in case it was an easy change!

sudo-jarvis: I'll try solving it once again today and if I am not able to do so then maybe we can merge this as is and if needed change it in the future

sudo-jarvis: @Julian , facing issue creating access to the podman socket inside the codespace. Ig we can merge it as it is for now.

Julian:
Sounds good! Thanks again.

Issue #926, opened 2024-02-23
Badges on implementation page


harrel56:
Bowtie badges are still kind of a hidden feature. Maybe if we displayed all the available badges on each implementation page it'd be more clear what is actually accessible. Anyone could just copy the url/link to such badge and use it on his own.

So each implementation would get:
- [![Supported spec](https://img.shields.io/endpoint?url=https%3A%2F%2Fbowtie-json-schema.github.io%2Fbowtie%2Fbadges%2Fjava-dev.harrel.json-schema%2Fsupported_versions.json)](https://img.shields.io/endpoint?url=https%3A%2F%2Fbowtie-json-schema.github.io%2Fbowtie%2Fbadges%2Fjava-dev.harrel.json-schema%2Fsupported_versions.json)
- [![Compliance](https://img.shields.io/endpoint?url=https%3A%2F%2Fbowtie-json-schema.github.io%2Fbowtie%2Fbadges%2Fjava-dev.harrel.json-schema%2Fcompliance%2FDraft_2020-12.json)](https://img.shields.io/endpoint?url=https%3A%2F%2Fbowtie-json-schema.github.io%2Fbowtie%2Fbadges%2Fjava-dev.harrel.json-schema%2Fcompliance%2FDraft_2020-12.json) - and one of those for each dialect/draft version the implementation supports.

@Julian ofc let me know what you think

Julian:
This is kind of a duplicate of #211 but better written and a better suggestion :)

So yes I agree we shouldn't wait any longer and that putting this on the implementation page is a great idea!

Julian:
After #927 is finished, sure!

sd1p:
Hi, I am new to this project. I need some guidance to work on this issue. After reading the issue description carefully, I tried to implement a prototype for the required changes in the UI. Please provide a feedback if I am understanding issue correctly. 

 
![image](https://github.com/bowtie-json-schema/bowtie/assets/43758514/72e78ac7-f665-448c-926b-d84d9eadd7c5)
![image](https://github.com/bowtie-json-schema/bowtie/assets/43758514/62dff64c-cc7b-4f41-ae25-a5fb06ba8dd8)

harrel56:
@sd1p The placement of the badges looks good but it's still missing the one with all supported dialects. 

[...]

PR #929, opened 2024-02-23
Added badges for implementation dialects

sd1p:
Implementing https://github.com/bowtie-json-schema/bowtie/issues/926
Added badges of size width-200px , height-18px for implementation dialects.

image

image

----
📚 Documentation preview 📚: https://bowtie-json-schema--929.org.readthedocs.build/en/929/


code-Gambler (changes requested):
You need to also add this label Draft Label to the supported Dialects Drafts below

sd1p:
> You need to also add this label Draft Label to the supported Dialects Drafts below

I will work on that right now.

sd1p:
Added badges for Supported Dialects under Compliance section. 
image
image

sd1p:
@code-Gambler  Please review the UI changes.
image

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -42,7 +42,18 @@ const DialectCompliance: React.FC<{
               .map(([dialectName, result], index) => {
                 return (
                   <tr key={index}>
-                    <td>{Dialect.forPath(dialectName).uri}</td>
+                    <td>
+                      {Dialect.forPath(dialectName).uri}{" "}
+                      <img
+                        width={"auto"}
```
harrel56, line 48: Is defining width & height necessary?

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -42,7 +42,18 @@ const DialectCompliance: React.FC<{
               .map(([dialectName, result], index) => {
                 return (
                   <tr key={index}>
-                    <td>{Dialect.forPath(dialectName).uri}</td>
+                    <td>
+                      {Dialect.forPath(dialectName).uri}{" "}
```
harrel56, line 46: Please don't use spaces for positioning purposes - this should be handled with CSS styles

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -42,7 +42,18 @@ const DialectCompliance: React.FC<{
...
+                      <img
```
harrel56, line 47: Just my opinion: I feel like badges always appear as clickable links - maybe this can be a link to:

- badge image itself: https://img.shields.io/endpoint?url=https%3A%2F%2Fbowtie-json-schema.github.io%2Fbowtie%2Fbadges%2Fjava-dev.harrel.json-schema%2Fsupported_versions.json
- badge data actually hosted by bowtie: https://bowtie-json-schema.github.io/bowtie/badges/java-dev.harrel.json-schema/supported_versions.json

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -42,7 +42,18 @@ const DialectCompliance: React.FC<{
...
+                        width={"auto"}
+                        height={20}
+                        src={`https://img.shields.io/endpoint?url=https%3A%2F%2Fbowtie-json-schema.github.io%2Fbowtie%2Fbadges%2F${encodeURIComponent(
```
harrel56, line 50: I think logic for constructing this image source can be extracted to a function. Also maybe it would be more clear if you actually URL encoded whole query param - so first construct something like this *https://bowtie-json-schema.github.io/bowtie/badges/java-dev.harrel.json-schema/supported_versions.json* and then URL encode whole thing

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
```diff
@@ -42,7 +42,18 @@ const DialectCompliance: React.FC<{
...
+                          implementation.language,
+                        )}-${implementation.name}%2Fcompliance%2FDraft_${
+                          Dialect.forPath(dialectName).prettyName.split(" ")[1]
```
harrel56, line 53: Got to say that depending your logic on `prettyName` which, as the name suggests, is just for presentation purposes is a bad idea. All the dialects data can be found in file `dialects.json` - I'd say add appropriate value to each dialect, adjust `Dialect.ts`, and then use this value here. NGL this path for badges should be just different (probably include `dialect.shortName`)

harrel56 (commented):
Badge for supported dialects is presented multiple times - why? It's always the same information. I'd suggest following layout:
image

Where:
1. is a badge for supported dialects (on the left column, but below text OR above the dialect list in second column)
2. is a badge for compliance %

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
harrel56: Also, linter is complaining about lack of `alt` attribute - you may as well just add it

Julian:
> Badge for supported dialects is presented multiple times - why? It's always the same information. I'd suggest following layout: image
> 
> Where:
> 
>     1. is a badge for supported dialects (on the left column, but below text OR above the dialect list in second column)
> 
>     2. is a badge for compliance %

I like all of these ideas clearly -- just mentioning I think we could even consider removing the "Supported Dialects" *text* at the top, and *just* having the badge even.

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
Julian: Also the URL isn't the right one -- it should be https://bowtie.report -- but really, please use whatever way I presume vite/react has for getting the current base URI, which is even better, and will work even in the rare event we ever change domains.

harrel56:
> just mentioning I think we could even consider removing the "Supported Dialects" text at the top, and just having the badge even.

Totally agree

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
Julian: Linking somewhere to https://shields.io/badges/endpoint-badge seems also maybe like a decent idea, to send people to how they would customize the badge if they feel like doing so.

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
sd1p: I tried to remove height and width parameters and UI is working as intended.

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
sd1p: Using margin mx-1 instead of spaces.

sd1p:
Thanks for the constructive the feedback, I will work on the suggestions for now.

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
harrel56: Hmm, you're right - in `index.tsx` we compute the URI for report fetching:
```ts
const reportHost =
  import.meta.env.MODE === "development"
    ? "https://bowtie.report"
    : window.location.href;
const reportUri = new URI(reportHost).directory(import.meta.env.BASE_URL);
```
I guess this could be extracted to a separate file and exported, so it could be used anywhere in code. 

I believe this URI will work fine also for badges but I haven't given much thought to this so maybe it would need some modifications

sd1p: > I think logic for constructing this image source can be extracted to a function. Also maybe it would be more clear if you actually URL encoded whole query param - so first construct something like this _https://bowtie-json-schema.github.io/bowtie/badges/java-dev.harrel.json-schema/supported_versions.json_ and then URL encode whole thing
I used encodedURIComponent for implementation.language to encode "C++" string to URL encoded format `C%2B%2B`.
Also encoding the whole string is isn't helpful. 
image

frontend/src/components/ImplementationReportView/DialectCompliance.tsx
sd1p: Added `badgeName`  to each dialects. in `dialects.json` and adjusted Dialect.ts. 
`dialects.json` changes example-

```json
  {
    "$schema": "tag:bowtie.report,2024:models:dialect",
    "uri": "https://json-schema.org/draft/2020-12/schema",
    "shortName": "draft2020-12",
    "prettyName": "Draft 2020-12",
    "badgeName": "Draft_2020-12",
    "firstPublicationDate": "2020-01-28",
    "aliases": ["2020", "202012", "2020-12", "draft202012"]
  }

File: frontend/src/components/ImplementationReportView/DialectCompliance.tsx (hunk 4 above)
Comment (harrel56) [reply frontend/src/components/ImplementationReportView/DialectCompliance.tsx:50]: > Also encoding the whole string is not helpful.

What do you mean by that?

Comment (sd1p) [reply frontend/src/components/ImplementationReportView/DialectCompliance.tsx:50]: I implemented above logic by encoding the query parameter URL.  To make this logic modular where should we create the file to store the required function? Like the file path in the folder structure, something like libs or utils folder maybe.

```jsx                      
<img
    alt={`${Dialect.forPath(dialectName).badgeName}`}
    src={`https://img.shields.io/endpoint?url=${encodeURIComponent(
      `https://bowtie-json-schema.github.io/bowtie/badges/${
        implementation.language
      }-${implementation.name}/compliance/${
        Dialect.forPath(dialectName).badgeName
      }.json`
    )}`}
  />`

**Comment by sd1p** [comment]
I tried to make the implementation page more responsive for mobile view.  For now I have make the logic for constructing this image source modular using functions. 
bowtie

**Comment by sd1p** [comment]
For placing the badge for supported dialects I had 3 suggestions. Please guide on which suggestion to implement and I will push the respective changes.

Suggestion 1-
image

Suggestion 2-
image

Suggestion 3-
image

File: frontend/src/components/ImplementationReportView/DialectCompliance.tsx (hunk 5 above)
Comment (Julian) [reply frontend/src/components/ImplementationReportView/DialectCompliance.tsx:53]: I haven't looked closely but I think you've only changed this on the frontend -- that's going to break, since you haven't changed the path itself in the backend which generates the file, so you'll be looking in the wrong place.
Harrel's totally right that this is very wrong to use `prettyName`, so I agree 100% we should change it, but you have to change it in both places, and if/once you do I think you won't even need this new field, you can just make the badge name / path be `.shortName` itself and be done with it probably.

**Comment by Julian** [comment]
I still would suggest just removing all of the URLs you have there in the "Supported Dialects" box and just leaving the supported version badge and that's it, but let's see what @harrel56 thinks as well.

**Comment by harrel56** [comment]
Yeah, replacing all the URLs with badges seems fine :) I was also a little worried that suggestion 1 & 2 expand the header column a bit too much, but that's is a minor concern and I'm fine with any of the suggestions.

PS why the list of supported dialects is seemingly in a random order?

**Comment by Julian** [comment]
> PS why the list of supported dialects is seemingly in a random order?

The order is "descending order of compliance" right now, but I agree that's very non-obvious. We could make it more obvious by having an explicit column with that percentage presumably, though that's also somewhat related to #37 and generally showing what column we're sorting on in a table.

**Comment by harrel56** [comment]
IMHO it could just be sorted from newest to oldest - seems most intuitive and ordering is consistent between different impls

**Comment by Julian** [comment]
That also sounds reasonable, it's not a long list (feel free to open an issue for someone to do it obviously).

**Comment by sd1p** [comment]
Replaced supported dialects URLs with badges. Which suggestion should I pick out of 3? 
image

**Comment by harrel56** [comment]
@sd1p the one which you have on your last screenshots is fine - just push what you've already done :)

**Review by harrel56 — COMMENTED** [review]
Still lacking this parts:
> 1. To extract the code I posted above (from index.tsx) and then use reportUri variable to construct this URL - instead of using https://bowtie-json-schema.github.io/bowtie. As for the file placement - idk, /data folder seems to have all kinds of utilities now, so maybe it can go there.
> 2. To extract creation of image URL (https://img.shields.io/endpoint...) - just put it above react component definition

**Comment by sd1p** [comment]
> Still lacking this parts:
> 
> > 1. To extract the code I posted above (from index.tsx) and then use reportUri variable to construct this URL - instead of using https://bowtie-json-schema.github.io/bowtie. As for the file placement - idk, /data folder seems to have all kinds of utilities now, so maybe it can go there.
> > 2. To extract creation of image URL (https://img.shields.io/endpoint...) - just put it above react component definition

While implementing 1, I moved the reportURI constructing logic to frontend/src/data/ReportHost.ts as importing directly from index.ts may cause circular imports.

2, is also done.

Now I am working on


https://github.com/bowtie-json-schema/bowtie/pull/929#discussion_r1501877787

File: frontend/src/data/Dialect.ts [hunk 6]
```diff
@@ -9,6 +9,7 @@ import { fromSerialized } from "./parseReportData";
 export default class Dialect {
   readonly path: string;
   readonly prettyName: string;
+  readonly shortName: string;
```
Julian, line 12: `shortName` is (should be) exactly the same as what the UI calls `path`, I just thought that name made no sense when I extracted out the data for use in the backend too -- we should probably rename `path` rather than adding a duplicate field, or else make `path` be a property that just returns `shortName` or something, but that also could wait for a follow up PR.

sd1p: I can work on this probably in the next PR. This will require many changes in the backend as well as frontend.

Julian:
Good! Yes not only are you allowed, you have to :) that test is the one that tells us the behavior's changed -- ideally we would have an *integration* test which also would have caught your previous error (changing the frontend without changing that `bowtie/_cli.py` part). If you really want extra credit you could try thinking about that (with inspiration from how Dialect.test.ts works, I did a first example there) but it may not be trivial to do that, so feel free to just do the bit updating `integration_tests.py` as well.

frontend/src/data/Dialect.ts
sd1p: I clearly got it now and made changes in Dialect.ts and now frontend uses `path` instead of `shortName`.

Julian:
Whoops, I didn't mean to merge that, but I managed to confuse myself between our main branch and yours. No matter, I was going to show you a few tricks and then let you do the last bit, but I guess since I merged I'm going to have to do that. The site is anyways going to break because we push UI changes immediately from main, I've been meaning to fix that and only have the UI update on release, same as the backend, but ok, not a big deal, this isn't 0 downtime software.

PR #932, opened 2024-02-24
Generating implementations file during hatch(ling) build process

### LINKED ISSUE #921: Generate the then implementations json file during the hatchling build process
Most of the discussion and reasoning part has been covered in PR #920 

@Julian kindly assign this issue to me

adwait-godbole:
Fixes #921 

Hey @Julian . This is just my first try / thoughts that I have come up with referring to the official docs of `hatchling` to use a [custom build hook plugin](https://hatch.pypa.io/1.9/plugins/build-hook/custom/). 

Here rather than executing the `bowtie info` command to generate the `implementations.json` file in the hatch build hook plugin (where we technically don't have access to the `bowtie` command) I have abstracted that logic to the GitHub actions workflow `ci.yml` where the final `build` actually happens (`nox -s app -- .`).

I have used the `hatch build hook plugin` to `force-include` the generated json file during the workflow in the final `(wheel)` packaged application. So the file is still dynamically getting generated during the build process but not explicitly in the hook plugin rather that plugin is just responsible for taking that dynamically generated file and `force-including` it in the packaged app.

Please have a look and let me know, always looking forward for your inputs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,21 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - name: Install Bowtie
+        uses: ./
+      - run: docker system df
+      - name: Pre-fetch all images
+        run: |
+          for each in implementations/*; do
+            docker pull "ghcr.io/bowtie-json-schema/$(basename $each):latest" &
+          done
+          wait
+      - name: Generate implementations.json file
+        run: |
+          bowtie info $(ls implementations/ | sed 's/^/-i /') --format json > implementations.json
+      - run: echo "IMPLEMENTATIONS_METADATA_FILE_PATH=$(pwd)/implementations.json" >> $GITHUB_ENV
+      - run: docker system df
+        if: always()
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,10 +123,11 @@ jobs:
             docker pull "ghcr.io/bowtie-json-schema/$(basename $each):latest" &
           done
           wait
-      - name: Generate implementations.json file
+      - name: Generate implementations metadata file
         run: |
           bowtie info $(ls implementations/ | sed 's/^/-i /') --format json > implementations.json
-      - run: echo "IMPLEMENTATIONS_METADATA_FILE_PATH=$(pwd)/implementations.json" >> $GITHUB_ENV
+      - name: Set the generated file path env variable
+        run: echo "IMPLEMENTATIONS_METADATA_FILE_PATH=$(pwd)/implementations.json" >> $GITHUB_ENV
       - run: docker system df
         if: always()
       - name: Set up Python
@@ -141,6 +142,8 @@ jobs:
       - name: Set up nox
         uses: wntrblm/nox@2023.04.22
       - name: Build a Bowtie PyApp
+        env:
+          RUNNING_HATCH_SESSION: "true"
         run: nox -s app -- .
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,22 +114,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Install Bowtie
-        uses: ./
-      - run: docker system df
-      - name: Pre-fetch all images
-        run: |
-          for each in implementations/*; do
-            docker pull "ghcr.io/bowtie-json-schema/$(basename $each):latest" &
-          done
-          wait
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -114,6 +114,22 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - name: Install Bowtie
```
Julian, line 117: The goal is to not need any changes to the CI setup. I haven't reviewed the rest yet, it actually from a skim looked like it was probably correct or close to it though -- have you tried it without any of these CI changes?

hatch_build.py
```diff
@@ -0,0 +1,41 @@
...
+        version: str,
+        build_data: dict[str, Any],
+    ) -> None:
+        """
+        Includes the json file in the final build.
+
+        Reads the JSON file path from environment variable and
+        force includes it in the final build.
+        """
+        if not os.environ.get("RUNNING_HATCH_SESSION"):
```
Julian, line 32: Oh I see, you don't want the file generated outside -- here itself is where you should be generating this file / data!

adwait-godbole:
I hadn't realized it until today that me doing changes in `ci.yml` file had already guaranteed the inclusion of `implementations.json` file in the final build but regardless that was not the correct way.

adwait-godbole:
@Julian I did the changes only constrained to the build hook plugin now but I have been unsuccessful in removing this error.

`hatch_build.py:40:13: S603 `subprocess` call: check for execution of untrusted input`

Also I am not even sure if my code is even getting past the below line although I have set the env variable in `noxfile.py`

```python
if not os.environ.get("RUNNING_HATCH_SESSION"):
     return
```

Also can you guide me how do I verify that the `implementations.json` file has been included in the final build ? I tried doing `py -m build` and can see a `.tar.gz` and `.whl` but none of them really help to find out.

Also the fact that this build hook plugin will have access to the underlying file system and environment variables of the ubuntu machine of Github job seems hard to digest.

Julian:
You did a really good job here, you did the hard part which is getting a bit in the bowels of the packaging tool, so well done. The last bit really is the easy bit! I did it, rather than us going back and forth, because probably I wasn't clear at how to do this simply -- you can see what I did is really trivial, you don't need any installation, or to run bowtie, or anything -- your job is to simply "write down" the list of implementation names, because if we have those, then at runtime we have all the info we need.

I'm being a bit terse and not answering all your questions, which I apologize for -- badger me for answers if you're curious on something I didn't answer. Other than that you can have a look at the changes I made, hopefully they're self-explanatory.

> I tried doing py -m build and can see a .tar.gz and .whl generated but none of them really help to find out.

You have to untar (and unzip) these, and look at what's inside them and see whether it contains what you expect. But you're right that the way to test is first:

* `pyproject-build` to see whether the things in `dist/` have what you need
* but then if they do, step 2 is to create a virtualenv, install each of them (one first, then new venv then the other) and try *using* the project in the installed venv
* then of course have tests for the behavior
* then a bonus step which maybe distracted you is indeed checking the `app` env in the noxfile, but that builds what's called a PyApp, it's not even the "main" way people use Bowtie, so it's really just the last corner thing to check

But yeah well done again on tackling something not trivial and basically getting it close enough to the finish line that the rest was easy for me to knock off.

Julian:
(Once again GitHub is confused because I rebased your commits onto `main` before merging, but this time the changes are indeed there, you can have a look!)

adwait-godbole:
> You did a really good job here, you did the hard part which is getting a bit in the bowels of the packaging tool, so well done. The last bit really is the easy bit! I did it, rather than us going back and forth, because probably I wasn't clear at how to do this simply -- you can see what I did is really trivial, you don't need any installation, or to run bowtie, or anything -- your job is to simply "write down" the list of implementation names, because if we have those, then at runtime we have all the info we need.
> 
> I'm being a bit terse and not answering all your questions, which I apologize for -- badger me for answers if you're curious on something I didn't answer. Other than that you can have a look at the changes I made, hopefully they're self-explanatory.
> 
> > I tried doing py -m build and can see a .tar.gz and .whl generated but none of them really help to find out.
> 
> You have to untar (and unzip) these, and look at what's inside them and see whether it contains what you expect. But you're right that the way to test is first:
> 
> * `pyproject-build` to see whether the things in `dist/` have what you need
> * but then if they do, step 2 is to create a virtualenv, install each of them (one first, then new venv then the other) and try _using_ the project in the installed venv
> * then of course have tests for the behavior
> * then a bonus step which maybe distracted you is indeed checking the `app` env in the noxfile, but that builds what's called a PyApp, it's not even the "main" way people use Bowtie, so it's really just the last corner thing to check
> 
> But yeah well done again on tackling something not trivial and basically getting it close enough to the finish line that the rest was easy for me to knock off.

Thanks a lot @Julian!

Issue #931, opened 2024-02-24
Don't center the commit information in the UI's run info


Julian:
We are centering the `Commit` information on [the homepage](https://bowtie.report) it seems:

![Screenshot 2024-02-24 at 9 46 32 AM](https://github.com/bowtie-json-schema/bowtie/assets/329822/775df754-83c4-4ee5-b896-133d46eb86c3)

but we should just treat that like all the other fields and have it left aligned.

Akshaybagai52:
Hey @Julian Can I work on this issue ?

Akshaybagai52:
> Go for it!

Thanks @Julian

Akshaybagai52:
Hey @Julian this is the PR 
https://github.com/bowtie-json-schema/bowtie/pull/936
I've applied col-md-2 class with this mobile design remains unaffected 
let me know about your views
thanks

Julian:
Fixed, well done!

PR #933, opened 2024-02-24
Table styling modified

Swarnendu0123:
Fixes: #931

## Details:
I've incorporated the following CSS style:
```css
style={{ width: "25%" }}
```
This style setting ensures a consistent 25% width, maintaining the aspect ratio across various screen dimensions.

## Screenshot:
Screenshot (88)

----
📚 Documentation preview 📚: https://bowtie-json-schema--933.org.readthedocs.build/en/933/

### DIFF
--- a/frontend/src/components/RunInfo/RunInfoSection.tsx
+++ b/frontend/src/components/RunInfo/RunInfoSection.tsx
@@ -16,11 +16,11 @@ const RunInfoSection = ({ runInfo }: { runInfo: RunInfo }) => {
         <table className="table table-sm table-hover">
           <thead>
             <tr>
-              <td className="align-top">Dialect</td>
+              <td className="align-top" style={{ width: "25%" }}>Dialect</td>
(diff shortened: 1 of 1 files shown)


Julian:
I don't see what's in your screenshot here. Here I see:

Screenshot 2024-02-24 at 5 47 12 PM

How did you produce that screenshot? From the code here it looks like you have one more place that's missing the class modification (which is the one that would actually fix the issue). Perhaps you forgot to save, but quite odd :)?

Also, for consistency can you use the Bootstrap class rather than manual CSS, we I believe use that in most places already so should stick to it until/unless we decide to ditch it.

Swarnendu0123:
>Perhaps you forgot to save, but quite odd 

Yeah, my bad! Now i guess it's okay, actually I was just trying to implement bt inline css :)

Julian:
Similar to #933 you seem to have picked this up even though someone else had asked to work on it. I again didn't notice you were someone else until now. Please don't do that.

Issue #930, opened 2024-02-24
Reduce the amount of padding in the UI on mobile


Julian:
The mobile UI has too much padding:

![Screenshot 2024-02-24 at 9 45 20 AM](https://github.com/bowtie-json-schema/bowtie/assets/329822/6d03c310-0dba-4e88-b622-3da010877c3c)

We should make that look more normal, where there's only a small bit of space.

nomandhoni-cs:
Hey @Julian  I am, already doing some testing with this, how much padding will look better in mobile devices.
Also I am trying to test different style of navbar in mobile devices.

harrel56:
I noticed that implementation page (e.g. https://bowtie.report/#/implementations/ruby-json_schemer) still overflows on mobile - maybe it's worth to take care of it in scope of this issue too

nomandhoni-cs:
@harrel56 sure, I will try to take care of that issue too. Thanks for pointing that out.

Julian:
Fixed in da4a6edc51e4927a1748685a5d11de868ad9a838 though as I said on the (closed) PR I hadn't noticed that @nomandhoni-cs it was you and not the PR author that had commented. Sorry! If I had, I firstly would have closed the PR and told that person this ticket was assigned, but certainly I would not have put in a fix myself. Have a look at my commit and see if you agree with it or have a suggestion. I also didn't address @harrel56's overflow, so maybe you can help fix that, or otherwise your help is of course appreciated on any ticket!

PR #934, opened 2024-02-24
Padding and Schema & Instance issue fixed

Swarnendu0123:
Fixes: #930
 
## Details:
1. Changed padding from `p-4` to `p-2`, and it almost looks good on both, mobile and desktop mode. 
2. Also made the container scrollable by adding `overflow-auto`

## Screenshots:
Screenshot (94)
Screenshot (95)
Screenshot (92)
Screenshot (93)

----
📚 Documentation preview 📚: https://bowtie-json-schema--934.org.readthedocs.build/en/934/

### DIFF
--- a/frontend/src/components/Cases/SchemaDisplay.tsx
+++ b/frontend/src/components/Cases/SchemaDisplay.tsx
@@ -14,9 +14,9 @@ const SchemaDisplay = ({
   const schemaFormatted = JSON.stringify(schema, null, 2);
   const instanceFormatted = JSON.stringify(instance, null, 2);
   return (
-    <div className="card mb-3 mw-100">
+    <div className="card mb-3 mw-100 overflow-auto">
       <div className="row">
-        <div className="col-8 pe-0">
+        <div className="col-6 pe-0">
           <div className="d-flex align-items-center highlight-toolbar ps-3 pe-2 py-1 border-0 border-top border-bottom">
             <small className="font-monospace text-body-secondary text-uppercase">
               Schema
@@ -31,7 +31,7 @@ const SchemaDisplay = ({
             </pre>
           </div>
         </div>
-        <div className="col-4 border-start ps-0">
+        <div className="col-6 border-start ps-0">
           <div className="d-flex align-items-center highlight-toolbar ps-3 pe-2 py-1 border-0 border-top border-bottom">
             <small className="font-monospace text-body-secondary text-uppercase">
               Instance


Julian:
Thanks!

As I say often I'm no frontend expert, but it sounds odd that the fix would still have us having the same amount of padding on both devices -- does bootstrap not have classes which adjust either based on percentage of the screen size or else responsive ones? Surely it must (IIRC it has both).

Swarnendu0123:
> As I say often I'm no frontend expert, but it sounds odd that the fix would still have us having the same amount of padding on both devices -- does bootstrap not have classes which adjust either based on percentage of the screen size or else responsive ones? Surely it must (IIRC it has both).
Yes it have, but we need to configure bootstrap for it, https://getbootstrap.com/docs/4.0/layout/overview/

Julian:
(Those are old docs) but what do you mean by configure it?

Swarnendu0123:
Screenshot (96)

I am typing `w-lg-75` but the bootstrap is not responding, seems like we have missed something. Again I might be wrong. Just let me look around the whole codebase by tomorrow.

Swarnendu0123:
> (Those are old docs) but what do you mean by configure it?

I was talking about https://getbootstrap.com/docs/5.3/migration/#new-_mapsscss

Julian:
There's no way I buy either of those things I'm afraid :) -- you should be suspicious if such a simple thing seems so difficult, it happens of course, but certainly suspicion first.

Suggesting a major change in technologies for 2 things which are so similar is also never a good first thing to say (I'm telling you this because it's common in software to approach everything wanting to bring the tools you know -- indeed often a tool is better or worse than another, but it's something to be very careful with).

This was somewhat easy to fix, I flipped through the Bootstrap docs, found what seemed right, played a bit, and I believe have something better than before. It's likely still wrong for some other case, which we can fix, but yeah, feel free to have a look (closing this, as the fix is in da4a6edc51e4927a1748685a5d11de868ad9a838). Appreciate your attempt.

Julian:
Oh. Worse than that, looking at the issue, you didn't comment on it, and someone else had made it clear they were working on it! I honestly hadn't noticed that wasn't you until right now. Please don't do that, it's quite impolite. Sorry @nomandhoni-cs ! Have a look at what I did and tell me if you have suggestions for improvements.

Swarnendu0123:
> Oh. Worse than that, looking at the issue, you didn't comment on it, and someone else had made it clear they were working on it! I honestly hadn't noticed that wasn't you until right now. Please don't do that, it's quite impolite. Sorry @nomandhoni-cs ! Have a look at what I did and tell me if you have suggestions for improvements.

Sorry @Julian! Actually I thought it was a simple issue, so I just raised the PR directly. Next time I will definitely discuss before doing anything.

nomandhoni-cs:
@Julian,
Appreciate your taking over, now everything is aligned perfectly! Your use of ```<main>``` is fantastic, it really streamlines the code. I'm going start working on the navbar redesign. Please let me know if there's anything I can do.

Issue #896, opened 2024-02-19
Don't show the "About Bowtie" section on the local report upload page


Julian:
In #368 we added an "About Bowtie" section to explain what Bowtie is.

We should not show this explanation on the "local upload results" page (i.e.after someone uploads a report to https://bowtie.report/#/local-report/ ) -- someone who has done so almost certainly is very familiar with Bowtie, as they've managed to run and upload a test run.

Arguably we should even only show the About Bowtie info on the homepage, and not if someone browses away to a different page.

Siddharth-Singh-2004:
@Julian, I'll work on this.

Akshaybagai52:
Hey @Julian If it's not assigned then I would love to work on this 
Thanks

ashmit-coder:
Hey @Julian can you look into the pr #937

Julian:
@Siddharth-Singh-2004 going to assume you haven't made progress here -- @Akshaybagai52 please go for it if you're still interested.

Akshaybagai52:
Hey @Julian 
can you clarify me this that you only want to show about bowtie for this url 
![image](https://github.com/bowtie-json-schema/bowtie/assets/100753870/033c9653-1f51-461c-9e9d-c521e535d38a)

Right now it is about bowtie page for all the url's of dialects as well as after anyone uploads local-report
![image](https://github.com/bowtie-json-schema/bowtie/assets/100753870/c8d2c643-5d01-4223-b0a0-25f40e2925c2)

Question :- Only want to remove the about bowtie section for 'local-report' page or for all the dialects pages as well 
Thanks

Julian:
Only making it show on the homepage (https://bowtie.report/#/) would probably be the best fix!

What we have right now for that homepage is equivalent to the 2020-12 dialect page (https://bowtie.report/#/dialects/draft2020-12). but if possible, yeah, make it only show up if someone visits the homepage, not if they directly go to the 2020-12 dialects page (or any other page).

But if that's tricky, you can remove it only from the local reports page, since that's the main one where it really doesn't make sense.

Let me know if that clarifies.

Akshaybagai52:
Okay will let you know about the fix

Akshaybagai52:
Hey @Julian 
Can you review this pr https://github.com/bowtie-json-schema/bowtie/pull/938
let me know about your thoughts
Thanks

PR #937, opened 2024-02-25
fixes: #896, dont show about in report page

ashmit-coder:
Fixes: #896 

The About section has been condinitionally rendered according to the above issue.

The component now returns null if it is being rendered in /local-report else renders else where

----
📚 Documentation preview 📚: https://bowtie-json-schema--937.org.readthedocs.build/en/937/

### DIFF
--- a/frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
+++ b/frontend/src/components/BowtieInfo/BowtieInfoSection.tsx
@@ -4,7 +4,7 @@ const BowtieInfoSection = () => {
   // prettier-ignore
   const location = useLocation();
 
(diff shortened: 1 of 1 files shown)


ashmit-coder:
> Hi, thanks. Please don't send PRs for things where it seems like someone else is working on it unless you specifically clear that it's available, it's quite impolite to your peers.
> 
> Also, have you tried this? It doesn't appear to work, though the reason why looks trivial to fix.

Sorry for the inconvinience I thought as no one was assigned for the issue and no existing PR, it would be a good idea to just geneate a PR and then communicate over it. 

Will look into being assigned at first next time :)

PR #938, opened 2024-02-25
fix: showing about bowtie to only home page

Akshaybagai52:
----
📚 Documentation preview 📚: https://bowtie-json-schema--938.org.readthedocs.build/en/938/

### DIFF
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -8,10 +8,10 @@ import { useSearchParams } from "./hooks/useSearchParams.ts";
 
 export const DialectReportView = ({
   reportData,
-  topPageInfoComponent,
+  topPageInfoSection,
 }: {
   reportData: ReportData;
-  topPageInfoComponent: React.ReactElement | null;
+  topPageInfoSection?: React.ReactElement;
 }) => {
   const params = useSearchParams();
   const languages = useMemo(() => {
@@ -37,9 +37,7 @@ export const DialectReportView = ({
   return (
     <main className="container-lg">
       <div className="col col-lg-8 mx-auto">
-        {topPageInfoComponent &&
-          React.isValidElement(topPageInfoComponent) &&
-          topPageInfoComponent}
+       {topPageInfoSection}
         <RunInfoSection runInfo={filteredData.runInfo} />
         <FilterSection languages={languages} />
         <SummarySection reportData={filteredData} />
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -37,7 +37,7 @@ export const DialectReportView = ({
   return (
     <main className="container-lg">
       <div className="col col-lg-8 mx-auto">
-       {topPageInfoSection}
+        {topPageInfoSection}
         <RunInfoSection runInfo={filteredData.runInfo} />
         <FilterSection languages={languages} />
         <SummarySection reportData={filteredData} />
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -11,7 +11,7 @@ export const DialectReportView = ({
   topPageInfoComponent,
 }: {
   reportData: ReportData;
-  topPageInfoComponent:  React.ReactElement | null;
+  topPageInfoComponent: React.ReactElement | null;
 }) => {
   const params = useSearchParams();
   const languages = useMemo(() => {
@@ -37,7 +37,9 @@ export const DialectReportView = ({
   return (
     <main className="container-lg">
       <div className="col col-lg-8 mx-auto">
-       {topPageInfoComponent && React.isValidElement(topPageInfoComponent) && topPageInfoComponent}
+        {topPageInfoComponent &&
+          React.isValidElement(topPageInfoComponent) &&
+          topPageInfoComponent}
         <RunInfoSection runInfo={filteredData.runInfo} />
         <FilterSection languages={languages} />
         <SummarySection reportData={filteredData} />
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -1,32 +1,19 @@
 import CasesSection from "./components/Cases/CasesSection";
-import BowtieInfoSection from "./components/BowtieInfo/BowtieInfoSection";
 import RunInfoSection from "./components/RunInfo/RunInfoSection";
 import SummarySection from "./components/Summary/SummarySection";
-import { useMemo } from "react";
+import React, { useMemo } from "react";
 import { ReportData } from "./data/parseReportData.ts";
 import { FilterSection } from "./components/FilterSection.tsx";
 import { useSearchParams } from "./hooks/useSearchParams.ts";
 
 export const DialectReportView = ({
   reportData,
+  topPageInfoComponent,
 }: {
   reportData: ReportData;
+  topPageInfoComponent:  React.ReactElement | null;
 }) => {
   const params = useSearchParams();
-
-  function showAboutBowtie() {
-    const initialUrl = `${window.location.protocol}//${window.location.host}${window.location.pathname}`;
-    const currentUrl = window.location.href;
-    const validUrls = [
-      `${initialUrl}#/#cases`,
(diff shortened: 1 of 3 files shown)


frontend/src/DialectReportView.tsx
```diff
@@ -14,6 +14,19 @@ export const DialectReportView = ({
 }) => {
   const params = useSearchParams();
 
+  function showAboutBowtie() {
+    const initialUrl = `${window.location.protocol}//${window.location.host}${window.location.pathname}`;
+    const currentUrl = window.location.href;
```
AgniveshChaubey, line 19: Since we are using React, it is better to utilize the available hooks instead of plain javascript. I guess `useLocation` hook is what you need. - https://reactrouter.com/en/main/hooks/use-location

Julian: I'm not sure if this is the same comment or a related one, but it would be a lot better to not hardcode a list of URLs as well -- surely there's a way of doing this where we only enable it on the component which renders the homepage, and not just do that filtering afterward by URL.

harrel56 (commented):
You really shouldn't do path related checks in a component - all that stuff should be in a React Router's routes configuration (in `index.tsx`). What if in the future someone changed `#cases` to e.g. `#test-cases`? This would break and it would be really unexpected.

TBH I think React Router lacks an easy functionality of passing down some params - e.g. if "about info" should be displayed. It looks like it can only be achieved via loaders. So, in the current state local report page does not have any loader data (opposed to any other report page). Therefore something like:
```ts
const loaderData = useLoaderData()
const showAboutInfo = Boolean(loaderData)
```
should be enough to make it work

frontend/src/DialectReportView.tsx
Akshaybagai52: > I'm not sure if this is the same comment or a related one, but it would be a lot better to not hardcode a list of URLs as well -- surely there's a way of doing this where we only enable it on the component which renders the homepage, and not just do that filtering afterward by URL.

Okay I will check some other approach and will let you know about the progress.

Akshaybagai52:
> You really shouldn't do path related checks in a component - all that stuff should be in a React Router's routes configuration (in `index.tsx`). What if in the future someone changed `#cases` to e.g. `#test-cases`? This would break and it would be really unexpected.
> 
> TBH I think React Router lacks an easy functionality of passing down some params - e.g. if "about info" should be displayed. It looks like it can only be achieved via loaders. So, in the current state local report page does not have any loader data (opposed to any other report page). Therefore something like:
> 
> ```ts
> const loaderData = useLoaderData()
> const showAboutInfo = Boolean(loaderData)
> ```
> 
> should be enough to make it work

@Julian is this what I need to do?

Julian:
Yes, @harrel56 said more or less the same thing I believe but was actually more helpful because he knew how to do what I mentioned off-hand :)

(Though I should say I'd really have expected this to work where you have some component which is "the dialect report", and then the home page should be `"<about bowtie><dialect report>"` where it composes the content from the other page. But yeah anything is better than hardcoding URLs.

Akshaybagai52:
> Yes, @harrel56 said more or less the same thing I believe but was actually more helpful because he knew how to do what I mentioned off-hand :)
> 
> (Though I should say I'd really have expected this to work where you have some component which is "the dialect report", and then the home page should be `"<about bowtie><dialect report>"` where it composes the content from the other page. But yeah anything is better than hardcoding URLs.

Okay will check that

harrel56:
I believe, for local report loader data was undefined - if it's not, what kind of object is it? I cannot currently check it by myself

harrel56:
After @Julian comment I now can see some different options for implementing this:
1. Pass boolean 'showAboutInfo' as a component property
2. Pass a whole "about info" component as a child
3. Pass a whole "about info" component as a component property

I think these are more clear than the approach with loader data. IMO option 3 seems best, in future we can control more elements like this if needed

Akshaybagai52:
> 1. showAboutInfo

Let me check that

Akshaybagai52:
Hey @Julian @harrel56 can you review the PR now
Thanks

frontend/src/components/DragAndDrop/DragAndDrop.tsx
```diff
@@ -80,7 +80,7 @@ export const DragAndDrop = () => {
   };
 
   if (lines) {
-    return <DialectReportView reportData={lines} />;
+    return <DialectReportView reportData={lines} topPageInfoComponent={null} />;
```
harrel56, line 83: Maybe make the prop optional and don't set it here at all?

frontend/src/DialectReportView.tsx
```diff
@@ -37,7 +37,9 @@ export const DialectReportView = ({
   return (
     <main className="container-lg">
       <div className="col col-lg-8 mx-auto">
-        <BowtieInfoSection />
+        {topPageInfoComponent &&
```
harrel56, line 40: Wouldn't it work if you just tried to render this info component? Without the additional checks? If it's null or undefined then it just doesn't get rendered I think?

Akshaybagai52:
@harrel56 can you review now ?

frontend/src/DialectReportView.tsx
```diff
@@ -1,19 +1,19 @@
 import CasesSection from "./components/Cases/CasesSection";
-import BowtieInfoSection from "./components/BowtieInfo/BowtieInfoSection";
 import RunInfoSection from "./components/RunInfo/RunInfoSection";
 import SummarySection from "./components/Summary/SummarySection";
-import { useMemo } from "react";
+import React, { useMemo } from "react";
```
harrel56, line 4: React import is unused, right?

harrel56 (approved):
lgtm

frontend/src/DialectReportView.tsx
Akshaybagai52: Removed and thanks for your help.

Siddharth-Singh-2004:
@Julian, @harrel56 @AgniveshChaubey: Apologies for delayed response due to exams. 

I was looking at solution that involved **Context & Hooks** reason being:

Apart from local report  not needing  'About Bowtie':
I was attempting to handle the following situations:
(1). Any click on implementations link has to be handled remembering we came in either from Local Reports or a Dialect. What ever was the **state** of About Bowtie remains.
(2).  Clicking the Bowtie icon (aka, behave like someone has come in the first time or clicked the home button). The **state** of About Bowtie reverts to show the info section.
(3). Clicking Bowtie icon from Implementations link screen should retain context of local report or dialect report (as in 1 above). 
(4). User goes to Local Report but does not Upload, but instead uses (1) or (2) above to navigate back. Previous **state** remains.

Summary: Navigating to various links from the screen loaded on local report or dialect drop down would remember the **context** of the loader.
Was trying my best to ensure all cases above are handled if we were to meet the objectives of the original issue. 

Please tick or cross the above cases so @Akshaybagai52 can enhance the PR to handle the additional checks above.

frontend/src/DialectReportView.tsx
Julian: Did CI not fail with an unused import? If not can you open a bug, though it looked like tooling in JS for imports is .. not great.

Siddharth-Singh-2004:
@harrel56, Thank you for the revert. I missed out on the additional clarifications in the PR during my quick read. 

Only thing that  is not addressed in this code fix is the condition when the report loads up for the first time or when user clicks bowtie icon/hompage - https://bowtie.report/#/ explicitly, the About Bowtie information must show up. I'm assuming this behaviour is desired and therefore needs a fix.

Julian:
Thanks! I think this definitely looks like an improvement. Will merge and we can see what we think about the behavior if it needs any other tweaks. Well done @Akshaybagai52 

> I'm assuming this behaviour is desired and therefore needs a fix.

harrel56:
@Siddharth-Singh-2004 
> Only thing that is not addressed in this code fix is the condition when the report loads up for the first time or when user clicks bowtie icon/hompage - https://bowtie.report/#/ explicitly

But it does show up, right? Or am I missing something

Siddharth-Singh-2004:
I confirm that it is working with the latest merge.

Issue #37, opened 2022-09-21
Add (client-side) table filters and sorts to the report


Julian:
* Show only cases with disagreements
* Show "odd-man-out" implementations?
* Show errors

Swarnendu0123:
hello, Can you explain more about this issue?

Julian:
It's not well described clearly (which is why it's not marked good first issue). But in short a first version of this would be to (live) filter the accordions at the bottom of the report page by a substring contained in their description, so that only matching test caess were shown. A second would be a dropdown which would let you select between showing all of the accordions, or just ones with failures, or just ones with errors. (And there are some more fancy versions of the above, but that's a quick first 2).

If you think you can help it'd definitely be appreciated, in which case let me know, and then feel free to tear off just 1 or both of those 2 filters and see what you can do!

Swarnendu0123:
Yes! let me give it a try

Swarnendu0123:
https://github.com/bowtie-json-schema/bowtie/assets/94161758/465fdfa3-6e42-434b-a13d-10b15521cffb

I have developed, let me know if there are any improvements. my laptop is little bit lagging, sorry for that.

PR #939, opened 2024-02-25
Developed filter and search functionality in the Cases Section

Swarnendu0123:
Fixes: #37
## Description:
Developed filter and search functionality in the Cases Section.

## Screenshot(s):
Screenshot (97)
Screenshot (98)

## Video Demonstration:

https://github.com/bowtie-json-schema/bowtie/assets/94161758/5823aa1f-c4e7-4f0a-b859-02c14f1b8303

----
📚 Documentation preview 📚: https://bowtie-json-schema--939.org.readthedocs.build/en/939/


Julian:
Will take a closer look in the next few days, video looks nice though!

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,19 +1,83 @@
...
-          caseData={caseData}
-          implementations={implementations}
-        />
-      ))}
+      <Row>
+        <Col
+          md={{ span: 6, offset: 6 }}
+          className="d-flex align-items-center justify-content-end"
+        >
+          <input
```
Julian, line 25: Something strange seems to be happening with the search box causing the top of the first accordion to be missing:

Screenshot 2024-02-27 at 8 05 26 PM

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,19 +1,83 @@
...
+            type="text"
+            onChange={handleSearchChange}
+            placeholder="Search"
+            className="form-control m-4"
+          />
+          <DropdownButton
+            title={filterCriteria ? `Filter: ${filterCriteria}` : "Filter"}
```
Julian, line 32: Probably better for this to be checkboxes, i.e.:

Test Result
* Success
* Failed
* Errored
* Skipped

and you check or uncheck which ones you want to see. It should also be possible to dynamically get the list of possible result statuses rather than hardcoding them here -- I'm not sure we'll ever have any others, but maybe it's worth doing so anyways, it'll make the code a bit shorter I suspect.

Julian (commented):
I've left a few comments on the code itself, but here are some other general ones:

* As you type, the page jumps around. In particualr, when nothing matches (e.g. if you type `asdf`) the page sort of snaps quite unsatisfyingly, so that the search bar is at the bottom. It'd be a lot better if the position of the search bar didn't change as you type.
* I think it'd be quite nice to have highlighing in some form -- in other words, while you type, it would be nice to highlight the parts of descriptions that are matching your live search in the accordions below.
* I'm not 100% sure it'd help as I haven't used it, I'd just saved it to look at, but perhaps https://github.com/algolia/autocomplete can help us with this functionality.
* It would be nice to be able to search also within the test description (as opposed to the test case description). I'm not sure whether we need a separate box for each, maybe we just always search through both? Then again I think it would be nice to also search through the schema and instance -- and there it does seem like you should be able to turn that on or off. Maybe we should have togglable buttons for "Search in: [ ] schema [ ] instance [ ] test [ ] test case".
* Taking that yet further, probably it would be nice to have a dropdown which lets you select which *implementations* you want to see results for. But feel free to leave that for a next PR.

Thanks for your work so far, let me know what you think.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: > okay, so we can have a fixed scrollable container, inside which we can display all the tests, and if the search string is not matching to any tests, we can show no match found or something like that

Its due to some padding issue, I will handle it, while implementing.

frontend/src/components/Cases/CasesSection.tsx
Julian: Just the 4 should be fine, if someone wants all they can just check all 4 (and we can have all 4 selected by default). You can also have "only" links next to each box which unselect all but the one.

Swarnendu0123: cool! I will do it.

Swarnendu0123:
What about the UI now?

https://github.com/bowtie-json-schema/bowtie/assets/94161758/fa921df2-1597-48bb-81ea-75c26cca4778

Swarnendu0123:
> It would be nice to be able to search also within the test description (as opposed to the test case description). I'm not sure whether we need a separate box for each, maybe we just always search through both? Then again I think it would be nice to also search through the schema and instance -- and there it does seem like you should be able to turn that on or off. Maybe we should have togglable buttons for "Search in: [ ] schema [ ] instance [ ] test [ ] test case".

I have an idea for enhancing the search functionality. Within the same SearchBox, users could specify the scope of their search. For instance, if someone wants to search within the schema, they could type `schema: properties`. This way, they can narrow down their search to specific areas such as properties in the schema. We could extend this functionality to other areas as well.

@Julian what do you suggest?

Julian:
You're talking about Lucene syntax. That would certainly be nice yeah. Though I don't want to tackle too big of a bite all at once. I'd like to see something fully mergeable in <200 lines of changes, and if need be to add more functionality in follow-ups. So I fully believe making lucene syntax work is doable, and probably there are libraries that will help us do it just like e.g. GitHub search works, but yeah try not to go wild all at once unless you can get it all working well, it's more important that you work reliably -- your first version here was too hasty and didn't work well, so err on the side of "working even if less functionality".

Swarnendu0123:
I have changed the description type to `description: string | JSX.Element`. that's why I guess it is breaking the system again and again, let me update the corresponding files.

Julian:
That sounds like a very strange change. But also again, work slow and correct rather than haphazardly.

Julian:
Politely no. It's been less than an hour since we had a design discussion, and four hours total since we first discussed. There's no way that's enough time spent on your part just making sure you have as much worked out as you can before asking specific questions. Being direct, it makes it seem even more like laziness -- I've tried to subtly give you this feedback so I'll try less subtle now :). When I say "work more carefully" this is what I mean. Don't throw something poorly thought out over to me for me to work out the problems with. It's ok to get stuck, it's not ok to spend minimal time and then kick it over.

Swarnendu0123:
Actually, I didn't mean that, because it is perfectly working in my local system (I have shown it in the video).

We have to make it JSX.Element because the description is highlighted now, so it can't be a simple string:
```jsx
<mark key={index} className="bg-primary text-dark"> 
       {part} 
 </mark>
```
That's why I have changed the description type. I just want to know how the UI looks. Just in general how you feel about the UI.

I will push a commit over it, after that most probably the tests will be passing.

Swarnendu0123:
Hello @Julian! please review my changes and give me some feedback.

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,116 @@
...
+              Successful
+            </Dropdown.Item>
+            <Dropdown.Item onClick={() => handleFilterChange("errors")}>
+              Errors
+            </Dropdown.Item>
+            <Dropdown.Item onClick={() => handleFilterChange("skipped")}>
+              Skipped
+            </Dropdown.Item>
+            <Dropdown.Item onClick={() => handleFilterChange("failed")}>
+              Unsuccessful
```
Julian, line 76: "Failed" and "Unsuccessful" mean different things (in the parlance of Bowtie). Finding 2 words that communicate the difference between them is quite difficult, which is why in other places we have tooltips (at least until we come up with something better). But definitely we should not *mix* them by labeling one number with the other. Have a look at other places that use the two and see if you can tell what the difference is, otherwise ask.

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,116 @@
...
+  const filteredCases = Array.from(reportData.cases.entries())
+    .filter(([seq]) => {
+      if (!filterCriteria) return true;
+      const caseResults: (CaseResult | undefined)[] = Array.from(
+        reportData.implementations.values(),
+      )
+        .map((impl) => impl.cases.get(seq))
+        .flat();
+      switch (filterCriteria) {
+        case "successful":
```
Julian, line 27: (Here too, we talked about having checkboxes so that someone can express "either X or Y" -- you didn't do that, which is fine, but you at least should say "I left that for later" or "I don't have time to work on that right now".)

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,116 @@
...
+          className="d-flex align-items-center justify-content-end"
+        >
+          <input
+            type="text"
+            onChange={handleSearchChange}
+            placeholder="Search"
+            className="form-control me-2"
+          />
+          <DropdownButton
+            title={filterCriteria ? `Filter: ${filterCriteria}` : "Filter"}
```
Julian, line 59: Saying just "filter" isn't clear enough I suspect, so at very least, ignoring the other feedback, we should try something like 
```suggestion
            title={filterCriteria ? `Filter: ${filterCriteria}` : "Filter by Outcome"}
```

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,116 @@
...
+        case "failed":
+          return caseResults.some((result) => result?.state === "failed");
+        default:
+          return true;
+      }
+    })
+    .filter(
+      ([, caseData]) =>
+        typeof caseData.description === "string" &&
+        caseData.description.toLowerCase().includes(searchText),
```
Julian, line 42: This seems like it leaves open the possibility for edge case bugs -- specifically, in the other part of the code you use a Regex to match on the description, and here you use `string.includes`. That seems like it will go wrong if/when the two of those ever disagree. It seems like perhaps you want to do the matching *ahead* of time so you already know what part matched here.

Julian (commented):
This sort of does a somewhat arbitrary subset of the things we discussed, so I'm not sure what to even be expecting here.

It's quite odd honestly, I gave you feedback here previously, what you should learn (for your career!) is that a good way to respond is to say something like "Hey! Here's an explanation of what I did, I did X Y and Z, and while working I noticed A B and C, and we discussed ___ and I think ___ and that's why ___" to show you're putting at least some thought in here. Most notably, a big thing we discussed last round was a library to do some of this work -- did you look at it? What happened when you did? Was it not useful, etc. All these are the sorts of things others will expect if you submit a PR and have some discussion and then go off and do some work.

Your work is again appreciated, but again directly, you have a somewhat strange work style so far.

I left some comments on the code.

Thanks.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Got it, thanks for the clarification.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Understood, I'll make sure to address the checkbox functionality and provide clarity in future updates. Thank you for the reminder.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Is it possible for us to implement something similar to that?
Screenshot (110)
Screenshot (109)

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Sure, let's utilize regex in both instances:

```javascript
    .filter(
      ([, caseData]) => {
        if (typeof caseData.description !== "string") return false;
        const regex = new RegExp(searchText, 'i');
        return regex.test(caseData.description);
      }
    );
```

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+
+  const handleCheckboxChange = (criteria: string) => {
+    if (filterCriteria.includes(criteria)) {
+      setFilterCriteria(filterCriteria.filter((item) => item !== criteria));
+    } else {
+      setFilterCriteria([...filterCriteria, criteria]);
+    }
+  };
+
+  const filteredCases = Array.from(reportData.cases.entries())
```
harrel56, line 29: This heavy filtering is done on every rerender - let's memoize it

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+  ButtonGroup,
+} from "react-bootstrap";
 
 const CasesSection = ({ reportData }: { reportData: ReportData }) => {
-  const implementations = Array.from(reportData.implementations.values());
+  const [searchText, setSearchText] = useState<string>("");
+  const [filterCriteria, setFilterCriteria] = useState<string[]>([]);
+
+  const handleSearchChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setSearchText(e.target.value.toLowerCase());
```
harrel56, line 18: I think the state change should happen with help of `useTransition` (https://react.dev/reference/react/useTransition). Haven't really used it much but probably you should use it on this state or maybe on `filteredCases`

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+        reportData.implementations.values(),
+      )
+        .map((impl) => impl.cases.get(seq))
+        .flat();
+      return caseResults.some((result) =>
+        filterCriteria.includes(result?.state ?? ""),
+      );
+    })
+    .filter(([, caseData]) => {
+      if (typeof caseData.description !== "string") return false;
```
harrel56, line 42: I think `caseData.description` is always a string?

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+    if (filterCriteria.includes(criteria)) {
+      setFilterCriteria(filterCriteria.filter((item) => item !== criteria));
+    } else {
+      setFilterCriteria([...filterCriteria, criteria]);
+    }
+  };
+
+  const filteredCases = Array.from(reportData.cases.entries())
+    .filter(([seq]) => {
+      if (filterCriteria.length === 0) return true;
```
harrel56, line 31: Just don't call `filter` at all if there are no filter criterion

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+      const caseResults: (CaseResult | undefined)[] = Array.from(
+        reportData.implementations.values(),
+      )
+        .map((impl) => impl.cases.get(seq))
+        .flat();
+      return caseResults.some((result) =>
+        filterCriteria.includes(result?.state ?? ""),
```
harrel56, line 38: can `result` and `result.state` be null?

frontend/src/components/Cases/CasesSection.tsx
harrel56: Also, maybe trim the search text?

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
-          seq={seq}
-          caseData={caseData}
-          implementations={implementations}
-        />
-      ))}
-    </Accordion>
+    <div>
+      <Row className="mt-3">
+        <Col
+          md={{ span: 6, offset: 6 }}
```
harrel56, line 51: input and dropdown share the same space - if the texts on dropdown button changes eveyrhing changes its width slightly. I would put them i separate columns

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+          className="d-flex align-items-center justify-content-end"
+        >
+          <input
```
harrel56, line 54: I'd suggest putting a nice search icon on the right. Maybe it can be achieved with this? https://react-bootstrap.netlify.app/docs/forms/input-group/

harrel56 (commented):
I'm assuming the rest of the suggestions made by @Julian are not in scope of this PR. But as for the major concerns:
- using 3rd party library - suggested lib has React variant (https://www.npmjs.com/package/react-instantsearch-core) but it looks kind of big for what we are trying achieve here, still maybe it's the best solution. I had good experience with `mui-base` but it's still in beta and you probably would still have to implement all the filtering and highlighting - but I love it for its flexible API
- filtering by not just a test description - schema and instance data would have to be converted to string (from json). But even now it feels kind of laggy so I don't know how it would be like. BTW probably there can be just a one regex test - maybe that would help a little

Swarnendu0123:
Thanks for the review @harrel56 !
>using 3rd party library - suggested lib has React variant (https://www.npmjs.com/package/react-instantsearch-core) but it looks kind of big for what we are trying achieve here, still maybe it's the best solution. I had good experience with mui-base but it's still in beta and you probably would still have to implement all the filtering and highlighting - but I love it for its flexible API

Yes! We can probablly add that point into https://github.com/json-schema-org/community/issues/609, as we are redesigning the whole section in this project.

filtering by not just a test description - schema and instance data would have to be converted to string (from json). But even now it feels kind of laggy so I don't know how it would be like. BTW probably there can be just a one regex test - maybe that would help a little.

That's actually a great idea! and I actually suggested `Lucene Syntax` for this section. for example, 
```
schema:"something" AND instance:"something"
```
this will make much easier the search. Also it can we include in the https://github.com/json-schema-org/community/issues/609.

what are your views on it? @harrel56 @Julian.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: cool! I will do it.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: I have tried this, 
Screenshot (114)
How is it looking now?

frontend/src/components/Cases/CaseItem.tsx
```diff
@@ -97,10 +100,46 @@ const CaseContent = ({
...
+}: CaseProps) => {
   const [content, setContent] = useState(<></>);
   const [, startTransition] = useTransition();
   const schemaDisplayRef = useRef<HTMLDivElement>(null);
+  const { isDarkMode } = useContext(ThemeContext);
+
+  const highlightDescription = (
+    description: string,
+    searchText: string,
+  ): JSX.Element => {
```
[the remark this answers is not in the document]
Swarnendu0123: Yes! you are right! Thanks

frontend/src/components/Cases/CasesSection.tsx
harrel56: For me the layout on last pic looks best!

frontend/src/components/Cases/CaseItem.tsx
```diff
@@ -97,10 +100,46 @@ const CaseContent = ({
...
+    if (!searchText) {
+      return <>{description}</>;
+    }
+
+    const regex = new RegExp(`(${searchText})`, "gi");
+    const parts: string[] = description.split(regex);
+    return (
```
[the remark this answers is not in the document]
Swarnendu0123: Yes! I have seen the problem. But I am not sure how to fix it. Can you suggest anything?

harrel56:
Hey @Swarnendu0123, I'm gonna be out for like a 2 weeks so probably won't be able to respond. Hopefully, you @Julian could check out my comments and pick up from there if needed, thanks!

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
harrel56: Unfortunately had no ideas for fixing this. But I'd suggest looking up some sites/libs with "highlighting" functionality and seeing in browser how it looks like in DOM - maybe they are using different tags or some CSS rules to make it work?

frontend/src/components/Cases/CaseItem.tsx
```diff
@@ -97,10 +100,46 @@ const CaseContent = ({
...
+  caseData,
+  implementations,
+  searchText,
+}: CaseProps) => {
   const [content, setContent] = useState(<></>);
   const [, startTransition] = useTransition();
   const schemaDisplayRef = useRef<HTMLDivElement>(null);
+  const { isDarkMode } = useContext(ThemeContext);
+
+  const highlightDescription = (
```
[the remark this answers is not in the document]
Swarnendu0123: Yeah! Implemented it. working very smooth now compared than previous!

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: cool!

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Yes! you are right! thanks for pointing out.

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Actually I have added it, result can be undefined.
Screenshot (119)

frontend/src/components/Cases/CasesSection.tsx
```diff
@@ -1,20 +1,119 @@
...
+
+        {filteredCases.length > 0 && (
+          <Accordion id="cases">
+            {filteredCases.map(([seq, caseData], index) => (
+              <CaseItem
+                key={index}
+                seq={seq}
+                caseData={{
+                  ...caseData,
+                  description: caseData.description,
```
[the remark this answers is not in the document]
Swarnendu0123: Oh sorry! actually the `highlighteDescription`  function was initially in `CasesSection.tsx`. but in the next commit I moved it into `CaseItem.tsx`. Thanks for pointing this out.

Swarnendu0123:
I have noticed something odd here, on selecting `successful` we are getting the whole array! 

https://github.com/bowtie-json-schema/bowtie/assets/94161758/04b83208-6a2d-4bfb-8983-ba5df4a52a01

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Cool, let's finalize it.

frontend/src/components/Cases/CaseItem.tsx
```diff
@@ -97,10 +100,46 @@ const CaseContent = ({
...
+    description: string,
+    searchText: string,
+  ): JSX.Element => {
+    if (!searchText) {
+      return <>{description}</>;
```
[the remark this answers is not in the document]
Swarnendu0123: >Wrapping it in fragment shouldn't be necessary.

Yeah!

> Also wondering if the whole if statement can be factored out?

I couldn't get this. Can you please explain a little bit more?

frontend/src/components/Cases/CaseItem.tsx
```diff
@@ -97,10 +100,46 @@ const CaseContent = ({
...
+    }
+
+    const regex = new RegExp(`(${searchText})`, "gi");
+    const parts: string[] = description.split(regex);
+    return (
+      <>
+        {parts.map((part, index) =>
+          regex.test(part) ? (
```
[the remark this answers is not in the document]
Swarnendu0123: Yeah! That will definitely be a better approach. Do you want me to implement that?

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: Like, if we don't call the function. On the initial render there will be no teat case. Do you mean that here?

Julian:
@Swarnendu0123 is this ready for review you think?

Swarnendu0123:
> @Swarnendu0123 is this ready for review you think?

I have implemented almost all the features which I could. Some suggestions by @harrel56 are still left. But we will improve it eventually through future PRs. So yes! it is ready for the review.

Swarnendu0123:
Hello @Julian! What the update? 
Also, I am happy to tell you that, I want to apply for GSoC 2024 the project is almost on the similar topic which we are trying to archive here. I have written the initial doc https://docs.google.com/document/d/1QY9bOBYp703pRjM15OKnNu09qs8GPKehlE26AJVSxnw for it. It will be very helpful, if you kindly go through the doc, and give me some feedback.

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
harrel56: I meant that you can get rid of this if statement and call the function like that: `searchText ? highlightDescription(caseData.description, searchText) : searchText` but honestly it does not make any difference so no need to change this

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
harrel56: Sure, why not

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
harrel56: Did you manage to fix it?

frontend/src/components/Cases/CasesSection.tsx
harrel56: Did you try using `useTransition`?

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
Swarnendu0123: I have researched on it, and found https://www.npmjs.com/package/react-highlight-words, seems like it can potentially solve the problem. Can we do that?

frontend/src/components/Cases/CasesSection.tsx
harrel56: Not really sure what you mean. I meant changing:
```ts
array.filter(() => if (condition) return true)
```
to
```ts
if (condition) {
  array.filter(...)
}
```

So `filter` is only called when `condition` is true

frontend/src/components/Cases/CasesSection.tsx
harrel56: That IDE error appears because you wrote this:
```ts
const caseResults: (CaseResult | undefined)[] = ...
```
I guess the real question here is: can `implResult.cases.get(seq)` return undefined? I'm not sure about this, but if it really can be undefined I would suggest filtering out cases where it's undefined, so:
```ts
          .filter(implResult => implResult.cases.has(seq))
          .map((implResult) => implResult.cases.get(seq)!)
          .flat();
```

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: done!

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: I am not sure about it. Because I have never used it. Though let me try once.

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
Swarnendu0123: Done!

frontend/src/components/Cases/CaseItem.tsx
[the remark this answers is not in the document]
Swarnendu0123: Oh! sorry for that. Fixed in latest commit,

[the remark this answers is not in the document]
harrel56: Seems like it's still the same?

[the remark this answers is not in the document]
Swarnendu0123: Oh sorry, I have memorized the function again, I will push a commit by tomorrow.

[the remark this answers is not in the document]
Swarnendu0123: I guess now it is done!

frontend/src/components/Cases/CasesSection.tsx
Swarnendu0123: done!

Swarnendu0123:
> Just to summarize slack discussion: The only blocking issue is [#939 (comment)](https://github.com/bowtie-json-schema/bowtie/pull/939#discussion_r1513542530), if it's sorted out and the 2 mentioned cases pass, I think we can finally merge it. Also the lib for highlighting (https://www.npmjs.com/package/react-highlight-words) looks fine :)

Yes! it is ready for merging! @Julian (Just mentioning you for the final review and merge)

Julian:
@Swarnendu0123 Is that blocking issue resolved then?

Swarnendu0123:
> @Swarnendu0123 Is that blocking issue resolved then?

No, this the blocking issue needs https://www.npmjs.com/package/react-highlight-words, this PR is already have changes ~190 lines. It will be hard to review if I implement it in the same PR. I will implement the lib in subsequent PR(s). We can merge it for now.

Julian:
The word "blocking" means "this isn't usable without it" :) so that doesn't sound right, but I'll have a look myself at what state this is in at this point.

PR #942, opened 2024-02-26
Hide the implementation version column in mobile UI

### LINKED ISSUE #935: Hide the implementation version column in the mobile UI
For each implementation, we show a column containing its version. This is definitely useful information, as it tells a user what version of the library we're testing.

But it's definitely *less* useful than the actual test results.

On mobile right now, one sees:

Screenshot 2024-02-24 at 9 55 24 PM

where a lot of horizontal space is taken up between the name and version columns. We should instead condense the table horizontally on mobile so that the version either collapses into the name column (if that's easy to do) or else not show it entirely.

Sahil-Shadwal:
Fixes #935 

image

Removed implementation version column from mobile view .

### DIFF
--- a/frontend/src/components/Summary/ImplementationRow.tsx
+++ b/frontend/src/components/Summary/ImplementationRow.tsx
@@ -32,7 +32,7 @@ const ImplementationRow = ({
           {mapLanguage(implementation.metadata.language)}
         </small>
       </th>
-      <td className="align-middle">
+      <td className="align-middle d-none d-sm-table-cell">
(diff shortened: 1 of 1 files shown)


Julian:
Thanks, appreciated, looks great! Merging.

If you care for a follow up, it looks like the headings are misaligned in your screenshot (and I see that's the case on mobile). The "errors" heading is supposed to be a label for the first column, and et cetera.

Opening a bug and sending a PR for that would be another good contribution!

PR #948, opened 2024-02-26
Added PR and Issue Template

Swarnendu0123:
Added PR and Issue Template under `.github`, which will help contributors to Explain about their PR and Issues. And also help Maintains for validation of any PR and issu.

----
📚 Documentation preview 📚: https://bowtie-json-schema--948.org.readthedocs.build/en/948/


Julian:
Thanks! I'm not a huge fan of these personally honestly -- specifically when there's not something "special" / unique to say in them about the project itself. The reason is pretty simple and demonstrated by what you did here -- namely that nearly all of what's in them is the same for any project you'll contribute to, so it's just boilerplate.

We already have [contributor's guide](https://docs.bowtie.report/en/stable/contributing/). If there's something missing from there, we should add it, but otherwise yeah, I don't think this is necessary honestly, it doesn't have huge benefits in my experience with other projects.

Issue #945, opened 2024-02-26
Issue Description:
When a user clicks on the copy button, the button state changes to indicate that the copy action has been performed (e.g., the button icon changes to "Copied!"). However, the button remains in the "Copied!" state indefinitely, which may confuse users as they might expect the button to revert to its original state after a certain period.


Swarnendu0123:
![IMG_20240226_171631.jpg](https://github.com/bowtie-json-schema/bowtie/assets/94161758/57d4606c-bf7d-4ace-996d-33c7fa3ed3c2)

Expected Behavior:
After the copy button is clicked and the copy action is performed, the button should display the "Copied!" state for a brief period (e.g., 3 seconds) before reverting back to its original state (e.g., displaying "Copy" as the button text).

Proposed Solution:
Implement a mechanism to reset the copy button state to its initial state after a specified duration (e.g., 3 seconds) following a click event. This can be achieved using JavaScript's `setTimeout` function to trigger the state reset after the specified time period elapses.

Swarnendu0123:
Can I start working on this issue?

Julian:
Sounds good, sure.

PR #949, opened 2024-02-26
Copy button is now going to it's initial state, 7 seconds after copying

Swarnendu0123:
Fixes: https://github.com/bowtie-json-schema/bowtie/issues/945

Copy button is now going to it's initial state, 3 seconds after copying

----
📚 Documentation preview 📚: https://bowtie-json-schema--949.org.readthedocs.build/en/949/

### DIFF
--- a/frontend/src/components/CopyToClipboard.tsx
+++ b/frontend/src/components/CopyToClipboard.tsx
@@ -13,7 +13,13 @@ function CopyToClipboard({ textToCopy }: CopyProps) {
 
   const handleCopy = () => {
     navigator.clipboard.writeText(textToCopy).then(
-      () => setCopied(true),
+      () => {
+        setCopied(true);
+        // Reset 'copied' state after 5 seconds
+        setTimeout(() => {
+          setCopied(false);
+        }, 3000);
+      },
       () => setCopied(false),
     );
   };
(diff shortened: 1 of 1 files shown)


frontend/src/components/CopyToClipboard.tsx
```diff
@@ -13,7 +13,13 @@ function CopyToClipboard({ textToCopy }: CopyProps) {
 
   const handleCopy = () => {
     navigator.clipboard.writeText(textToCopy).then(
-      () => setCopied(true),
+      () => {
+        setCopied(true);
+        // Reset 'copied' state after 3 seconds
```
Julian, line 18: This comment doesn't seem to add much.

Arguably the same for the comment below.

Swarnendu0123: Okey, I have removed the comment

Swarnendu0123:
Maybe we can change the timing to 5 Second or 10 Seconds which will give a smooth experience to user.

Julian:
I see no difference in behavior here in Firefox, it immediately transitions.

Swarnendu0123:
> I see no difference in behavior here in Firefox, it immediately transitions.

But, how can it be possible, I have increases 3000 to 7000, it should be stable now.

Julian:
Have you tried it? Does it work for you?

Julian:
Is that Firefox? Here I'm on 124.0b4 which looks to be the latest (beta).

Swarnendu0123:
> Is that Firefox? Here I'm on 124.0b4 which looks to be the latest (beta).

No it is Microsoft edge, but I think you are looking at the old commit, I have pushed a new commit.

Had you considered that?

Julian:
Yes, I was on your new commit.

Julian:
Feel free to resend this when you get it working.

Issue #943, opened 2024-02-26
Instance copy button overflowing in Mobile mode


Swarnendu0123:
![IMG_20240226_113520.jpg](https://github.com/bowtie-json-schema/bowtie/assets/94161758/47a44fef-0f52-4dd4-8cf3-5dad27330894)

Swarnendu0123:
Can I get this issue assigned?

Julian:
Go for it, thanks.

PR #950, opened 2024-02-26
Instance copy button is now not overflowing in Mobile mode

Swarnendu0123:
Fixes: #943

Instance copy button is now not overflowing in Mobile mode
Screenshot (100)

----
📚 Documentation preview 📚: https://bowtie-json-schema--950.org.readthedocs.build/en/950/

### DIFF
--- a/frontend/src/components/Cases/SchemaDisplay.tsx
+++ b/frontend/src/components/Cases/SchemaDisplay.tsx
@@ -33,7 +33,7 @@ const SchemaDisplay = ({
         </div>
         <div className="col-4 border-start ps-0">
           <div className="d-flex align-items-center highlight-toolbar ps-3 pe-2 py-1 border-0 border-top border-bottom">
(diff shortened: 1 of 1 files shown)


Julian:
Thanks -- that indeed looks better, but also a bit cramped. Specifically it looks like the  copy button should be off further to the right (with some padding or something). In the screenshot it's almost on top of the word "Instance". Can you have a look at that as well?

Swarnendu0123:
> Thanks -- that indeed looks better, but also a bit cramped. Specifically it looks like the copy button should be off further to the right (with some padding or something). In the screenshot it's almost on top of the word "Instance". Can you have a look at that as well?

Ya sure! let me do that!

Swarnendu0123:
Screenshot (101)
Is it looks good now?

Julian:
Yes seems likely!

Julian:
Nice, thanks!

PR #951, opened 2024-02-26
Add unit tests for test success icons

SimonDMC:
### Summary
This unit test makes sure `<CaseResultSvg />` renders the right icon for each result state:
- `CheckCircleFill` if `valid: true`
- `XCircleFill` if `valid: false`
- `ExclamationOctagon` if `state: "skipped"`
- `ExclamationOctagon` if `state: "errored"`

Ideally, the unit tests should also check for the class name, because it currently can't distinguish between a skipped and an errored test case and between a successful and a failed test case. However, that might require an additional DOM testing library since the one I went for doesn't seem to support it.

### Implementation notes
- I added the `react-test-renderer` library for checking if a component contains another component. Virtually every other DOM testing library only supports checking for raw HTML elements, though since Bowtie is using another library, `react-bootstrap-icons`, for its icons, checking for direct HTML didn't make sense in this case.
- I changed one of the parameters in `tsconfig.json` to get Jest to recognize the React components in the test file, though I'm pretty sure that shouldn't break anything.


Julian:
Hey! This looks great. I haven't reviewed it in detail, I'll try to do that this afternoon now -- one quick note since it's the first thing I see.

I don't know what the normal convention is for React apps (nor do I have any personal preference). What I saw from my quick pass looking was what we have now, namely `foo.ts` gets tested by an adjacent `foo.test.ts`. As I say, I'm not a JS/TS developer with existing strong opinions there, but I would like consistency -- if you think `__tests__` is better can you say in very short what you like about it, and then if so we should move the existing tests to that structure. Otherwise if not I think we should keep the current structure and move this one to adjacent to what it's testing.

But yeah looks like a nice example, will stare at it and leave any other comments.

frontend/src/components/Cases/__tests__/CaseResultSvg.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { create } from "react-test-renderer";
+import CaseResultSvg from "../CaseResultSvg";
+import { CaseResult } from "../../../data/parseReportData";
+import {
+  CheckCircleFill,
+  XCircleFill,
+  ExclamationOctagon,
+} from "react-bootstrap-icons";
+
+test("Case Result Icon Successful Expected True", () => {
```
Julian, line 11: The structure of these tests are identical, so they look like they could be done as some sort of parameterized test where you specify the test case data and expected results in a list of 4 things and then just run the same test over the 4 inputs.
Can you see whether vitest/whatever we're using has something for that, assuming you agree.

SimonDMC:
> As I say, I'm not a JS/TS developer with existing strong opinions there, but I would like consistency -- if you think __tests__ is better can you say in very short what you like about it, and then if so we should move the existing tests to that structure.

I took the `__tests__` convention from the [jest documentation](https://jestjs.io/docs/configuration), but from what I can tell it doesn't have any actual benefit. I agree that sticking to one convention makes more sense, and since there already are some `*.test.ts` tests I'll change it to that.

frontend/src/components/Cases/__tests__/CaseResultSvg.test.tsx
SimonDMC: Yeah, makes sense to me. I don't think there's a built-in way of doing that but I can definitely write a wrapper for it tomorrow if you think it's better - I'm neutral on this.

Julian: `describe.each` appears to be the way: https://vitest.dev/api/#describe-each

SimonDMC: Oh, I see, my bad. For the record I think we're using jest, not vitest, but the `describe.each` syntax is the same in both. Thanks!

PR #958, opened 2024-02-27
Removed `urijs` and used `URL`

### LINKED ISSUE #952: Remove urijs and use `URL`
We use urijs for URIs in the browser, but modern browsers seem to have a `URL` type already built in.

Unless there's some reason not to, we should likely switch to using that

Swarnendu0123:
# In Progress: Fixes #952 
Removed `urijs` and used `URL`

----
📚 Documentation preview 📚: https://bowtie-json-schema--958.org.readthedocs.build/en/958/

### DIFF
--- a/frontend/src/data/Dialect.ts
+++ b/frontend/src/data/Dialect.ts
@@ -14,7 +14,7 @@ export default class Dialect {
     path: string,
     prettyName: string,
     uri: string,
-    firstPublicationDate: Date
+    firstPublicationDate: Date,
   ) {
     if (Dialect.all.has(path)) {
       throw new DialectError(`A "${path}" dialect already exists.`);
@@ -40,7 +40,7 @@ export default class Dialect {
   static newest_to_oldest(): Dialect[] {
     return Array.from(Dialect.known()).sort(
       (d1: Dialect, d2: Dialect) =>
-        d2.firstPublicationDate.valueOf() - d1.firstPublicationDate.valueOf()
+        d2.firstPublicationDate.valueOf() - d1.firstPublicationDate.valueOf(),
     );
   }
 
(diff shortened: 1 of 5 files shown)


Swarnendu0123:
I see some conflicts are there, let me resolve them!

Julian:
Possibly, but your CI also doesn't pass as-is.

Swarnendu0123:
> Possibly, but your CI also doesn't pass as-is.

I think it's because I have changed the version     `"@typescript-eslint/parser": "^7.0.0",`

frontend/src/data/Dialect.ts
```diff
@@ -32,9 +27,9 @@ export default class Dialect {
     this.firstPublicationDate = firstPublicationDate;
   }
 
-  async fetchReport(baseURI: URI = siteURI) {
-    const url = baseURI.clone().filename(this.path).suffix("json").href();
-    const response = await fetch(url);
+  async fetchReport(baseURL: URL = siteURL) {
+    const url = new URL(this.path + ".json", baseURL.href);
```
harrel56, line 31: I believe this behaves differently in some cases:
- old way: `https://harrel56.github.io/bowtie` resolves to  `https://harrel56.github.io/bowtie/draftX`
- new way:  `https://harrel56.github.io/bowtie` resolves to  `https://harrel56.github.io/draftX` - which is just wrong

Possibly there are differences in other parts too. Certainly, before merging it should be checked if GH pages on forks works correctly

Julian:
Going to close this as I suspect it's equally easy for someone picking this up to start from scratch than to reference this, but feel free to send a new one if you get back to this.

(Please only do so after you get your other PR in mergeable state.)

PR #965, opened 2024-02-28
Fixed current Navbar and Container Related Issue.

nomandhoni-cs:
### Navbar Collapse in Mobile Device
Collapsed

### Navbar Open in Mobile Device

Frame 1

### Old Container 
image

### New Expanded Container

Screenshot_57

----
📚 Documentation preview 📚: https://bowtie-json-schema--965.org.readthedocs.build/en/965/

### DIFF
--- a/frontend/src/components/NavBar.tsx
+++ b/frontend/src/components/NavBar.tsx
@@ -36,20 +36,18 @@ const NavBar = () => {
             <img src={logo} alt="Bowtie Logo" width="128px" />
           </Link>
           <div className="d-flex align-items-center">
-            <div className="d-lg-none">
-              <Button className="btn btn-secondary border-0 me-1">
-                <Link
-                  className="nav-link"
-                  to="https://docs.bowtie.report/"
-                  target="blank"
-                >
-                  <Book size={20} />
-                </Link>
-              </Button>
+            <div className="d-lg-none d-flex justify-content-between align-items-center me-2">
+              <Link
+                className="nav-link border border-primary rounded-3 d-inline-block p-2 me-1"
+                to="https://docs.bowtie.report/"
+                target="_blank"
+              >
+                <Book size={20} />
+              </Link>
               <button
-                className={`btn ${
+                className={`btn d-flex align-items-center justify-content-center ${
                   isDarkMode ? "btn-light" : "btn-secondary"
-                } rounded me-1`}
+                } rounded me-1 p-2`}
                 onClick={() => toggleDarkMode!()}
               >
                 {isDarkMode ? <MoonStarsFill size={20} /> : <Sun size={20} />}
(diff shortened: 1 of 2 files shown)


Julian:
All of this looks great until the last screenshot!

What made you expand the width? From what I've heard, very wide sites are bad for accessibility, which is why nearly every site you visit does not look like that!

nomandhoni-cs:
@Julian, Thanks for the catch! Current container width of 830px seemed a bit narrow to me. In the design industry, 1200-1300px is more common for large screens.

To ensure consistency with design standards, I made the container wider. Do you think keeping this change aligns better, or should I revert it to the original size?


## Container Size in Bootstrap
image

## Container size in Tailwind CSS
image

Julian:
You're confusing container size with the content inside of it I believe.

Here for example on GitHub.com, on my big monitor, the main content of pages is still only 871 pixels wide.

It's true the container is larger -- that's because they have sidebars with additional content. I think we too should have sidebars soon, but *then* we can expand the size of the container for the sidebar. The actual main content of the page itself though should be narrower. At least that's what I've seen -- if you do care to you could educate us both by finding some design guidelines!

But please don't mix unrelated things in the same PR in that case -- none of the other stuff seems related to this change, so you can do it in separate ones!

nomandhoni-cs:
Oops, my mistake. We had a bunch of empty space, which is why I decided to beef up the existing content.

I undo my Commit. I discard the changes in the `DialectReportView.tsx`.

Also, I'm curious, do you have any ideas on how you want the "New Bowtie Test Case Widgets" to look? Any specific designs or styles you're thinking about?

frontend/src/components/NavBar.tsx
```diff
@@ -30,23 +31,44 @@ const NavBar = () => {
...
-            aria-label="Toggle navigation"
-          >
-            <span className="navbar-toggler-icon"></span>
-          </button>
+          <div className="d-flex align-items-center">
+            <div className="d-lg-none">
+              <Button className="btn btn-secondary border-0 me-1">
+                <Link
+                  className="nav-link"
+                  to="https://docs.bowtie.report/"
```
Julian, line 43: Clicking this link doesn't appear to work, possibly because you have `target="blank"` below (as opposed to `_blank`), but there's no reason for `_blank` either, you can just remove that entirely.

Julian: Oh, it seems `target="blank"` is present in the current version, and that one works. So not sure what the reason is, but yeah have a look.

Julian: Oh very strange, the behavior is even weirder, I'm not sure what's happening. Here it works when clicked the first time, and not after that:


https://github.com/bowtie-json-schema/bowtie/assets/329822/49a613dc-991b-4b90-bdbf-804320f154de

When nothing is happening I'm still mashing clicking the button, but nothing is happening. I'd try fixing the `target` to `_blank`, or removing it. But let me know if you can reproduce at all.

nomandhoni-cs: Thanks a lot, It was tricky to figure out the problem, why is this doing weird behavior. Now every thing works fine, as expected

Frame 2 (4)

## Why It was doing weird behavior

- As you can See in the above Image Link was only in the text "Docs". Icon was outside of the link. that's why this was doing weird behavior.
- Also I used the link inside of a `button`. That's why when you clicked on the white space like below image, The Link wasn't creating a new window

Frame 2 (5)

### Also I have learned a new thing today
image

- If you want to know more you can pay a visit to https://css-tricks.com/targetblank/

# Is there anything I can do?

Julian: Very interesting, well done, thanks for sharing and laying things out clearly.

PR #967, opened 2024-02-28
[Docs] Added Instructions for building the documentations 

JeelRajodiya:
# This new sections is added to build the documentation 
image

@Julian, I think there is some room for improvement, but I need your feedback on how to improve it 

----
📚 Documentation preview 📚: https://bowtie-json-schema--967.org.readthedocs.build/en/967/

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -115,7 +115,7 @@ Feel free to file issues or pull requests as well if you wish something was bett
 Building the Documentation Locally
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The code for the documentation is in the ``docs/`` directory of our :org:`github repository <bowtie>`.
+The code for the documentation is in the :gh:`docs directory of the Bowtie repository <tree/main/docs>`.
 
 You can build the documentation locally by running:
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -119,7 +119,7 @@ Feel free to file issues or pull requests as well if you wish something was bett
 
 The code for the documentation is in the :gh:`docs directory of the Bowtie repository <tree/main/docs>`.
 
-You can build the documentation locally by running the following command from the root of the repository:
+You can build the documentation locally by running the following command from the root of the repository
 
 .. code:: sh
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -119,7 +119,7 @@ Feel free to file issues or pull requests as well if you wish something was bett
 
 The code for the documentation is in the :gh:`docs directory of the Bowtie repository <tree/main/docs>`.
 
-You can build the documentation locally by running on the root of the repository:
+You can build the documentation locally by running following the command on the root of the repository:
 
 .. code:: sh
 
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -112,6 +112,19 @@ Documentation is written in `Sphinx-flavored reStructuredText <sphinx:index>`, s
 
 Feel free to file issues or pull requests as well if you wish something was better documented, as this too can help prioritize.
 
+Building the Documentation Locally
```
Julian, line 115: I think this seems like it can go combined in the previous section rather than with its own heading maybe.

As in:

```
Documentation is written in `Sphinx-flavored reStructuredText <sphinx:index>`, so you'll want to at least have a quick look at the `tutorial <sphinx:tutorial>` if you have never written documentation using it.

The code for the documentation is in the ``docs/`` directory of our :org:`github repository <bowtie>`.

etc.
```

Perhaps test out combining them. Will leave another comment or two about the specific content.

docs/contributing.rst
```diff
@@ -112,6 +112,19 @@ Documentation is written in `Sphinx-flavored reStructuredText <sphinx:index>`, s
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The code for the documentation is in the ``docs/`` directory of our :org:`github repository <bowtie>`.
```
Julian, line 118: There's already a `gh` role for the repository itself, so you can do

```suggestion
The code for the documentation is in the :gh:`docs directory of the Bowtie repository <tree/main/docs>`.
```

docs/contributing.rst
```diff
@@ -112,6 +112,19 @@ Documentation is written in `Sphinx-flavored reStructuredText <sphinx:index>`, s
...
+
+You can build the documentation locally by running:
+
+.. code:: sh
+
+    $ nox -s "docs(dirhtml)" -- dirhtml
+
+This will generate ``/dirhtml`` directory containing the documentation in html.
```
Julian, line 126:
```suggestion
which will generate a directory called ``dirhtml`` containing the built HTML.
```

Julian (changes requested):
Thanks for improving this! Left some small comments.

JeelRajodiya:
@Julian Updated!
image

JeelRajodiya:
Updated version 
image

docs/contributing.rst
```diff
@@ -113,6 +113,21 @@ Documentation is written in `Sphinx-flavored reStructuredText <sphinx:index>`, s
 Feel free to file issues or pull requests as well if you wish something was better documented, as this too can help prioritize.
 
 
+**Building the Documentation Locally**
```
Julian, line 116: If you want another layer of heading (which seems OK), you should use a heading rather than just bolding the text.

The syntax for that is:

```suggestion
Building the Documentation Locally
===========================
```

but I'll apply that change manually and merge.

Thanks again!
```

Julian:
Oh, also, for what it's worth, there's no need to screenshot for docs changes, the PR will automatically include a link with the generated changes. You can see yours which was automatically edited into the PR comment.

Perhaps that itself is worth mentioning in a sentence in this same section.

Issue #947, opened 2024-02-26
Clicking on links in the navbar doesn't proper taking to the route of that path i.e some of the 


Akshaybagai52:
go to bowtie homepage :- https://bowtie.report/ 
![image](https://github.com/bowtie-json-schema/bowtie/assets/100753870/59bf3449-717d-467f-b41b-5bf833a79ebd)

and then click on any of the links present in the navbar you will see that some part of the route getting behind the navbar as shown in the image
![image](https://github.com/bowtie-json-schema/bowtie/assets/100753870/1cae72fe-fa2b-4413-a518-13b0ac8da637)

Expected behaviour :
![image](https://github.com/bowtie-json-schema/bowtie/assets/100753870/3b865d0d-9b2f-40da-8778-9257ea930071)

I would like to work on this issue

Julian:
I thought I'd filed this already but I guess not -- yes go for it, thanks!

Akshaybagai52:
> I thought I'd filed this already but I guess not -- yes go for it, thanks!

Actually you've made this issue for website repo in json-schema

Abhinav-Saxena-dev:
@Akshaybagai52 The required components are scrolling into view as expected, however, the issue is due to the nav-bar hiding the required part of the view that we wish for the user to see.

The fix will be adding an offset to the scroll position. This accounts for the fixed navbar's height and ensures the target section is properly visible.

Can you please assign this to me?

CC: @Julian

Abhinav-Saxena-dev:
Oh! Please nevermind. I overlooked it is already been worked upon. My apologies.

Julian:
All good, there should be other issues still marked as good first issue but if not feel free to ask in Slack (or you may even find some yourself by playing with the site), there's plenty more to be done!

Akshaybagai52:
Hey @Julian 
I'm not able to find any gloabl.css file as well as there aren't any css files for individual components i.e. "RunInfoSection", "SummarySection", "CasesSection". Right now what I did is to make new css file for DialectReportView.css in which I'm saving the className and then will put in the respected components so is this good ?
![image](https://github.com/bowtie-json-schema/bowtie/assets/100753870/05283c36-ad5d-4441-b6b2-d96eb89be268)

PR #968, opened 2024-02-28
fix: corrected the underflow scroll in anchor tags of navbar

ashmit-coder:
Fixes : #947 

As per the given issue the Navbar anchor tags work as expected and the sticky navbar no more affects them.
It has also been maid responsive for the mobile view with media queries

----
📚 Documentation preview 📚: https://bowtie-json-schema--968.org.readthedocs.build/en/968/

### DIFF
--- a/frontend/src/global.css
+++ b/frontend/src/global.css
@@ -1,9 +1,9 @@
-html{
-    scroll-padding-top: 6rem;
+html {
+  scroll-padding-top: 6rem;
 }
 
 @media screen and (max-width: 991px) {
-    html{
-        scroll-padding-top: 19rem; 
-    }
+  html {
+    scroll-padding-top: 19rem;
+  }
 }
--- a/frontend/src/global.css
+++ b/frontend/src/global.css
@@ -1,11 +1,9 @@
-html {
-  scroll-behavior: smooth;
-  scroll-padding-top: 5.5rem;
+html{
+    scroll-padding-top: 6rem;
 }
 
 @media screen and (max-width: 991px) {
-  html {
-    scroll-behavior: smooth;
-    scroll-padding-top: 18.5rem;
-  }
+    html{
+        scroll-padding-top: 19rem; 
+    }
 }
--- a/frontend/src/global.css
+++ b/frontend/src/global.css
@@ -1,11 +1,11 @@
-html{
-    scroll-behavior: smooth;
-    scroll-padding-top: 5.5rem;
+html {
+  scroll-behavior: smooth;
+  scroll-padding-top: 5.5rem;
 }
 
 @media screen and (max-width: 991px) {
-    html{
-        scroll-behavior: smooth;
-        scroll-padding-top: 18.5rem; 
-    }
+  html {
+    scroll-behavior: smooth;
+    scroll-padding-top: 18.5rem;
+  }
 }
--- a/frontend/src/global.css
+++ b/frontend/src/global.css
@@ -1,0 +1,11 @@
+html{
+    scroll-behavior: smooth;
+    scroll-padding-top: 5.5rem;
+}
+
+@media screen and (max-width: 991px) {
+    html{
+        scroll-behavior: smooth;
+        scroll-padding-top: 18.5rem; 
+    }
+}
--- a/frontend/src/components/Cases/CasesSection.css
+++ b/frontend/src/components/Cases/CasesSection.css
@@ -1,0 +1,8 @@
+#cases{
+    scroll-margin-top: 5.5rem;
+}
+@media screen and (max-width: 991px) {
+    #cases {
+        scroll-margin-top: 18.5rem; 
+    }
+}
--- a/frontend/src/components/Cases/CasesSection.css
+++ b/frontend/src/components/Cases/CasesSection.css
@@ -1,8 +1,0 @@
-#cases{
-    scroll-margin-top: 5.5rem;
-}
-@media screen and (max-width: 991px) {
-    #cases {
-        scroll-margin-top: 18.5rem; 
-    }
-}
--- a/frontend/src/components/Cases/CasesSection.tsx
+++ b/frontend/src/components/Cases/CasesSection.tsx
@@ -1,7 +1,7 @@
 import { ReportData } from "../../data/parseReportData";
 import CaseItem from "./CaseItem";
 import { Accordion } from "react-bootstrap";
-import "./CasesSection.css"
+
 const CasesSection = ({ reportData }: { reportData: ReportData }) => {
   const implementations = Array.from(reportData.implementations.values());
   return (
--- a/frontend/src/components/Cases/CasesSection.tsx
+++ b/frontend/src/components/Cases/CasesSection.tsx
@@ -1,7 +1,7 @@
 import { ReportData } from "../../data/parseReportData";
 import CaseItem from "./CaseItem";
 import { Accordion } from "react-bootstrap";
-
+import "./CasesSection.css"
 const CasesSection = ({ reportData }: { reportData: ReportData }) => {
   const implementations = Array.from(reportData.implementations.values());
   return (
--- a/frontend/src/components/RunInfo/RunInfoSection.css
+++ b/frontend/src/components/RunInfo/RunInfoSection.css
@@ -1,0 +1,9 @@
+#run-info{
+    scroll-margin-top: 5.5rem;
+}
+
+@media screen and (max-width: 991px) {
+    #run-info {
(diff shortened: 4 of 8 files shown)


Julian:
Thanks! We use Bootstrap, please have a look and use whatever the Bootstrap way to fix this would be rather than manually messing with CSS!

ashmit-coder:
> Thanks! We use Bootstrap, please have a look and use whatever the Bootstrap way to fix this would be rather than manually messing with CSS!

Okay sir will look into it. I am not quite sure if there is a way to achieve it through bootstrap. Any guidance would be helpfull.

Julian:
I know for sure that manually messing with media queries isn't the right way, which is how I knew this was wrong, bootstrap uses classes for that. In terms of the underlying issue, I don't know, but I'd start by reading the Navbar part of the Bootstrap docs and seeing what we do wrong.

ashmit-coder:
Hey @Julian I looked in on ways we could achieve the same with help of bootstrap. But there does not seem to be any results.

The issue at hand occurs due to the sticky nature of the navbar component. So the proposed solution was to add scroll margin for the anchors.

As of now there does not seem to be a way to manipulate the scroll margin property in bootstrap.

As for the proposed solution the media query is used for the two modes that the app uses lg ,that is large displays and md with sm that is tablet and mobile displays.

I will look some more if anything can be for it. 
If not then we may close the PR if you suggest.

Julian:
Always such strange things missing from these frameworks. Fair enough, but then I really don't want to do this manually for each component. What's a way that we can automatically apply this style to any component that appears on pages underneath the navbar without manually remembering we need to copy paste what you have here?

ashmit-coder:
> Always such strange things missing from these frameworks. Fair enough, but then I really don't want to do this manually for each component. What's a way that we can automatically apply this style to any component that appears on pages underneath the navbar without manually remembering we need to copy paste what you have here?

There are two ways in which may achieve this.

1. We can just apply the property to index.css on all div. (This will be a very brute force soultion and may cause issues in the future even if it will sound good currently and works after testing.)

2. We can create a class 

        .anchor-pad{
            scroll-margin-top: 5.5rem;
        }
        
        @media screen and (max-width: 991px) {
            #anchor-pad {
                scroll-margin-top: 18.5rem; 
            }
        } 

and import it in index.css and use this anchor-pad in the className of whichever div we want to place under the navbar.

Now that I think this would have been a way better approach than my current one. :sweat_smile:

JeelRajodiya:
> Always such strange things missing from these frameworks. Fair enough, but then I really don't want to do this manually for each component. What's a way that we can automatically apply this style to any component that appears on pages underneath the navbar without manually remembering we need to copy paste what you have here?

## A simple solution here would be to add a global property for the html tag ( inspired from [this PR](https://github.com/json-schema-org/website/pull/368) )
image

check out this branch for testing-> https://github.com/JeelRajodiya/bowtie/tree/navbar-underflow

## Demo
output

Julian:
That looks significantly easier to manage, yes, please use that.

ashmit-coder:
> > Always such strange things missing from these frameworks. Fair enough, but then I really don't want to do this manually for each component. What's a way that we can automatically apply this style to any component that appears on pages underneath the navbar without manually remembering we need to copy paste what you have here?
> 
> ## A simple solution here would be to add a global property for the html tag ( inspired from [this PR](https://github.com/json-schema-org/website/pull/368) )
> image
> 
> ## Demo
> ### check out this branch for testing-> https://github.com/JeelRajodiya/bowtie/tree/navbar-underflow
> output [ output ](https://private-user-images.githubusercontent.com/63534268/309000964-b5de15d5-5ca0-4b4f-b983-b42910c9a7dc.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDkyMzMwODcsIm5iZiI6MTcwOTIzMjc4NywicGF0aCI6Ii82MzUzNDI2OC8zMDkwMDA5NjQtYjVkZTE1ZDUtNWNhMC00YjRmLWI5ODMtYjQyOTEwYzlhN2RjLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjI5VDE4NTMwN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTkyMWYwMDYwMmNmOTZmZjliMDU0NDk3MDI3OWQ1MzJmZjE5NGZmMWQ5ZmI3NTA2NWVlYWM3NDJiZDZhN2I5MDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.n3eBIuR68z3fUVRzhDPI289Xb77gBiMoEseL8xsm-TI) [ ](https://private-user-images.githubusercontent.com/63534268/309000964-b5de15d5-5ca0-4b4f-b983-b42910c9a7dc.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDkyMzMwODcsIm5iZiI6MTcwOTIzMjc4NywicGF0aCI6Ii82MzUzNDI2OC8zMDkwMDA5NjQtYjVkZTE1ZDUtNWNhMC00YjRmLWI5ODMtYjQyOTEwYzlhN2RjLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjI5VDE4NTMwN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTkyMWYwMDYwMmNmOTZmZjliMDU0NDk3MDI3OWQ1MzJmZjE5NGZmMWQ5ZmI3NTA2NWVlYWM3NDJiZDZhN2I5MDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.n3eBIuR68z3fUVRzhDPI289Xb77gBiMoEseL8xsm-TI)

Thank You very much @JeelRajodiya :). Will use this.

JeelRajodiya:
> Thank You very much @JeelRajodiya :). Will use this or if you want you can make your own PR and I will close this one.

@ashmit-coder
Please continue with your PR

frontend/src/global.css
```diff
@@ -0,0 +1,11 @@
+html {
+  scroll-behavior: smooth;
+  scroll-padding-top: 5.5rem;
+}
+
+@media screen and (max-width: 991px) {
+  html {
+    scroll-behavior: smooth;
```
Julian, line 8: Is this not going to still be covered by the above?

ashmit-coder: No sir as in the mobile view the padding needed is different 

Desktop-view:- 
Screenshot from 2024-03-01 00-48-56

Tab and Mobile view

Screenshot from 2024-03-01 00-48-45

Julian: The padding is different but I was asking about the `scroll-behavior` which you have in both sections, and which presumably should get inherited from the top one.

But considering that even more carefully, let's just remove that entirely, I'm not a fan of silly animations if they make things slower, let's just leave it at the default behavior and just add the padding.

Also please add a little extra padding so that borders don't end up at the top -- say, `6rem` and `19rem` look like they leave the little extra space at the top which looks a bit better I think.

ashmit-coder: @Julian Gotcha sir.

Julian:
Great! Thanks again!

Issue #959, opened 2024-02-27
Add "Badges" pills to the implementation pages


Julian:
#929 added badges to various portions of the implementation page (e.g. https://bowtie.report/#/implementations/java-jsonschemafriend). Those badges are placed close to the data they report on (so for example, the dialect compliance badge appears at the bottom of the page where numbers appear).

From a UX perspective, that means it's easy for an implementer who's looking at data for a certain metric to then get a badge for that data.

From another UX perspective however, it means an implementer has no way of answering "what are *all* the badges I could possibly show off for my implementation".

We should have a single spot on the page for copying / embedding badges which when clicked opens a set of [pills](https://getbootstrap.com/docs/5.3/components/navs-tabs/#pills) containing *all badges in one spot*.

The purpose of these pills should be for an implementer looking to embed the badge somewhere, so they should contain not just a preview of the badge but also the URL someone needs, or the HTML or Markdown they'd want to embed.

Look at the shields.io component at https://shields.io/badges (on the right side) which does this well.

sd1p:
Can I work on this issue?

Julian:
Go for it!

PR #969, opened 2024-02-28
Added functionality for users to embed badges

sd1p:
Implementing Issue https://github.com/bowtie-json-schema/bowtie/issues/959

https://github.com/bowtie-json-schema/bowtie/assets/43758514/4dccdc35-203c-4e69-a5a9-443ba02805ce

Current Skeleton for the UI
image

----
📚 Documentation preview 📚: https://bowtie-json-schema--969.org.readthedocs.build/en/969/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,8 @@ jobs:
       fail-fast: false
       matrix:
         noxenv: ${{ fromJson(needs.list.outputs.noxenvs) }}
+        include:
+          - noxenv: app
 
     steps:
       - uses: actions/checkout@v4
--- a/noxfile.py
+++ b/noxfile.py
@@ -132,7 +132,7 @@ def build(session):
             )
 
 
-@session(tags=["build"])
+@session(default=False, tags=["build"])
 def app(session):
     """
     Build a PyApp which will run Bowtie.


Julian:
Nice start! You've got some immediate conflicts. It looks like when you started you started off of an old commit, there were some changes on main that looks like you're missing. Hopefully shouldn't be hard to pull those in.

In terms of the design -- nice mockup. I think my first reaction is it's too prominent though, this functionality is really for implementers, who are only a secondary user of that page. I think a good thing to try is to put it in the *heading* of the `Compliance` card -- so a button aligned right in the heading of the card which when clicked, opens a dropdown with this widget. Let me know if that makes sense!

sd1p:
> Nice start! You've got some immediate conflicts. It looks like when you started you started off of an old commit, there were some changes on main that looks like you're missing. Hopefully shouldn't be hard to pull those in.
> 
> In terms of the design -- nice mockup. I think my first reaction is it's too prominent as an accordion though, this functionality is really for implementers, who are only a secondary user of that page. I think a good thing to try is to put it in the _heading_ of the top card on the page -- so a button aligned right in the heading of the card (so next to the name of the implementation) called "Bdages", which when clicked, opens a dropdown with this widget. Let me know if that makes sense!

This sounds good. I have created a new PR and I will work on that.

PR #974, opened 2024-02-29
fix: Misaligned heading in Mobile UI

### LINKED ISSUE #973: Misaligned heading in Mobile UI
307677133-4a81085e-7d68-498f-b5f9-7720e639bcd4

The error heading is misaligned in Mobile View . The "errors" heading is supposed to be a label for the first column, and et cetera.

Sahil-Shadwal:
Fixes #973 

Resolves  the misaligned error heading , using React useState & useEffect.
Screenshot from 2024-02-29 11-58-38

----
📚 Documentation preview 📚: https://bowtie-json-schema--974.org.readthedocs.build/en/974/

### DIFF
--- a/frontend/src/components/Summary/SummaryTable.tsx
+++ b/frontend/src/components/Summary/SummaryTable.tsx
@@ -84,7 +84,7 @@ const SummaryTable = ({ reportData }: { reportData: ReportData }) => {
                 a.skippedTests -
                 b.failedTests -
                 b.erroredTests -
-                b.skippedTests
+                b.skippedTests,
             )
             .map((implementation, index) => (
               <ImplementationRow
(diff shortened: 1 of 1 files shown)


Julian:
Hey! Thanks! I've seen this -- but it doesn't seem like the right way to do this to me, even though I don't know what the right one is which is why I can't suggest anything yet. But explicitly hardcoding magic numbers (for the breakpoint) and having to do this in JS both don't feel right. Perhaps look around to see if you can find another way to do this. It's possible I'm wrong and there's no other way, but yeah I'm not comfortable with this as-is which is why I haven't merged it.

Sahil-Shadwal:
Okay will look around . Thanks :)

Julian:
Let me know if you find some other way to do this!

PR #975, opened 2024-02-29
Add Support for Syntax Highlighting using jsonschema_lexer

### LINKED ISSUE #863: Syntax highlight schemas + instances in the terminal
`rich` has support for syntax highlighting, so we should possibly make use of it in order to highlight schemas + instances in the terminal.

Doing this though probably would benefit from writing a Pygments highlighter that understands JSON Schema dialects -- something that likely is a good idea anyhow -- so this is probably the first step here.

sudo-jarvis:
Closes #863.

The goal was to support syntax highlighting for schemas and instances in bowtie's cli.

**Implementation**
Created a custom lexer that understands JSON schema dialect. The lexer has already been published and can be installed using `pip install jsonschema_lexer`. Using this lexer, its pretty easy to add syntax highlighting support using rich's syntax module.

**Comparison of output**

**- Output now looks like:**

Screenshot 2024-02-29 at 7 13 44 PM

**- Output before:**

Screenshot 2024-02-29 at 7 14 18 PM

----
📚 Documentation preview 📚: https://bowtie-json-schema--975.org.readthedocs.build/en/975/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -17,8 +17,8 @@ import sys
 from aiodocker import Docker
 from attrs import asdict
 from diagnostic import DiagnosticError
-from jsonschema_lexer.lexer import ( # type: ignore[reportMissingTypeStubs]
-    JSONSchemaLexer,  
+from jsonschema_lexer.lexer import (  # type: ignore[reportMissingTypeStubs]
+    JSONSchemaLexer,
 )
 from pygments.lexers.data import (  # type: ignore[reportMissingTypeStubs]
     JsonLexer,
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -17,8 +17,11 @@ import sys
 from aiodocker import Docker
 from attrs import asdict
 from diagnostic import DiagnosticError
-from jsonschema_lexer.lexer import (
-    JSONSchemaLexer,  # type: ignore[reportMissingTypeStubs]
+from jsonschema_lexer.lexer import ( # type: ignore[reportMissingTypeStubs]
+    JSONSchemaLexer,  
+)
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -60,6 +60,7 @@ dependencies = [
   "structlog",
   "typing-extensions; python_version<'3.11'",
   "url.py",
+  "jsonschema_lexer",
```
Julian, line 63: This list is alphabetically sorted.

bowtie/_cli.py
```diff
@@ -520,14 +524,24 @@ def _validation_results_table(
 
         for test, test_result in test_results:
             subtable.add_row(
-                Text(json.dumps(test.instance)),
-                *(
-                    Text(test_result[each.id].description)
-                    for each in implementations
+                Syntax(
+                    json.dumps(test.instance),
+                    lexer=JSON_SCHEMA_LEXER,
```
Julian, line 531: This isn't right, is it? The instance isn't a JSON Schema, it should use the normal JSON lexer.

Julian:
This looks like it works -- well done! But I think we may as well wait to merge it until we support all dialects, as otherwise it's just likely to cause confusion I think, given it highlights anything other than 2020 incorrectly. You're going to need to change the interface at least slightly, because you need to be able to pass a specific dialect to the lexer -- i.e., you have to have a way to explicitly select the draft 6 dialect in the lexer in order to highlight the schema `{"additionalItems": false}` correctly.

sudo-jarvis:
@Julian , updated the PR now explicitly specifying the dialect from the report

Julian:
Very nice! Merging, let's see how it looks. Thanks for all your work!

PR #981, opened 2024-03-02
Install the exact version of bowtie when creating report

### LINKED ISSUE #955: Explicitly pass the released Bowtie version to the report workflow when we do a release
There's a small race condition when regenerating reports just after a release -- namely the Install Bowtie step might install the previous version because it gets an old response from PyPI.

We should instead at least directly `==newversion` when we do the install, which will fail perhaps if the version still isn't downloadable from PyPI but at least will do so loudly.

OptimumCode:
This PR adds the following parameters to bowtie action

- exact version to be installed. If not specified the latest available will be installed
- timeout in seconds to wait until the specified version becomes available

Resolves #955 

----
📚 Documentation preview 📚: https://bowtie-json-schema--981.org.readthedocs.build/en/981/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,9 @@ jobs:
     needs: ci
     runs-on: ubuntu-latest
 
+    outputs:
+      version: ${{ steps.version.outputs.tag }}
+
     environment:
       name: PyPI
       url: https://pypi.org/p/bowtie-json-schema
@@ -144,8 +147,14 @@ jobs:
           files: |
             app/bowtie*
           generate_release_notes: true
+      - name: Current version
+        id: version
+        if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+        run: echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
 
   report:
     needs: packaging
     uses: ./.github/workflows/report.yml
+    with:
+      bowtie-version: ${{ needs.packaging.outputs.version }}
     if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -6,6 +6,11 @@ on:
   release:
     types: [published]
   workflow_call:
+    inputs:
+      bowtie-version:
+        type: string
+        required: false
+        default: ''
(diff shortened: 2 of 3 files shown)


OptimumCode:
Hi @Julian, I tried to simplify the retry part as I could but if this still looks complicated I think the best option is to stick with your original idea - fail loudly. The report job can be retried after some time manually in this case

Julian:
Nice work.

> I tried to simplify the retry part as I could but if this still looks complicated I think the best option is to stick with your original idea - fail loudly. 

There's also some retry step actions I know. E.g. [this one](https://github.com/nick-fields/retry) -- I've never used it before but I've definitely seen it in the wild.

Maybe that's worth a shot? Otherwise I'm good with trying either way out to start and we can always decide to change.

OptimumCode:
Didn't know about this action (should have checked that before writing my own implementation 😅), thanks for sharing it with me. I will take a look at it closely later today or tomorrow but from the first look I see only one problem - this action will retry command on any error. My implementation will retry only in case the package version was not found. But I think I can build the command in a way the step fails only if package version is not found. Or if it doesn't matter, we can retry in case of any error - it will definitely simplify the action step

OptimumCode:
With `retry` action it looks a bit simpler. Also, I added parameters in a few places in `report` workflow which I missed before

Julian:
Nice! I was hoping that action would make all the shell go away honestly. But ok let's merge this and see how it goes, we can always make tweaks.

Thanks again for knocking this off! Definitely helps!

OptimumCode:
Hi, @Julian. Thank you for reviewing the PR. I can replace the shell with this:
```
      with:
        timeout_minutes: 10
        max_attempts: ${{ inputs.installation-attempts }}
        retry_wait_seconds: ${{ inputs.installation-wait-seconds }}
        shell: bash
        command: pipx install --python "${{ steps.localpython.outputs.python-path }}" bowtie-json-schema==${{ inputs.version }}
        retry_on: error
```
The only reason I used additional shell is to check that the failure was indeed because the version is not yet available. This was done just to make the failure faster in such cases

Julian:
Yeah, I think probably that (the simpler one) is honestly safe, because the only other kinds of failures should be some other kind of networking error, in which case retrying is *still* probably fine. But as I say I'm good with whatever at this point, we can always change.

PR #983, opened 2024-03-03
Doctesting examples in the documentation

### LINKED ISSUE #12: Doctest the README / doc CLI example
We should ensure the README stays accurate by doctesting it. Likely depends on #5.

adwait-godbole:
Fixes #12 

@Julian one assumption I have made here is to reflect the doctested output in the documentation page so it is in sync with the latest output that `bowtie` commands will produce.

Just one problem I have is with this example of [local checkouts](https://docs.bowtie.report/en/latest/cli/#running-test-suite-tests-from-local-checkouts). Not sure how to doctest it :(. Perhaps just using the `https` test suite url ?

For this PR I have referenced the official Sphinx Documentation for doctesting in `.rst` files:
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -28,6 +28,28 @@ The `bowtie validate <cli:validate>` subcommand can be used to test arbitrary sc
 
 Given some collection of implementations to check -- here perhaps two Javascript implementations -- it takes a single schema and one or more instances to check against it:
 
+.. testcode:: *
+
+    bowtie_validate = 'bowtie validate -i js-ajv -i js-hyperjump <(printf \'{"type": "integer"}\') <(printf 37) <(printf \'"foo"\') | bowtie summary --format pretty'
+    print(asyncio.run(run_command(command=bowtie_validate)))
+
+.. testoutput:: *
+   :options: +NORMALIZE_WHITESPACE
+
+                                         Bowtie
+    ┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+    ┃ Schema              ┃                                                        ┃
+    ┡━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
+    │                     │                                                        │
+    │                     │                                 hyperjump-json-sche…   │
+    │ {                   │   Instance   ajv (javascript)   (javascript)           │
+    │   "type": "integer" │  ────────────────────────────────────────────────────  │
+    │ }                   │   37         valid              valid                  │
+    │                     │   "foo"      invalid            invalid                │
+    │                     │                                                        │
+    └─────────────────────┴────────────────────────────────────────────────────────┘
+                                      2 tests ran
+
 .. code:: sh
 
     $ bowtie validate -i js-ajv -i js-hyperjump <(printf '{"type": "integer"}') <(printf 37) <(printf '"foo"')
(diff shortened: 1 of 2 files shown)


adwait-godbole:
By the way @Julian , forgot to mention I have referenced the official Sphinx Documentation for doctesting in `.rst` files from here:
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

Julian:
This is nice work, but not fully the right "style" here, even if it's close -- the point of doctests should be to test that what we show in the docs is what someone gets when they really run what we suggest (and therefore to detect errors in our docs when they go out of date). But you've changed what we document subtly -- so users will now see Python code rather than shell commands (which of course you're then executing anyhow).

I'm not sure whether you looked at how this renders, but compare [this, which is your doc](https://bowtie-json-schema--983.org.readthedocs.build/en/983/cli/#validating-a-specific-instance-against-one-or-more-implementations) to the [current one](https://docs.bowtie.report/en/stable/cli/#validating-a-specific-instance-against-one-or-more-implementations).

We want the docs to be as clear as humanly possible here, after all, all these examples are "easily discoverable" by reading the `--help`, so users who need them are likely to be very new to Bowtie and can't handle any additional cognitive burden to get up and running -- they need to see "tell me what to run", and may have 0 Python experience whatsoever.

Last I looked, doctesting shell output was not easy in Sphinx, but that's the kind of thing we're after -- where users still see shell commands, but where we verify that the output is what we've manually shown.

Basically -- we want to show essentially exactly what we show right now, but have it be tested, so any solution shouldn't really change the former.

(Going to close this given the above, but let me know what you think nevertheless -- and as usual thanks for taking initiative! As I say, you're definitely close here, but close in this case I think won't get us there with this strategy exactly as-is).

Julian:
But that's the point of a doctest! It's to show some code (`testcode`) and then to show some output (`testoutput`) and then to have a tool assert the output shown is the one someone would really get!

The issue here is just that our testcode needs to be written in shell.

adwait-godbole:
> But that's the point of a doctest! It's to show some code (`testcode`) and then to show some output (`testoutput`) and then to have a tool assert the output shown is the one someone would really get!
> 
> The issue here is just that our testcode needs to be written in shell.

Got it! Thanks!

adwait-godbole:
https://github.com/tillahoffmann/sphinxcontrib-shtest

@Julian I found this repo what are your thoughts on it ?

Julian:
Certainly looks like it has the right kind of functionality -- it's only a year old, which means it may be newer than the last time I looked at this, but also doesn't seem like it has any users from my 3 second look.

In short -- it looks like it should do the right thing, but without users it'd require checking it out to know whether it *does* do the right thing! Not a bad find either way though, that's what we want!

adwait-godbole:
Should I try it out ?

Julian:
That'd be great! I'd recommend at least reading it briefly first before you do (as you would any software but more so software that may not have been run by anyone but its author)! But yes let me know how it goes!

adwait-godbole:
@Julian I tried using this library. The library offers two approaches.

**First Approach:**

It provides the `shtest` directive using which we can perform the doctesting. This is what we wanted. But I am unsure of why this truncation of output happens  (below screenshot). If this happens then our docs will also have to ideally show this truncated output ?. 

Screenshot (65)

Also I am not sure how to `shtest` [running suite across all implementations](https://docs.bowtie.report/en/stable/cli/#running-the-official-suite-across-all-implementations) and [local checkouts](https://docs.bowtie.report/en/stable/cli/#running-test-suite-tests-from-local-checkouts). One more problem with `shtest` .

**Second Approach:**

We can directly use a directive called `sh` and that directive will run our command and display the output on our docs as it is. So no need of **doctesting** really because whenever the docs build happens the command also gets run and the output gets shown. So no manual work of typing out the output ig ? I might be missing out some flaws in here which you could help me point out perhaps :)

Julian:
> It provides the shtest directive using which we can perform the doctesting. This is what we wanted. But I am unsure of why this truncation of output happens (below screenshot). If this happens then our docs will also have to ideally show this truncated output ?. By truncated output I mean "hyperjump-json-sche..."

I think this is rich doing this actually, and probably you can reproduce it by making your terminal narrower than normal. I've noticed it in some cases but not spent time trying to disable it (or even think of what behavior we want honestly). But yeah probably it executes with some narrower tty than yours, and we should think about what we want.

> Also I am not sure how to shtest [running suite across all implementations](https://docs.bowtie.report/en/stable/cli/#running-the-official-suite-across-all-implementations) and [local checkouts](https://docs.bowtie.report/en/stable/cli/#running-test-suite-tests-from-local-checkouts).

I guess by cloning the repo. You can probably do part of that in a setup directive (assumign the library has an equivalent of `testsetup`) and then use an environment variable in the real `shtest`, so you use `PATH_TO_REPO` or whatever, and your setup actually really sets that variable.

> We can directly use a directive called sh and that directive will run our command and display the output on our docs as it is. So no need of doctesting really because whenever the docs build happens the command also gets run and the output gets shown. So no manual work of typing out the output ig ? I might be missing out some flaws in here which you could help me point out perhaps :)

The disadvantage here is if running the command doesn't produce the output we're expecting it to (e.g. because of a bug), then things silently pass.

PR #988, opened 2024-03-04
Fix bowtie action: simplify the retry logic

OptimumCode:
Looks like I didn't test the action well... Sorry for the problems it caused.
Related to #955 

----
📚 Documentation preview 📚: https://bowtie-json-schema--988.org.readthedocs.build/en/988/


Julian:
All good, I was making coffee and then was going to do this myself, so thanks for saving me the time :D

PR #994, opened 2024-03-06
Using `implementation_subcommand` helper on more commands that start implementations

adwait-godbole:
Part of fixing #24.

1)  Now running `bowtie run`, `bowtie suite` or `bowtie validate` without any explicit `-i` will default to `Implementation.known()` 
     or `sys.stdin` in case of `ls implementations/`

     @Julian do we want the above behaviour ?

2) I am not sure whether `reporter.startup_failed` and `reporter.no_such_image` would still be used given that now we rely on `implementation_subcommand` helper handling and printing these two errors which it does in its own way (using `rich` and `panel.Panel`). Though I think we could still use `reporter` for these 2 errors but it seems like here there's a choice between `rich`+`panel.Panel` and `reporter` for these 2 ig? Correct me if I am wrong here or missing something. In this PR, I defaulted to letting `implementation_subcommand` handle these 2.

Also can you please read my last comment on the issue, although I understood some parts of your reply still I am a bit lost on overall what all is expected to completely fix the issue 😅.


adwait-godbole:
> NOTE: Now running bowtie run, bowtie suite or bowtie validate without an explicit -i will default to Implementation.known() or sys.stdin

If this happens then `--format pretty` on `bowtie summary` looks restricted

Screenshot (68)

Julian:
> https://github.com/bowtie-json-schema/bowtie/pull/994

Julian:
> Now running bowtie run, bowtie suite or bowtie validate without any explicit -i will default to Implementation.known() or sys.stdin in case of ls implementations/

Yep that's the desired end state.

> it seems like here there's a choice between rich+panel.Panel and reporter for these 2 ig? Correct me if I am wrong here or missing something. In this PR, I defaulted to letting implementation_subcommand handle these 2.

Yep that's exactly part of the reason I didn't respond yet to your other comment, as that was what made the refactor incomplete, it's not clear yet whether `Reporter` should still exist at all, or if so how it should interact with creating `Report`s, or if so how it should interact with which commands.

And part of the above is also made more challenging by the lack of integration tests with the UI, which means when making changes to the report, it's hard to tell whether those have broken anything in the UI. I added *one* to `parseReportData`, but we need more.

All of which is to say, well done again on having initiative to tackling something here, but I'm not sure, some of it may be in the right direction and some not, depending on the answers to the above, which aren't clear until literally just digging in and seeing what should change and how.

This looks small though as a changeset, so it's more likely to be helpful than not, although I have an initial suspicion it may slightly change behavior of commands (even beyond just the `-i` bit).

adwait-godbole:
Hmm, seems like too much on the plate perhaps! What do you suggest I do here then ?

Julian:
Perhaps starting to help on the integration tests is a good immediate step -- I suspect the UI parsing logic needs to look more like the backend one (but again hard to refactor without tests).

You could perhaps have a look at the test I added and see about adding one or two more at least which test some other cases.

adwait-godbole:
Can you perhaps raise a separate issue and assign me over there please ?

adwait-godbole:
@Julian so now that #999 is done. How do I move forward with this one :) ?

adwait-godbole:
@Julian so now that https://github.com/bowtie-json-schema/bowtie/issues/997 is done. How do I proceed further with this one :) ?

Julian:
> @Julian so now that https://github.com/bowtie-json-schema/bowtie/issues/997 is done. How do I proceed further with this one :) ?

I have I believe another good intermediate task before this PR, which is to do the minimal amount of refactoring to get `bowtie filter-implementations` (i.e. with no arguments) to work *without* starting any containers.

Specifically, running that should immediately be able to print our known implementations, because we should just be printing `Implementation.known()` and exiting.

To do that you'll want to "push down" the responsibility of starting the implementations I think from inside `implementation_subcommand` into the function it wraps (basically pass the context manager down without starting it).

I believe that kind of change should be very small, and also useful for others of these commands.

It also should allow having a test for `bowtie filter-implementations` (again with no arguments), which we may not have right now (I forget and am not checking at the minute).

Do you want to perhaps give that a shot in a new PR, assuming you follow the above?

bowtie/_cli.py
```diff
@@ -880,7 +880,7 @@ def _check_fail_fast_provided(
     default="-",
     type=click.File(mode="rb"),
 )
-async def run(
```
adwait-godbole: tests failed though :(

adwait-godbole:
> Do you want to perhaps give that a shot in a new PR, assuming you follow the above

Yeah got it! Sure!

adwait-godbole:
> To do that you'll want to "push down" the responsibility of starting the implementations I think from inside implementation_subcommand into the function it wraps (basically pass the context manager down without starting it).

I am not sure if I got this 100% but this "push down"  happens only when user enters `bowtie filter-implementations` right ? and would not happen if he passes "-i" ?

Julian:
Push down is a handwavy term, sorry. What is usually meant by it is "change which part of code is responsible for some bit of behavior, moving it from a calling function into the *called* function".

Specifically, right now, `implementation_subcommand` is responsible for actually *starting* the implementations before it calls the wrapped function (e.g. the actual `validate` function, or `filter_implementations`).

What I was suggesting to try is to change that so that now the functions (`filter_implementations`, `validate`) do the starting (by adding a callable argument `start`, and then calling `with start: ...` inside of them when you want to start). That will allow you to *not* enter that in `filter_implementations` if we're simply showing all implementations.

Let me know if that makes sense.

adwait-godbole:
Okay! I will try :)

adwait-godbole:
Just a doubt here. That would result in duplication of code right because then each of `filter-implementations`, `filter-dialects`,`info` and `smoke` which use `@implementation_subcommand` will  now have the same logics of `with start` ? And also that would mean no more of `@implementation_subcommand`. Am i right here or lost somewhere ?

adwait-godbole:
@Julian let me know your thoughts so I can move forward with what you have said / what I have perceived :)

adwait-godbole:
@Julian sorry to ping again here but let me know so I can proceed forward :)

Julian:
d90b6e86e61ee64cdfa2783ef70f822b0ba89b74 is most of the way towards what I was describing.

You might now try and see if there's an easy next step towards continuing on top of it.

adwait-godbole:
> It also should allow having a test for bowtie filter-implementations (again with no arguments), which we may not have right now (I forget and am not checking at the minute).

How will we test this ? Is there a way to mock the output of `ls implementations/` that `bowtie filter-implementations` with no arguments command does internally (since it is test environment I don't think it will know about `known_implementations.json` ?  Or should i just assert if the `stdout` was equal to `Implementation.known()` in the pytest ?

Julian:
Asserting against .known sounds right!

PR #1009, opened 2024-03-11
Add tests for `Badge.ts` and `Site.ts`

### LINKED ISSUE #952: Remove urijs and use `URL`

adwait-godbole:
Fixes #952 

----
📚 Documentation preview 📚: https://bowtie-json-schema--1009.org.readthedocs.build/en/1009/


Julian:
Nice. Well done. Can you think of any tests you can add here as part of this to ensure this works the same as previously?

adwait-godbole:
Sure!

frontend/src/data/Badge.ts
```diff
@@ -1,36 +1,39 @@
...
-const SHIELDS = new URI("https://img.shields.io/endpoint");
+const SHIELDS = new URL("https://img.shields.io/endpoint");
 
-const BADGES = siteURI.clone().segment("badges");
+const BADGES = new URL(siteURI.toString());
+BADGES.pathname += "/badges";
 
-const badgeFor = (uri: URI): URI => SHIELDS.clone().addQuery("url", uri);
+const badgeFor = (url: URL): URL =>
+  new URL(SHIELDS.toString() + "?url=" + encodeURIComponent(url.toString()));
```
Julian, line 11: This also looks wrong. The whole point of using URL types is to completely avoid string manipulation, which is easily going to be wrong.

frontend/src/data/Badge.ts
```diff
@@ -1,36 +1,39 @@
...
-): URI =>
+): URL =>
   badgeFor(
-    implementationBadges(implementation)
-      .clone()
-      .segment("compliance")
-      .segment(dialect.shortName)
-      .suffix("json"),
+    new URL(
+      implementationBadges(implementation).toString() +
```
Julian, line 34: And here too. Yeah a lot of this looks wrong for the above reason.

frontend/src/data/Badge.ts
adwait-godbole: Thanks! I just fixed this on my local branch.

adwait-godbole:
@Julian can you please have a look now.

adwait-godbole:
Yeah adding shortly!

adwait-godbole:
@Julian I added tests for `Badge.ts`

adwait-godbole:
any other tests that I should add here @Julian ?

frontend/src/data/Site.ts
```diff
@@ -1,19 +1,17 @@
...
+siteURI.pathname = new URL(import.meta.env.BASE_URL, base).pathname;
 
 // FIXME: Presumably a future `Implementation` can handle the below, just
 //        as `Dialect` handles it for dialect reports.
-export const implementationMetadataURI = siteURI
-  .clone()
-  .filename("implementations.json")
-  .href();
+export const implementationMetadataURI = new URL(
+  "implementations.json",
```
Julian, line 13: > any other tests that I should add here @Julian ?

I was going to say "you tell me" :), but considering this line is trivially incorrect, there's at least one here!

Julian:
Please also double check whether there is a method or other operator for joining paths. You're again using string manipulation to add segments to existing URLs, when I would hope that's something that should be provided by `URL`.

adwait-godbole:
> Please also double check whether there is a method or other operator for joining paths. You're again using string manipulation to add segments to existing URLs, when I would hope that's something that should be provided by URL.

I can't seem to find anything that does this job in URL API. It's either using "+" or by `template strings` in JS. I think that's what `urijs` was good at :). It handles a lot of such nuances.

frontend/src/data/Site.ts
Julian: Wait is it me that's wrong here, does this actually work right? The relative path comes first? I guess I'm not sure myself, so the test couldn't hurt.

Julian:
OK, fair enough, I may take a closer look -- clearly there's been 2 or 3 things I'd guessed definitely should exist and then turn out not to in JS, so you may well be right and it could be I'm just out of luck again.

No matter what though the tests are going to be good, so this definitely will not be a waste of time I think!

frontend/src/data/Site.ts
adwait-godbole: Yes! The relative path comes first here which makes it really confusing rather having string templates `/` makes it understandable.

Julian:
Let's keep the test at least I think? But yeah thanks for the work regardless!

adwait-godbole:
> Let's keep the test at least I think?

Yeah sure @Julian ! I will have to make some changes in `Badge.test.ts` for it to pass but if possible can you have a please have a look at `Site.test.ts`. I tried hard for it to  pass but for some reason the environment variable never seem to get the correct site url.

Julian:
Great, thanks!

PR #1012, opened 2024-03-12
WIP:- Missing run infos in different languages

adwait-godbole:
Rust, Ruby, C++ are still some of the few pending. Might be some more as well which I am missing. 

But for whoever reading this PR feel free to add more support on this one. Particularly getting this info in C++ is not straightforward is what I came across while Googling.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1012.org.readthedocs.build/en/1012/


Julian:
Nice. This looks mergeable I think even though it's marked WIP? If it fixes these three languages then definitely more PRs for the others are still welcome. And yeah my recollection also was that C++ was the not easy one here...

Going to try it.

PR #1017, opened 2024-03-13
More of missing run infos

adwait-godbole:
Support for Kotlin, Lua, Ruby and Rust. Ig C++ is the only one that remains now :)


adwait-godbole:
@Julian for Rust I used these crates called [os_info] (https://crates.io/crates/os_info) and [rustc_version_runtime](https://docs.rs/rustc_version_runtime/latest/rustc_version_runtime/)  but before that I did went through many iterations of getting the necessary information in Rust without using these 3rd party libs but it never worked for me. Especially getting the OS Version wasn't straightforward until I found out about `os_info` and similarly getting the Rust runtime version also required starting a shell command from the code but after some digging I found that `rustc_version_runtime`  did the job.

adwait-godbole:
@OptimumCode I hope you don't mind me suppressing the `"ConstructorParameterNaming"` here :). It threw the here warning for `os_version` and `language_version` but the frontend expects data in that format (having underscores) so it left me no choice but to suppress the warning. If you have any better ideas for this then happy to hear them out! Thanks :)

implementations/kotlin-kmp-json-schema-validator/src/main/kotlin/Responses.kt
```diff
@@ -17,6 +17,9 @@ class Implementation(
     val homepage: String,
     val issues: String,
     val source: String,
+    val os: String,
+    val os_version: String,
```
OptimumCode, line 21: To keep the naming match kotlin naming convention you can use `kotlinx.serialization.SerialName` annotation.

`@SerialName("os_version") val osVersion: String`

OptimumCode:
Hi @adwait-godbole, I left a comment about how you can change serialization name and keep the property name match Kotlin naming conventions. If you need any help with that, please let me know - I will be happy to assist)

adwait-godbole:
Thanks @OptimumCode it worked like a charm!

OptimumCode (approved):
From my point of view, everything looks good. But I think we should wait for @Julian reply to merge the changes

adwait-godbole:
Yes ofcourse!

adwait-godbole:
Added C++ as well!

adwait-godbole:
The only problem I am facing is that for Clojure and C++ I tried hard to get the dependency versions of their respective json schema libraries in the `version` key. In python implementations it was very simple using `importlib.metadata.version("jsonschema")`. But for Clojure and C++, its not really that straightforward so perhaps we could leave that for some other PR :)

implementations/cpp-valijson/bowtie_valijson.cpp
```diff
@@ -54,6 +81,9 @@ int main() {
     request.Parse(line.c_str());
     string cmd = request["cmd"].GetString();
 
+    map<string, string> os_info = getOSInfo();
```
Julian, line 84: I think you're doing this on every request, rather than just in `start`.

adwait-godbole: Oops! right! sorry! Fixing it :)

implementations/rust-boon/src/main.rs
```diff
@@ -6,6 +6,7 @@ use serde_json::{json, Map, Value};
 fn main() -> Result<(), Box<dyn Error>> {
     let mut started = false;
     let mut draft = None;
+    let osinfo = os_info::get();
```
Julian, line 9: Same here, I think you're doing this on every request as opposed to just where you need it.

adwait-godbole:
@Julian fixed!

implementations/rust-boon/src/main.rs
adwait-godbole: Yeah sorry!

adwait-godbole:
@Julian Thanks! I fixed it for C++, Rust and Lua.

adwait-godbole:
Thanks for cancelling the concurrent builds :)!

PR #1022, opened 2024-03-17
Add tests to check that we emit valid markdown

### LINKED ISSUE #922: Test that we emit valid markdown for `--format markdown` commands
#839 added support for emitting markdown in a number of commands.

We should add test cases that test that the markdown we emit is... valid markdown.

They should do so by attempting to parse our output using a markdown parser, which we should add as a test dependency.

Note that there are different dialects of markdown. Because our target environment for these is GitHub, we should use a GitHub flavored markdown parser.

sudo-jarvis:
Closes #922.

Uses markdown-it-py with gfm and tries parsing the markdown to html. If the resultant token classes match our intended token classes, we get valid markdown.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1022.org.readthedocs.build/en/1022/

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1,5 +1,7 @@
 from contextlib import asynccontextmanager
 from io import BytesIO
+from markdown_it import MarkdownIt
+from markdown_it.tree import SyntaxTreeNode
 from pathlib import Path
 from pprint import pformat
 from textwrap import dedent, indent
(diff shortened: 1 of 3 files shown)


tests/test_integration.py
```diff
@@ -958,6 +960,35 @@ async def test_smoke_markdown(envsonschema):
...
+    stdout, stderr = await bowtie(
+        "smoke",
+        "--format",
+        "markdown",
+        "-i",
+        envsonschema,
+        exit_code=-1,  # because indeed envsonschema gets answers wrong.
+    )
+    parsed_markdown = MarkdownIt("gfm-like", {"linkify": False}).parse(stdout)
+    tokens = SyntaxTreeNode(parsed_markdown).pretty(indent=2)
```
Julian, line 974: Having to test this by going back to a string (i.e. calling `.pretty`) seems a bit non-ideal. Did you check how easy it is to assert against either the thing `.parse` returns directly or else `SyntaxTreeNode`?

sudo-jarvis: Parse actually is returning an array of token objects, with nested tokens part of the parent token itself in the children property. Asserting against this would be complex in itself ig thats why I just used the tokens in text form properly separated by indents according to their nested depth.

Output of parse func:

Screenshot 2024-03-18 at 7 00 50 PM

Julian: Got it, OK, and judging from https://github.com/executablebooks/markdown-it-py/blob/1064835b30325e857252f5b67b34e622261723ca/tests/test_api/test_main.py#L271 that indeed doesn't look like it has any helpers to help us create one. Oh well, that's something missing from the library from what I can tell, but what you did is fine then.

PR #1024, opened 2024-03-17
Stop NavLink's onClick propagation

### LINKED ISSUE #1023: Clicking on an Implementation link in the Summary section fetches all the dialect reports twice
https://github.com/bowtie-json-schema/bowtie/assets/68469605/ef9acea6-8e59-43dc-b822-2c22dcecae95

adwait-godbole:
Fixes #1023 

https://github.com/bowtie-json-schema/bowtie/assets/68469605/908195a4-f1d9-4ada-a664-650ba6918b6e

### DIFF
--- a/frontend/src/components/Summary/ImplementationRow.tsx
+++ b/frontend/src/components/Summary/ImplementationRow.tsx
@@ -2,7 +2,7 @@ import "./ImplementationRow.css";
 import { useState } from "react";
 import { DetailsButtonModal } from "../Modals/DetailsButtonModal";
 import { mapLanguage } from "../../data/mapLanguage";
-import { NavLink, useNavigate } from "react-router-dom";
+import { useNavigate } from "react-router-dom";
 import {
   Case,
   Implementation,
@@ -31,12 +31,9 @@ const ImplementationRow = ({
         onClick={() => navigate(`/implementations/${implementationPath}`)}
         scope="row"
       >
-        <NavLink
-          to={`/implementations/${implementationPath}`}
-          onClick={(e) => e.stopPropagation()}
-        >
+        <span className="text-decoration-underline text-primary">
           {implementation.name}
-        </NavLink>
+        </span>
         <small className="text-muted ps-1">
           {mapLanguage(implementation.language)}
         </small>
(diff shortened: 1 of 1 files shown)


Julian:
Thanks! Can you explain a bit what's happening here? I mean I understand what click propagation means, but what event on what thing is actually causing the double download?

adwait-godbole:
Yes sure @Julian! Basically if you see in the PR itself the <th> element above the NavLink also has an onClick navigate and the NavLink is also navigating to the same route so when you click on the NavLink the click happens on the <th> because of propagation so you get routed to the same route twice leading to same data fetching twice

adwait-godbole:
Did that make sense ?

Julian:
Kind of, but is there a reason we have a `NavLink` there (inside the row) in the first place then? Presumably that was left over from before we made the whole row be clickable. Is the right fix not just to remove the NavLink and let the `th` handle the click?

adwait-godbole:
We could do that as well. Sure 👍

adwait-godbole:
I will remove the NavLink

adwait-godbole:
I thought that NavLink was left there by intention to make both row and the link clickable.

Julian:
That's definitely the desired behavior, I guess I'm just asking if having the `NavLink` there is required to get it.

adwait-godbole:
Well in that case ig the NavLink is just present there to make it appear as a clickable link with blue colored text and underlining. its definitely not required there since the row is already clickable but if I remove the NavLink from there then ig the link blue color highlighting won't happen then

adwait-godbole:
I could use `e.preventDefault()` on the `NavLink` as well to disable its event entirely

adwait-godbole:
> Is there a class in bootstrap which makes regular text look like a link?

didn't find one for text but there's `btn-link` but it works on `button`. Ig writing custom css is the way ?

adwait-godbole:
Yeah I will make the changes 👍

Julian:
If there isn't one I think the `preventDefault` way sounds best in that case, if all that does is make it not do anything.

adwait-godbole:
@Julian sorry for the above comment but I instead made it a `<span>` and made applied some Bootstrap classes on it. It looks like a link now and doesn't download data twice!

Julian:
Great, looks good, thanks!

adwait-godbole:
I should've checked in dark mode as well! Should I raise another PR for this ?

Julian:
That sounds either like a bootstrap bug or that we don't quite have the right classes, or that we need a conditional. Check if uograding helps if we're behind, or else maybe double check the dark mode docs to see whether they say we're supposed to use a different color class?

PR #1030, opened 2024-03-18
Fix dark mode implementation link

adwait-godbole:
The class is called `text-primary-emphasis` in dark mode and `text-primary` in light mode. So it works correctly now :)

----
📚 Documentation preview 📚: https://bowtie-json-schema--1030.org.readthedocs.build/en/1030/


Julian:
Perfect, thanks!

PR #1042, opened 2024-03-24
Remove `results` from `Implementation` interface and simplifying  `ImplementationData`

adwait-godbole:
```typescript
// FIXME: Move this outside of here to some external map!?
results: Record<string, Partial<Totals>>;
```

Hi @Julian in these changes I've tried removing the above from `Implementation` interface and introduced a new interface called `ImplementationDialectCompliance` that holds the implementation metadata along with the multiple dialect compliance numbers as the follows:

```typescript
export interface ImplementationDialectCompliance {
  implementation: Implementation;
  dialectCompliance: Record<string, Partial<Totals>>;
}
```

The justification for doing the above is that earlier the  **`results`**  key in `Implementation` interface was only being used for [showing the `DialectCompliance` numbers across all dialects on the `ImplementationReportView` page](https://github.com/bowtie-json-schema/bowtie/blob/059ccd80865d4fbd4fa6eab4b8d4520bbd7f82b7/frontend/src/components/ImplementationReportView/DialectCompliance.tsx#L38) which was being set over [here](https://github.com/bowtie-json-schema/bowtie/blob/059ccd80865d4fbd4fa6eab4b8d4520bbd7f82b7/frontend/src/data/parseReportData.ts#L196) in `parseImplementationData` function. So I took it out of the `Implementation` interface just like what you had pointed out in the FIXME and introduced a new interface that holds this data (implementation data + dialect compliance numbers for it) together.

After this I made a try at simplifying `ImplementationData` interface because after doing the above changes the only difference remaining between `Implementation` and `ImplementationData` was `dialects: string[]` in `ImplementationData` as opposed to `dialects: Dialect[]` in `Implementation`. So I've tried a few changes for this like doing the below

```typescript
interface ImplementationData extends Omit<Implementation, "dialects"> {
      dialects: string[];
}
```

 Let me know your thoughts about the same. 

Thanks!

### DIFF
--- a/frontend/src/DialectReportView.tsx
+++ b/frontend/src/DialectReportView.tsx
@@ -19,6 +19,7 @@ export const DialectReportView = ({
   topPageInfoSection?: React.ReactElement;
   allImplementationsData: Record<string, Implementation>;
 }) => {
+  console.log(reportData);
   const params = useSearchParams();
   const languages = useMemo(() => {
     const langs = new Set<string>();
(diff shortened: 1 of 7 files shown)


Julian:
Very nice, this looks perfect, thanks!

PR #1048, opened 2024-03-25
In light of PR #1042 having tests for `prepareImplementationReport` (across all dialect compliances) won't hurt :)

adwait-godbole:
This test specifically tests [this function](https://github.com/bowtie-json-schema/bowtie/blob/0e642d67dc7c0344bab52ea4e34bd25419815f1c/frontend/src/data/parseReportData.ts#L177) which is called [here](https://github.com/bowtie-json-schema/bowtie/blob/0e642d67dc7c0344bab52ea4e34bd25419815f1c/frontend/src/index.tsx#L38) when we are on the `ImplementationReportView` page to collect dialect compliance data for all the dialects.

### DIFF
--- a/frontend/src/data/parseReportData.test.ts
+++ b/frontend/src/data/parseReportData.test.ts
@@ -131,7 +131,7 @@ describe("parseReportData", () => {
     const report = fromSerialized(lines);
 
     const metadata = report.runMetadata.implementations.get(
-      tag("envsonschema")
+      tag("envsonschema"),
     )!;
     const testCase = report.cases.get(1);
 
@@ -154,7 +154,7 @@ describe("parseReportData", () => {
         ]),
         report.runMetadata.bowtieVersion,
         report.runMetadata.started,
-        {}
+        {},
       ),
       implementationsResults: new Map([
         [
@@ -195,13 +195,13 @@ describe("parseReportData", () => {
 
     const lines = bowtie(
       ["run", "-i", tag("envsonschema"), "-D", "7"],
-      cases.join("\n") + "\n"
+      cases.join("\n") + "\n",
     );
 
     const report = fromSerialized(lines);
 
     const metadata = report.runMetadata.implementations.get(
-      tag("envsonschema")
+      tag("envsonschema"),
     )!;
 
     expect(report).toStrictEqual({
@@ -223,7 +223,7 @@ describe("parseReportData", () => {
         ]),
         report.runMetadata.bowtieVersion,
         report.runMetadata.started,
-        {}
+        {},
       ),
       implementationsResults: new Map([
         [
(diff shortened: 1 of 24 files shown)


Julian:
Thanks! I actually find/found that function quite obtuse when I noticed it wasn't doing the right thing. I honestly still don't understand what it's meant for, though if I stared at it I suspect I could find out. But I don't know what `parseImplementationData` means, it certainly doesn't seem to be what the function does, it has references to `allImplementations` though that isn't really all implementations it's some other calculation.

All of which is to say once you're here and helping perhaps you can either clarify the function name, write a docstring, or make it all-in-all more logical?

adwait-godbole:
That function indeed has a wrong logical name to start with. I didn't mind to change the name in the prior PR but what this function basically does is that it takes the dialect data that comes from all the downloaded dialect reports when we visit Implementation Report page and combines it together and assigns dialect compliance data for all the implementations (the reason why there existed a `results` key earlier on `Implementation`) from which we later only select the dialect compliance data of the implementation currently on the report view page.

My PR did the below changes:

**BEFORE**:

Screenshot (104)

**AFTER**:

Screenshot (106)

So I have some ideas here. First of all we should change the function name to something else maybe `parseImplementationDialectComplianceReports` and instead of it requiring to manipulate all Implementations data and returning it and us then filtering from that in the Implementation Report View component [here](https://github.com/bowtie-json-schema/bowtie/blob/0e642d67dc7c0344bab52ea4e34bd25419815f1c/frontend/src/components/ImplementationReportView/ImplementationReportView.tsx#L27), I think a better way would be to call the function with the language implementation that was selected and only make it filter and return the dialect compliance data for that selected implementation. 

I will make the changes and open a PR.

adwait-godbole:
PS. @Julian I updated the screenshots (zoomed a bit)

adwait-godbole:
@Julian I have a test tomorrow, so gotta go and study :). Will open the PR to fix the function tomorrow (it's still working correctly but in an unconventional way for now and yes not well documented at all) for sure. Thanks!

Julian:
No rush.

adwait-godbole:
Hi @Julian I have a question. Is there any scenario where the below happens ?

E.g. implementation metadata information for let's say `go-jsonschema`  in the `runMetadata` of `draft2020-12` report was different than its metadata information in the `runMetadata` of `draft7` ?

adwait-godbole:
> (At some point we may run reports for all dialects at once and then there may only even be one).

That would be awesome! whenever it happens :)

frontend/src/data/parseReportData.test.ts
```diff
@@ -258,29 +266,93 @@ describe("parseReportData", () => {
...
+          name: "envsonschema",
+          language: "python",
+          homepage: metadata.homepage,
+          issues: metadata.issues,
+          source: metadata.source,
+          dialects: metadata.dialects,
+          links: metadata.links,
+        },
+        dialectCompliance: {
+          [Dialect.withName("draft2020-12").shortName]: {
```
Julian, line 324: Dialect.withName("draft2020-12").shortName is just `draft2020-12` right? (And all the rest below)

adwait-godbole: @Julian I’ll have to fix this test code once you merge PR #1058

PR #1049, opened 2024-03-25
Remove colorama from test requirements

adwait-godbole:
For some reason when I ran `pip-compile --strip-extras test-requirements.in` for adding `pexpect` it added `colorama` as a test dependency [here](https://github.com/bowtie-json-schema/bowtie/pull/1020/files#diff-fac4c6890301d4de5c3f4266837803d5240c84a3d8b6c735bbc6a64c39d2f94e) (via `click` and `pytest`) which wasn't already present in our `test-requirements.txt`, so no need!. 

Also removed the `.split` logic and used `.removeprefix` just like how we do in the other case.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1049.org.readthedocs.build/en/1049/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -994,7 +994,7 @@ async def filter_implementations(
     """
     if not dialects and languages == KNOWN_LANGUAGES:
         for implementation in ctx.params["image_names"]:
-            click.echo(implementation.split("/")[-1])
+            click.echo(implementation.removeprefix(f"{IMAGE_REPOSITORY}/"))
         return
 
     async for each in start():
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -42,10 +42,6 @@ click==8.1.7
     # via
     #   -r requirements.txt
     #   bowtie-json-schema
-colorama==0.4.6
-    # via
-    #   click
-    #   pytest
 cryptography==42.0.5
     # via
     #   -r requirements.txt


Julian:
Cool, thanks, though wait, the only way to update the requirements files should basically be by running `nox -s requirements`, is that what you ran now?

adwait-godbole:
Ohh I thought we add the dependency in `test-requirements.in` and then run `pip-compile --strip-extras test-requirements.in`. I did this way. Sorry :(

adwait-godbole:
Reference:- https://github.com/bowtie-json-schema/bowtie/pull/1020/files#diff-fac4c6890301d4de5c3f4266837803d5240c84a3d8b6c735bbc6a64c39d2f94e

Julian:
All good

> pip-compile --strip-extras test-requirements.in to generate test-requirements.txt

adwait-godbole:
> This will do the same thing essentially if you run it with the same Python version, and if you remember to also update the other requirements files. The nox command exists basically to make sure you don't make a mistake there essentially, so yeah if that's not what you ran you can run it now and it'll fix up the files.

Thanks! I will remember this for the future but I confirmed [here](https://github.com/bowtie-json-schema/bowtie/pull/1020/files#diff-fac4c6890301d4de5c3f4266837803d5240c84a3d8b6c735bbc6a64c39d2f94e), `colorama` is the only unnecessary one that was added for now. Nothing else.

adwait-godbole:
So the fix is done 👍

Julian:
Great, merged, thanks.

PR #1057, opened 2024-03-26
Choose default dialect form $schema in bowtie validate command

### LINKED ISSUE #995: Make `bowtie validate` default to a detected `$schema` rather than the latest one
After #180 is done, we should make:

```
bowtie validate -i python-jsonschema <(printf '{"$schema": "https://json-schema.org/draft/2019-09/schema"}') <(printf '12')
```

default to Draft 2019-09 rather than 2020-12, by virtue of the present `$schema` property.

In other words, the behavior should be:

* if `-D` is not passed and `$schema` is found in the schema, use that as our dialect
* If `-D` is not passed and `$schema` is not found, default to the latest dialect
* if `-D` is passed, it must match as per #180, or else a warning is raised

sudo-jarvis:
Closes #995.

Configured `bowtie validate` to pick dialect from schema if not specified instead of the latest one. If dialect not specified in schema as well then it would default to the latest one.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1057.org.readthedocs.build/en/1057/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -722,7 +722,9 @@ def _set_dialect(ctx: click.Context, _, value: _Dialect):
     if value:
         return value
     schema = ctx.params.get("schema")
-    dialect_from_schema = schema.get("$schema") if schema else None
+    dialect_from_schema = (
+        schema.get("$schema") if isinstance(schema, dict) else None
+    )
     return (
         Dialect.from_str(dialect_from_schema)
         if dialect_from_schema
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -921,7 +921,7 @@ def run(
 @click.argument(
     "schema",
     type=click.File(mode="rb"),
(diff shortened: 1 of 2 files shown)


bowtie/_cli.py
```diff
@@ -715,6 +715,21 @@ def convert(
...
 
 
+def _set_dialect(ctx: click.Context, _, value: _Dialect):
+    """
+    Sets the dialect according to $schema if specified.
+    """
+    if value:
+        return value
+    schema = ctx.params.get("schema")
+    dialect_from_schema = schema.get("$schema") if schema else None
```
Julian, line 725: This looks like it will blow up if `schema` is the `True` schema.

sudo-jarvis: Oh yeah you are right!

sudo-jarvis:
@Julian, just to update you currently I am having my exams because of which I am not able to contribute much. I will resume after 3rd april.

Julian:
Merged after a minor tweak to the tests, thanks again!

PR #1058, opened 2024-03-26
Fix `parseImplementationData` function and minor interface renaming

adwait-godbole:
@Julian let me know if the function makes sense now :)


adwait-godbole:
P.S. I've added docstrings as well

frontend/src/data/parseReportData.ts
```diff
@@ -174,54 +174,73 @@ export const implementationFromData = (
     dialects: data.dialects.map((uri) => Dialect.forURI(uri)),
   }) as Implementation;
 
-export const parseImplementationData = (
-  loaderData: Record<string, ReportData>,
+/**
+ * Computes a summarized implementation report for the
```
Julian, line 178:
```suggestion
 * Compute a summarized implementation report for the
```

frontend/src/data/parseReportData.ts
```diff
@@ -174,54 +174,73 @@ export const implementationFromData = (
...
+ */
+export const getImplementationReport = (
+  allReportsData: Record<string, ReportData>,
+  langImplementation: string,
 ) => {
-  const allImplementations: Record<string, ImplementationDialectCompliance> =
-    {};
-  const dialectCompliance: Record<string, Record<string, Partial<Totals>>> = {};
+  let implementationReport: ImplementationReport | null = null;
+  const dialectCompliance: Record<string, Partial<Totals>> = {};
```
Julian, line 187: How big of a change is it if we make the properties here be `Dialect`s rather than string?

Julian (commented):
Thanks! Looks reasonable from first pass, well done, will take a closer look tomorrow.

frontend/src/data/parseReportData.ts
adwait-godbole: Should I change it to "Prepare a summarized implementation report...." ?

frontend/src/data/parseReportData.ts
adwait-godbole: Ig its not possible @Julian it shows the below warning

Screenshot (110)

adwait-godbole: Ig its not possible with `Record` @Julian it shows the below warning

Screenshot (110)

But it does not throw the warning if we use `Map`. So I'll look into it.

adwait-godbole: @Julian, it worked!, I changed it from `string` to `Dialect` :). Now it looks like this

Screenshot (111)

frontend/src/data/parseReportData.ts
adwait-godbole: I changed it to "Prepare" since we aren't really computing anything ig. So now the docstring reads as follows:

/**
 * Prepare a summarized implementation report for the
 * passed implementation using all the dialect reports data
 * that was fetched.
 */

Julian: The verb matters less to me (either's fine) compared to that the docstring is in imperative mood ("Do foo") rather than indicative ("Does foo") but yeah Prepare is good with me.

frontend/src/data/parseReportData.ts
Julian: Great

frontend/src/data/parseReportData.ts
```diff
@@ -174,54 +174,73 @@ export const implementationFromData = (
...
-      erroredTests: value.erroredTests,
-      skippedTests: value.skippedTests,
-      failedTests: value.failedTests,
-    };
+const updateDialectCompliance = (
+  dialectCompliance: Map<Dialect, Partial<Totals>>,
+  dialect: Dialect,
+  implementationResults: ImplementationResults,
+) =>
+  dialectCompliance.set(dialect, {
```
Julian, line 238: This is just a one line function that's only used in one place isn't it? Any reason not to just inline this above?

adwait-godbole: Cool I will do it inline. Just a sec

adwait-godbole: @Julian doing inline now 👍

adwait-godbole:
@Julian resolved the conflicts as well 👍. looking good or any other suggestions ?

PR #1063, opened 2024-03-27
Added test for `prepareImplementationReport`

adwait-godbole:
@Julian I am slightly lost on the naming convention here 😅. Let me know your thoughts.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1063.org.readthedocs.build/en/1063/


adwait-godbole:
Hi @Julian can you spare some time to review this PR please ? 

Also I had a query unrelated to this PR but I was writing the code for introducing a separate  `Implementation` class just like `Dialect` class and came across this [piece of code](https://github.com/bowtie-json-schema/bowtie/blob/d55e9b2ebc3cbe207a52434271436ca6c08e1258/frontend/src/data/Dialect.ts#L88) in `Dialect.ts`, here we are creating the instances of `Dialect` class at runtime when our frontend loads on the browser and they are stored in-memory for all the time period until a user leaves the website. 

So do we want to do something similar for `Implementation` class as well ? i.e. to have all `Implementation` instances created at runtime when the frontend loads ?

Julian:
> I am slightly lost on the naming convention here

Which naming convention bit is this referring to?

> So do we want to do the same for Implementation class as well ?

Not unless we have some runtime need for them, for Dialect we have one (routing).

Julian:
PR looks good I think, thanks!

adwait-godbole:
> Which naming convention bit is this referring to?

Not much basically I only changed one parameter name of `prepareImplementationReport` so I was asking about that.

> Not unless we have some runtime need for them, for Dialect we have one (routing). And even there it'd be nice to fix that, I just did the minimal improvement over how it worked previously essentially.

What fix do you mean here ?

Julian:
Removing the side effects there (creating a thing should never be used to mutate some random other thing, there `Dialect.all`, but I didn't have time to go change the place that relies on that). After that it's possible we indeed don't need to store dialects in memory, I think it's only needed for setting up routing but we'd need to check.

adwait-godbole:
@Julian apart from routing we also have `Dialect.known` right ? which returns all the known dialects stored in memory. So now do you also mean we create all the known dialects on fly ?

adwait-godbole:
Hi @Julian awaiting your reply on the above :)

Julian:
Possibly? I haven't tried to make the fix myself, the main problem at the minute is just the side effects, using `new` to build up a list of known dialects rather than just building up a list.

If we can *also* not have them always in memory and remove .known great, but if not that's also fine.

PR #1065, opened 2024-03-27
Dynamic Theming for Schema Display color 

nomandhoni-cs:
# In Light theme
### New
image

### Old
image

# In Dark theme

### New
image

### Old
image


Julian:
Thanks

Issue #1043, opened 2024-03-24
Opening Embed Badges dropdown on Implementation Report View page is too much on the left side


adwait-godbole:
https://github.com/bowtie-json-schema/bowtie/assets/68469605/4f29c216-d70f-4b64-bb1f-619530eaca3f

Perhaps we could center it a little bit or better we can try using [React Bootstrap's Modals](https://react-bootstrap.netlify.app/docs/components/modal/#live-demo) ig ? (just a suggestion, since it would look more user friendly to open a modal instead given that its a pretty huge component to show on a dropdown click)

Julian:
I'm certainly open to improvements here, as what's there was essentially just to get this at minimally good enough to go out, but can you elaborate a bit? Specifically what's not user friendly? Usually I'd understand user friendliness to mean "a user who wants to use this functionality can't figure out how to do so". But you seem to mean something else, having to do with how large the component is. What's wrong with a large component?

Julian:
Oh I see your video and the title rename now -- your viewport is at lower resolution than mine I think, I hadn't looked at it at that resolution -- I guess on lower resolutions we probably should pad less, but yeah whatever improvements are fine there certainly.

adwait-godbole:
Yeah @Julian my local machine is at a resolution of 1920x1080

adwait-godbole:
Fixed in #1066

PR #1066, opened 2024-03-27
WIP: Changing embed badge dropdown to a modal.

sd1p:
@Julian I was experimenting around making the embed badge UI responsive using modal instead of a dropdown, resolving issue https://github.com/bowtie-json-schema/bowtie/issues/1043. The new layout looks great when the viewport width is greater than 425px.

I was thinking, for mobile displays, can we have a different layout?

https://github.com/bowtie-json-schema/bowtie/assets/43758514/3c0655a8-735d-4bfd-b5bf-e046a5056fa0

### DIFF
--- a/frontend/src/components/ImplementationReportView/EmbedBadges.tsx
+++ b/frontend/src/components/ImplementationReportView/EmbedBadges.tsx
@@ -1,6 +1,5 @@
 import React, { useState } from "react";
 import Container from "react-bootstrap/Container";
-import DropdownButton from "react-bootstrap/DropdownButton";
 import Image from "react-bootstrap/Image";
 import ListGroup from "react-bootstrap/ListGroup";
 import Stack from "react-bootstrap/Stack";
@@ -8,6 +7,7 @@ import Stack from "react-bootstrap/Stack";
 import CopyToClipboard from "../CopyToClipboard";
 import { Implementation } from "../../data/parseReportData";
 import { Badge, badgesFor } from "../../data/Badge";
+import { Button, Modal } from "react-bootstrap";
 
 const EmbedBadges: React.FC<{
   implementation: Implementation;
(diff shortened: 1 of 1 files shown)


sd1p:
Created a new layout for mobile ( viewport width <=425px) for making the modal responsive.
  

https://github.com/bowtie-json-schema/bowtie/assets/43758514/0d89d5b6-a492-43f8-bf9a-63f68992494f

Julian:
That seems odd looking to me on mobile. At that width probably it should just be a full-screen modal I think.

sd1p:
Added overflow-x-scroll on card content to maintain viewport width within full screen size

https://github.com/bowtie-json-schema/bowtie/assets/43758514/d677736f-17ed-41b0-8841-869c220d1ee7

frontend/src/components/ImplementationReportView/EmbedBadges.tsx
```diff
@@ -16,90 +17,180 @@ const EmbedBadges: React.FC<{
...
+                          {badge.name}
+                        </ListGroup.Item>
+                      ))}
+                    </ListGroup>
+                  </ListGroup.Item>
+                ))}
+              </ListGroup>
+
+              <div className="overflow-hidden">
+                <div className="pb-3">
```
Julian, line 64: I haven't stared at this for long but there's obvious duplication here -- I assume for desktop and mobile? Is there not a way you can think of to remove it?

sd1p: Here's a quick prototype for layout without duplication which uses a flex box.

https://github.com/bowtie-json-schema/bowtie/assets/43758514/a06f200a-05ce-4922-9901-482281560ec0

sd1p: If the above layout looks good then I will push it.

Julian: I don't like the two selectors being next to each other, I put them on opposite sides intentionally, though I definitely could believe further improvements are possible, but putting them on the same side I think is a step backwards.

One selector is something someone will definitely touch (the "which badge do I need one"), and likely touch >1 time (because probably they'll want more than one badge).

One selector *most users* will not even need to touch at all, because we default to Markdown, and 95% of all users will want a Markdown badge.

Putting the two next to each other makes it way easier to click a different thing than you wanted to, and also is just messy feeling to me for a reason I can't quite put a finger on. I wanted to even use a different kind of widget for the two given the above, but couldn't think of a different one.

None of the above has anything to do with mobile though, it's just something I see you changed, not sure why.

On mobile, I think the most logical "small" change is to have things be vertical rather than horizontal.

sd1p: Thanks for explaining me the UX perspective of the UI. I have made changes accordingly. I also tried to implement syntax highlighting for consistency but It is not working `badgeEmbed` template string.


https://github.com/bowtie-json-schema/bowtie/assets/43758514/52abd8ef-cb7e-4099-a340-ebf0287b5417

Julian: Can you put the padding back on the code block? Having one single line is really hard to interact with, you can't scroll unless you put the cursor exactly on the line, instead of having a big region -- it's not like there's anything else near there, so it seemed very likely that the box should have lots of vertical padding even if the code is just one line.

sd1p: Thanks for the feedback, I have added vertical padding for the code block.

sd1p: Current version of the modal.

image

Julian:
This still looks different in not-strictly-better-seeming ways than the existing one. For comparison, this is what I see for the current component:

Screenshot 2024-04-09 at 9 53 18 AM

And on this branch I see:

Screenshot 2024-04-09 at 9 52 54 AM

I'm on vacation so I'll be slow to respond, but some observations are:

* you've changed the placement of the horizontal/vertical rules, and the other one seems to fit better to me
* the padding + margin is still too small on the code block
* overall yours feels a bit cramped

As I said there definitely are things I don't like about the existing one -- e.g. the width of the left column is really wrong. But the above doesn't seem to change that, it changes other things which I think were better before seemingly.

Can you try to make the improvement we want (changing to a modal) but less changing of other things unless you call them out and they seem notably better? (Appreciated on all the work of course!)

sd1p:
Apologies for the delay in addressing the PR, I was occupied with practical exams at my college.

sd1p:
I have made the modal responsive by using a grid layout, without making the CSS very complex or implementing any major changes.


https://github.com/bowtie-json-schema/bowtie/assets/43758514/26a9f849-62c6-482d-8240-a28a578a133d


image


✅ Improved padding + margin for the code block.
✅ Improved padding and width of the columns for better look and feel.

The only issue I have in the above commit is that for toggling the background color of the code block (for light/dark mode), I am using inline CSS. I have to move custom background colors to some external CSS file.

sd1p:
Apologies for the force pushes, my intention was to squash several commits, but it didn't go as planned.

Julian:
All fine I don't mind either way. Is this ready for review then?

sd1p:
> All fine I don't mind either way. Is this ready for review then?

Yeah.

Julian:
It still is narrower than before, and therefore looks awkward to me. It looks like to make the modal wider at this point there is no `xxl` in `react-bootstrap` and you'd need to use `dialogClassName`.

Can you also explain why you're explicitly defining colors (in CSS)? What color is missing from where without that, doing so seems very suspicious as we're using a theme for the code block, so those colors would be wrong with any theme change?

frontend/src/components/ImplementationReportView/EmbedBadges.tsx
```diff
@@ -1,105 +1,148 @@
...
   const allBadges = badgesFor(implementation);
+  const { isDarkMode } = useContext(ThemeContext);
 
   const [activeFormat, setActiveFormat] = useState(supportedFormats[1]);
   const [activeBadge, setActiveBadge] = useState(allBadges.Metadata[0]);
-
+  const [show, setShow] = useState(false);
   const badgeEmbed = activeFormat.generateEmbed(activeBadge);
 
+  const compareBadges = (a: Badge, b: Badge) => {
```
Julian, line 31: If you're going to define this (which I previously considered but decided it wasn't really any better than just relying on the name alone for now) I think it belongs in `data/Badge`, especially if/when we have a badge class rather than just an interface. But feel free also to leave it as it was before, I don't mind either way personally.

sd1p:
> Can you also explain why you're explicitly defining colors (in CSS)? What color is missing from where without that, doing so seems very suspicious as we're using a theme for the code block, so those colors would be wrong with any theme change?

I attempted to increase the size of the code-block, by adding padding to the code block and also added padding to it parent container, which required using explicit background color. I will try to implement it in a better way without explicitly declaring colors in CSS. I will also increase the width of the modal so it doesn't look squarish. 

Thanks for the feedback.

frontend/src/components/ImplementationReportView/EmbedBadges.tsx
sd1p: I have removed the comparator function and reverted the changes. Thanks.

sd1p:
I have made the modal more wider and removed hardcoded CSS colors.

image

sd1p:
In the previous commit the code block content overflow is not looking good. I tried to use `wrapLongLines` in syntax-highlighter to center the code content and reduced vertical padding as users don't need to scroll to copy code. 

I have checked the wrap code lines for all 5 available formats, everything looks good to me.

image

PR #1073, opened 2024-04-01
Tab completion for Bowtie commands and implementations

### LINKED ISSUE #802: Add tab completion for implementations (and otherwise)
Depends of course on something like #24 so we know what these implementations are.

adwait-godbole:
Fixes #802 

Hi @Julian 

The shell completion for implementations is working fine 👍 along with the completions for other bowtie commands and command options as well. Here's a demo :)

https://github.com/bowtie-json-schema/bowtie/assets/68469605/fb4d3fb7-e92e-41b9-aa37-3cfb5d14d181

I have some queries though:

1) We could add shell completions for languages as well ? and/or any other completions you've in mind ?

2) Regarding the automatically enabled shell completion for Homebrew installation that you asked for, pardon me but I am very much new to Homebrew formulae and Ruby 😅 (my local machine is Windows :( )  so I tried thinking of putting the below Ruby script in the [Homebrew install formula](https://github.com/bowtie-json-schema/homebrew-tap/blob/d106aa46ce80f643db9ccbd3f3cf9352205cc07b/bowtie.rb#L200) we currently have but this way didn't really work :(. The script doesn't enable the shell completion even if I just save it in a file and do `ruby test.rb`. Can you help me out here please ?

```ruby
require 'etc'

case ENV['SHELL']
when /bash/
  system("bash -c 'eval \"$(\_BOWTIE\_COMPLETE=bash\_source bowtie)\"'")
when /zsh/
  system("zsh -c 'eval \"$(\_BOWTIE\_COMPLETE=zsh\_source bowtie)\"'")
when /fish/
  system('_BOWTIE_COMPLETE=fish\_source bowtie | source')
else
  puts "Unsupported shell: #{shell}"
end
```

Thanks!

----
📚 Documentation preview 📚: https://bowtie-json-schema--1073.org.readthedocs.build/en/1073/

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -25,7 +25,7 @@ The Bowtie CLI supports tab completion using the `click module's built-in suppor
 Below are short instructions for your shell.
 
 .. tabs::
-    .. tab:: Bash
+    .. group-tab:: Bash
 
         Add this to ``~/.bashrc``:
 
@@ -33,7 +33,29 @@ Below are short instructions for your shell.
 
             $ eval "$(_BOWTIE_COMPLETE=bash_source bowtie)"
 
-        **OR**
+    .. group-tab:: Zsh
+
+        Add this to ``~/.zshrc``:
+
+        .. code:: sh
+
+            $ eval "$(_BOWTIE_COMPLETE=zsh_source bowtie)"
+
+    .. group-tab:: Fish
+
+        Add this to ``~/.config/fish/completions/bowtie.fish``:
+
+        .. code:: sh
+
+            $ _BOWTIE_COMPLETE=fish_source bowtie | source
+        
+        This is the same file used for the activation script method below. For Fish it's probably always easier to use that method.
+
+Using ``eval`` means that the command is invoked and evaluated every time a shell is started, which can delay shell responsiveness. 
+To speed it up, write the generated script to a file, then source that.
+
+.. tabs::
+    .. group-tab:: Bash
 
         Save the script somewhere.
 
@@ -47,15 +69,7 @@ Below are short instructions for your shell.
 
             $ . ~/.bowtie-complete.bash
 
-    .. tab:: Zsh
-
-        Add this to ``~/.zshrc``:
-
-        .. code:: sh
-
-            $ eval "$(_BOWTIE_COMPLETE=zsh_source bowtie)"
-
-        **OR**
+    .. group-tab:: Zsh
 
         Save the script somewhere.
 
@@ -69,15 +83,7 @@ Below are short instructions for your shell.
 
             $ . ~/.bowtie-complete.zsh
 
-    .. tab:: Fish
-
-        Add this to ``~/.config/fish/completions/bowtie.fish``:
-
-        .. code:: sh
-
-            $ _BOWTIE_COMPLETE=fish_source bowtie | source
-
-        **OR**
+    .. group-tab:: Fish
 
         Save the script to ~/.config/fish/completions/bowtie.fish:
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -47,10 +47,10 @@ Add tab completion for implementations (and otherwise)
         .. code:: sh
 
             $ _BOWTIE_COMPLETE=fish_source bowtie | source
-        
+
         This is the same file used for the activation script method below. For Fish it's probably always easier to use that method.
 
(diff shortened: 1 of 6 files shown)


docs/cli.rst
```diff
@@ -18,6 +18,38 @@ Below are a few sample command lines you might be interested in.
     If you have cloned Bowtie to :file:`/path/to/bowtie` you should be able to use ``$(ls /path/to/bowtie/implementations/ | sed 's/^| /-i /')`` in any command to expand out to all implementations.
     See `below <cli:running the official suite across all implementations>` for a full example.
 
+Add tab completion for implementations (and otherwise)
```
Julian, line 21:
```suggestion
Enabling Shell Tab Completion
```

docs/cli.rst
```diff
@@ -18,6 +18,38 @@ Below are a few sample command lines you might be interested in.
...
+
+    .. tab:: Fish
+
+        Add this to ``~/.config/fish/completions/bowtie.fish``:
+
+        .. code:: sh
+
+            $ _BOWTIE_COMPLETE=fish_source bowtie | source
+
+For Homebrew users, the shell completion is **automatically enabled** after running the Homebrew tap installation command.
```
Julian, line 51: Obviously all good if you don't have one handy and we don't do this now, but then we definitely shouldn't claim we did in the docs.

docs/cli.rst
```diff
@@ -18,6 +18,38 @@ Below are a few sample command lines you might be interested in.
     If you have cloned Bowtie to :file:`/path/to/bowtie` you should be able to use ``$(ls /path/to/bowtie/implementations/ | sed 's/^| /-i /')`` in any command to expand out to all implementations.
     See `below <cli:running the official suite across all implementations>` for a full example.
 
+Add tab completion for implementations (and otherwise)
+------------------------------------------------------
+
+`Click Shell Completion Reference`_
+
+.. tabs::
```
Julian, line 26: Like the click docs say, we should probably prefer suggesting that one saves the outputted contents here on their completion path rather than re-running it on each shell invocation, so maybe add a second line to each of these corresponding to the other box in the click docs.

docs/conf.py
```diff
@@ -76,6 +77,8 @@
 
 
 .. _official test suite: https://github.com/json-schema-org/JSON-Schema-Test-Suite
+
```
Julian, line 80: No need for this, the click docs are written in Sphinx, so normal intersphinx references will work as above.

Julian (commented):
Thanks!

Julian:
> We could add shell completions for languages as well ? and/or any other completions you've in mind ?

Yeah languages would likely be nice, not sure if any others come to mind.

adwait-godbole:
@Julian done the fixes 👍 and added completions for languages as well. Here's a demo :). Thanks!

https://github.com/bowtie-json-schema/bowtie/assets/68469605/3179513d-5714-41ae-b417-a4dd1e36cf37

docs/spelling-wordlist.txt
```diff
@@ -23,6 +23,7 @@ shortnames
 subcommand
 subcommands
 validator
+config
```
adwait-godbole, line 26: This was failing in `docs(spelling)` checks because of this line

`Add this to ~/.config/fish/completions/bowtie.fish:`

adwait-godbole:
@Julian done the fixes 👍 and added completions for languages as well. Here's a demo :). Thanks!

https://github.com/bowtie-json-schema/bowtie/assets/68469605/3179513d-5714-41ae-b417-a4dd1e36cf37

adwait-godbole:
@Julian did the fixes you suggested 👍 and later realized that completions for languages is enabled by default after activating the completion script and there is no need of implementing custom `shell_complete` function logic for them. 

Here's a demo :). Thanks!

https://github.com/bowtie-json-schema/bowtie/assets/68469605/3179513d-5714-41ae-b417-a4dd1e36cf37

adwait-godbole:
> Can you maybe just open an issue in the Bowtie homebrew tap repo and we can come back to it later?

Yup sure.

adwait-godbole:
> Actually probably even more common now (just from memory of having installed something) is that they simply print out a post-install message telling you what to do.

@Julian are you possibly referring to something of this sort 

https://stackoverflow.com/questions/13333585/how-do-i-replay-the-caveats-section-from-a-homebrew-recipe

Julian:
Exactly!

PR #1154, opened 2024-05-07
Re-enabling deploying the new test reports every time the scheduled report workflow reruns

### LINKED ISSUE #1133: Re-enable updating the deployed report data every time the workflow reruns
A few weeks back I made it so that the UI only redeploys on releases, just like the backend, just to prevent instability and/or the two of them going out of sync.

But that also made test reports only be redeployed when a release happens, so even though we release every few days, that's every few days instead of every few hours that we show new results.

We should likely decouple the two, so new data is published each time it's calculated, even if/while we're not redeploying the actual JS code as often.

adwait-godbole:
Fixes #1133 

Hi @Julian I think this should fix the issue

----
📚 Documentation preview 📚: https://bowtie-json-schema--1154.org.readthedocs.build/en/1154/

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -168,4 +168,6 @@ jobs:
     uses: ./.github/workflows/deploy-frontend.yml
     with:
       report_artifact_in_scope: true
-    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+    if: >
+      (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')) ||
+      (github.event_name == 'schedule')
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -168,6 +168,3 @@ jobs:
     uses: ./.github/workflows/deploy-frontend.yml
     with:
       report_artifact_in_scope: true
-    if: >
-      (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')) ||
-      (github.event_name == 'schedule')
--- a/.github/workflows/deploy-frontend.yml
+++ b/.github/workflows/deploy-frontend.yml
@@ -13,19 +13,20 @@ on:
 jobs:
   build-frontend:
     uses: ./.github/workflows/build-frontend.yml
+    if: inputs.report_artifact_in_scope == false
+
   collect:
     runs-on: ubuntu-latest
     needs: build-frontend
     steps:
-      - name: Create _site directory
-        run: mkdir _site
-      - name: Download frontend build
+      # if called from report workflow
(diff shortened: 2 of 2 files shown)


Julian:
This seems like it will not just deploy the new data it will also revert completely to the previous behavior and redeploy the whole frontend.

adwait-godbole:
Ohh right! I get the issue now so you meant only this artifact should get deployed and not the frontend builds (they only get deployed on new release) ? Correct ?

adwait-godbole:
@Julian if let's say the test reports get computed and want to get deployed so do we assume that there already exists a `frontend-build` deployed to GitHub pages prior to that. If so then I assume the below fix/hack would not rebuild the frontend and only deploy the test reports.

I mean In `deploy-frontend.yml` file we could do this ?

```yaml
jobs:
  build-frontend:
    if: inputs.report_artifact_in_scope == true
    uses: ./.github/workflows/build-frontend.yml
```

adwait-godbole:
But I guess the above would still **download** the `frontend-build`  that already exists (not rebuild it by triggering `build-frontend.yml`) and deploy the test reports along with it.

Julian:
If I follow yeah I think that sounds right, basically download the existing frontend build and use that one.

adwait-godbole:
@Julian updated the PR!

.github/workflows/report.yml
```diff
@@ -168,4 +168,6 @@ jobs:
     uses: ./.github/workflows/deploy-frontend.yml
     with:
       report_artifact_in_scope: true
-    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+    if: >
```
Julian, line 171: You can just remove the if entirely now no?

adwait-godbole: okay!

Julian:
Cool, let's try this out, thanks!

adwait-godbole:
@Julian in `deploy-frontend.yml` file below here should we default `report_artifact_in_scope` to false ?

```yml
on:
  workflow_call:
    inputs:
      report_artifact_in_scope:
        required: false
        type: boolean
  workflow_dispatch:
  release:
    types: [published]
```

Julian:
Would that change anything other than not needing to pass it explicitly from the `report` workflow? I don't mind either way (if you want to change it sure).

adwait-godbole:
``yml
  build-frontend:
    uses: ./.github/workflows/build-frontend.yml
    if: inputs.report_artifact_in_scope == false

```
I am not sure if this line would run if the workflow was triggered by a release but ig it would definitely run if the default was set to false perhaps ?

PR #1155, opened 2024-05-07
Default `report_artifact_in_scope` to false to be sure build frontend runs if workflow triggered by release published

adwait-godbole:
```yml
  build-frontend:
    uses: ./.github/workflows/build-frontend.yml
    if: inputs.report_artifact_in_scope == false

```
As mentioned in this [comment](https://github.com/bowtie-json-schema/bowtie/pull/1154#issuecomment-2098707093) in PR #1154 I am not sure if the above job would run if the workflow was triggered by a release publish but ig it would definitely run if the default for `report_artifact_in_scope` was set to false perhaps ? So ig let's just default it to false.

@Julian so let's remove `required` as well or set it to `true` explicitly ? since we are already defaulting `report_artifact_in_scope` to false.

### DIFF
--- a/.github/workflows/deploy-frontend.yml
+++ b/.github/workflows/deploy-frontend.yml
@@ -23,7 +23,7 @@ jobs:
         run: mkdir _site
 
       # if called from report workflow
-      - name: Download the previous frontend-build
+      - name: Download latest frontend-build
         uses: dawidd6/action-download-artifact@v3
         if: inputs.report_artifact_in_scope == true
         with:
@@ -41,14 +41,14 @@ jobs:
           path: _site
 
       # if called as a separate workflow
-      - name: Download the frontend-build from local artifact
+      - name: Download frontend-build from local artifact
         uses: actions/download-artifact@v4
         if: inputs.report_artifact_in_scope == false
         with:
           name: frontend-build
           path: _site
 
-      - name: Generate and Download latest test report
+      - name: Download latest test report
         if: inputs.report_artifact_in_scope == false
         uses: dawidd6/action-download-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


adwait-godbole:
Hi @Julian just had a look at the latest collect new test results run at https://github.com/bowtie-json-schema/bowtie/actions/runs/8990475368 and apparently the `collect` and `deploy` job didn't run :(. 

Is it likely because of the `needs` step in [here](https://github.com/bowtie-json-schema/bowtie/blob/6082e03599ad723f238147a10fec6184d0c2fbfe/.github/workflows/deploy-frontend.yml#L21) ?

adwait-godbole:
@Julian sorry to ping again, can you have a look ?

Julian:
It looks suspicious to me -- did you try it in a fork? Not having that needs line will mean the second job starts immediately, even when `inputs.report_artifact_in_scope == true`, i.e. you will be changing behavior even in the case we don't mean to change?

adwait-godbole:
@Julian sorry I had to delete my fork in here but for some reason the runs were consistently failing in my workflows so I am trying to refork and continue but nonetheless I think what you are saying definitely makes sense here and I think these docs look promising --> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs and would help us tackle this one. Let me know what you think about this meanwhile I am running the workflows again on my fork with these changes.

adwait-godbole:
Screenshot (145)


@Julian on my fork the `site` job keeps failing because of the reason mentioned in the below screenshot and due to this it is unable to trigger the `deploy-frontend` job from Collect  New Test Results (`report.yml`) since it has a **needs** dependency on `site` but nonetheless triggering deploy-frontend job manually is skipping the ones which were expected to be skipped and working fine.

PR #1163, opened 2024-05-08
Default `report_artifact_in_scope` to false to be sure build frontend runs if workflow triggered by release published

adwait-godbole:
Default `report_artifact_in_scope` to false to be sure build frontend runs if workflow triggered by release published and not requiring successful dependent jobs 

Reference Docs:- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs

----
📚 Documentation preview 📚: https://bowtie-json-schema--1163.org.readthedocs.build/en/1163/

### DIFF
--- a/.github/workflows/deploy-frontend.yml
+++ b/.github/workflows/deploy-frontend.yml
@@ -4,8 +4,8 @@ on:
   workflow_call:
     inputs:
       report_artifact_in_scope:
-        required: false
         type: boolean
+        default: false
   workflow_dispatch:
   release:
     types: [published]
@@ -18,15 +18,22 @@ jobs:
 
   collect:
     runs-on: ubuntu-latest
+    if: ${{ always() }}
     needs: build-frontend
     steps:
+      - name: Create _site directory
+        run: mkdir _site
+
       # if called from report workflow
-      - name: Download existing _site artifact having "frontend-build"
-        uses: actions/download-artifact@v4
+      - name: Download latest frontend-build
+        uses: dawidd6/action-download-artifact@v3
         if: inputs.report_artifact_in_scope == true
         with:
-          name: _site
+          workflow: build-frontend.yml
+          branch: main
+          name: frontend-build
           path: _site
+          if_no_artifact_found: warn
 
       - name: Download test report from local artifact
         uses: actions/download-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-frontend.yml
```diff
@@ -18,15 +18,22 @@ jobs:
 
   collect:
     runs-on: ubuntu-latest
+    if: ${{ always() }}
     needs: build-frontend
```
adwait-godbole, line 22: As the docs say, collect uses the always() conditional expression so that it always runs after build-frontend has completed, regardless of whether it was successful.

adwait-godbole: P.S. extremely sorry for these multiple pings 🙏

adwait-godbole: This workflow worked on my fork 👍

Julian: We don't want to continue if it fails though. Just if it succeeds or is skipped, no? We shouldn't try to deploy a known *failed* build.

adwait-godbole: I came across these docs then https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

so instead of `${{ always() }}` should we do `${{ !failure() }}` ? Ig it should then work even if it was successful / skipped

adwait-godbole: @Julian changing it to `${{ !failure() }}` also worked on my fork's workflow and didn't skip the `collect` job even when the `build-frontend` job was skipped.

Julian: Cool, merging.then, let's see.

adwait-godbole: After merge can you manually trigger the Collect new test reports workflow once ? so we can test out.

adwait-godbole:
@Julian ig it worked can you have a look here https://github.com/bowtie-json-schema/bowtie/actions/runs/9006736249/job/24745051174. I mean although `build-frontend` was skipped which was what was expected and `collect` job collected the latest frontend-build but I am not sure / will have to investigate why the `deploy` job was skipped.

PR #1172, opened 2024-05-13
One more try at renabling only test reports deployment

adwait-godbole:
@Julian these are the changes I had employed that I thought should've worked and below is the ss that gave me that validity that yes it's working as desired.

Screenshot (148)

----
📚 Documentation preview 📚: https://bowtie-json-schema--1172.org.readthedocs.build/en/1172/

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -134,11 +134,11 @@ jobs:
       - name: Create our Site Structure
         run: mkdir site
 
-      # - name: Include Previous Report
-      #   uses: actions/download-artifact@v4
-      #   with:
-      #     name: previous-reports
-      #     path: site/previous/
+      - name: Include Previous Report
+        uses: actions/download-artifact@v4
+        with:
+          name: previous-reports
+          path: site/previous/
 
       - name: Include New Reports
         uses: actions/download-artifact@v4


.github/workflows/deploy-frontend.yml
```diff
@@ -4,28 +4,37 @@ on:
...
+      # if called from report workflow
+      - name: Download latest frontend-build
+        uses: dawidd6/action-download-artifact@v3
+        if: inputs.report_artifact_in_scope == true
         with:
+          workflow: build-frontend.yml
+          branch: main
           name: frontend-build
           path: _site
+          if_no_artifact_found: warn
```
adwait-godbole, line 37: I am most concerned about this. Is this the correct way of downloading whatever last frontend-build was available ?

adwait-godbole: It seems to be downloading it correctly from the ss above.

Julian:
I'm still a bit confused -- I don't really care about screenshots :), I believe you it worked when you tried it but clearly something was different when it was merged here -- are you saying you did something different here / are those screenshots something that didn't work before and do work now? Is anything different about this PR than before or is it the same code changes essentially?

adwait-godbole:
> I'm still a bit confused -- I don't really care about screenshots :), I believe you it worked when you tried it but clearly something was different when it was merged here -- are you saying you did something different here / are those screenshots something that didn't work before and do work now? Is anything different about this PR than before or is it the same code changes essentially?

@Julian Yeah I agree I haven't been able to find/justify what went wrong the last time but I am kind of in an opinion here right now that maybe multiple different PRs/workflow runs might have caused some mismatch in the data/artifacts that were generated and I think we should give this one more try tbh. This time let's be careful and observant on the `github-pages` artifact that gets generated as I believe it is going to help us in debugging the error if something goes wrong this time as well. Last time I forgot to check this artifact and later it got expired :(. What do you say ?

Julian:
Got it, fine with me, let's see!

.github/workflows/deploy-frontend.yml
```diff
@@ -4,28 +4,37 @@ on:
...
-      - name: Download frontend build
-        uses: actions/download-artifact@v4
+
+      # if called from report workflow
+      - name: Download latest frontend-build
+        uses: dawidd6/action-download-artifact@v3
+        if: inputs.report_artifact_in_scope == true
         with:
+          workflow: build-frontend.yml
+          branch: main
```
adwait-godbole, line 33: @Julian the site seems to be functioning for now and I've downloaded the `github-pages` artifact that was generated. However I observed one thing that my fix of removing IMAGE_PREFIX from frontend did not show up on the website as in the above code downloading the latest frontend-build from `main` branch of `build-frontend.yml` did not have that fix of mine. So what do you think should be the best place to get the last / latest frontend-build from ?

adwait-godbole: I think this might also be the cause of previously failed efforts

Julian: It should probably always use the current branch for the workflow (which I'm guessing but haven't checked should be the default if you don't specify branch) but then we should never deploy from anywhere other than main

adwait-godbole: Yeah I checked if you dont specify any branch explicitly then it uses the current branch. So do you mean i should remove `branch: main` from here ?

Julian: Yes just double check that somewhere below only runs if it's main

adwait-godbole: Meaning ?

Julian: What I mentioned above -- it's fine to build every branch using its own workflow, but we should be sure to only finish / deploy if we're running on main.

adwait-godbole: Okay got it!

adwait-godbole: > It should probably always use the current branch for the workflow (which I'm guessing but haven't checked should be the default if you don't specify branch) but then we should never deploy from anywhere other than main

but I am guessing the problem here is that the current branch won't necessarily have ran the build-frontend workflow since it is purely relying on it to just get the artifacts that have already been ran from some other branch right ?

adwait-godbole: Sorry nevermind I figured it out :) Thanks!

PR #1173, opened 2024-05-13
Add container connectable

### LINKED ISSUE #874: Support attaching to a running implementation harness rather than starting one
It would be nice to be able to attach to an externally managed harness container rather than having Bowtie start and stop them.

This would be useful for #35 or generally for performance-related use cases, as it makes it slightly easier to ignore time spent starting containers.

Also relates to #19 in that generalizing here will help that and/or rely on it.

sudo-jarvis:
Closes #874.

This PR adds support for the connector connectable using which bowtie would be able to connect to already running implementation containers as well.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1173.org.readthedocs.build/en/1173/


sudo-jarvis:
@Julian Please have a look.

bowtie/_cli.py
```diff
@@ -37,6 +37,7 @@
     Dialect,
     Example,
     Implementation,
+    NoSuchContainer,
```
Julian, line 40: I'm in transit still today and tomorrow so I'll give this a closer read as soon as I can, but

1) you're definitely on the right track
2) try to cleanly separate between which modules have to know about your new thing -- in other words, the connectable nonsense is meant precisely to make sure that a module like this one does not need to know *how* it is speaking to implementations.

More specifically -- defining a new exception (`NoSuchContainer`) is likely perfectly fine, but catching it / letting it bubble up all the way here isn't -- by the time this module deals with it it should likely just be dealing with `NoSuchImplementation`, and something should have converted the `NoSuchContainer` into that -- basically at this module's layer it should simply be dealing with an exception which means "the user tried to talk to some connectable thing and that thing doesn't exist, regardless of whether it's an image name, a container id, whatever" -- I *think* `NoSuchImplementation` is already suited to be that but if not we can tweak it slightly. But yeah try to get your nice error message shown but without making it bubble all the way up here.

sudo-jarvis: Yeah you are right, I get it now. I'll display the message using the existing generic Exception (```StartuFailed```) in case connection establishment to a container fails.

bowtie/_connectables.py
```diff
@@ -76,6 +78,7 @@ class Connectable:
         (cls.connector, cls)
         for cls in [
             _containers.ConnectableImage,
+            _containers.ConnectedContainer,
```
Julian, line 81: Can you make the name of this match the other one? I.e. `ConnectableContainer` not `Connected` (the connection hasn't happened yet by the time this object is created).

bowtie/_containers.py
```diff
@@ -150,10 +150,9 @@ class Connection:
     handled here.
     """
 
-    _image: str = field(alias="image")
-
     _docker: aiodocker.docker.Docker = field(repr=False, alias="docker")
-    _stream: Stream = field(default=None, repr=False, alias="stream")
+    _image: str = field(default=None, alias="image")
+    stream: Stream = field(default=None, repr=False, alias="stream")
```
Julian, line 155: I'm suspicious of this being made public, it seems like a recipe for bugs (calling `attach_stream` twice for example).

It seems like what you have here is the reverse of what I was suggesting -- specifically it seems like you made `Connection` stay as the connection for `ConnectableImage` and you introduced `ManagedConnection` for `ConnectedContainer` -- that's the opposite of what I'd have expected, I'd have expected that a connection for a *container* is the more basic thing now, and that it's `ConnectableImage` that needs to do extra work (specifically it needs to make sure the image and container is deleted, whereas ConnectableContainer should do none of that).

But beyond being different from what I expect I'm not sure I see any bugs yet. I'm suspicious though that the tests are possibly hiding them (because of your `suppress()`) but I haven't stared long enough yet.

tests/test_connectables.py
```diff
@@ -110,3 +114,12 @@ def test_terse_implicit_image_no_repository():
     id = "bar"
     validator.validate(id)
     assert Connectable.from_str(id).to_terse() == "bar"
+
+
+def test_container_connectable():
+    id = "container:bar"
```
Julian, line 120: Use some realistic value here please -- (i.e. at least make it be some hex string).

You also should modify the JSON Schema for `connectables` to validate that kind of thing, and have a test which triggers that validation (see what the existing test does for that).

tests/test_integration.py
```diff
@@ -98,6 +98,35 @@ def tar_from_directory(directory):
     return fileobj
 
 
+def container(name, fileobj):
```
Julian, line 101: This seems like it copy/pastes `image()`. Presumably there should be a way for you to call `image()` instead, though I'm not 100% sure what the syntax is given we're dynamically generating those fixtures. But yeah have a look quickly at least?

sudo-jarvis: Yeah actually I tried doing this only but the value which image() is returning is a function which is leading into an error while creating the json containing data for container to get started.

I'll look into it further and see if there is a way in which this can be done.

bowtie/_containers.py
Julian: Not inheriting, composing, but otherwise yes.

sudo-jarvis: @Julian, Okay then this can be done within `ConnectableImage` itself then ig rather than having a separate `ManagedConnection` ?

Julian: Yes that seems reasonable, `ConnectableImage.connect` can probably do it if I'm guessing right -- it should really just be two or three lines of extra wrapping code.

sudo-jarvis: But then the functions like `_start_container()`, `_start_container_maybe_pull()` would also be present in `ConnectableImage` making it incompatible with the `Connector` protocol no? I believe these functions shouldnt be a part of `Connection` now right? `Connection` should manage connections with running containers irrespective of who started it.

Julian: I'm not following maybe because I haven't tried to make the change myself -- but `_start_container` and friends certainly shouldn't be part of the connector protocol, they're things specific to just container-connections not arbitrary ones. So they should be part of the concrete object which is responsible for "container connections" (which should overall be implementing the general `Connection` protocol) but not the general protocol. Does that answer your question or have I misunderstood it?

Julian: Perhaps slight confusion is because we have two things called `Connection`? There's `bowtie._core.Connection` (the protocol) and `bowtie._containers.Connection` (the concrete implementation of the protocol which is right now specific to image creation, but which we're discussing make instead be tied to container communication, and then wrapping that with a further layer when we *also* need to do image operations)

sudo-jarvis: Yeah I was talking about the latter one i.e. `bowtie._containers.Connection`. Since this would be a general communication tied to container communication. `_start_container()` and other stuff perhaps shouldnt be a part of it. Now the next question is where should they be ? If that stuff is now put into `bowtie._containers.ConnectableImage` then it would make it violate the `Connector` protocol no ?

We could instead have the `Connection` wrapper for image creation in something like `bowtie._containers.ManagedConnection`. Then `connectableImage.connect` can simply use this.

Julian: > If that stuff is now put into bowtie._containers.ConnectableImage then it would make it violate the Connector protocol no?

No, a concrete class may always have *more* methods than the protocol, especially ones private to its own implementation, that's precisely the point of a protocol -- it defines the methods that something may *rely* on any concrete implementation having vs. those which cannot be relied on (or don't even make any sense in general). `_start_connection` is an example -- that would not be part of the protocol, but it may certainly appear on some  concrete class implementing the protocol.

All of which is to say I'm not 100% sure that `ConnectableImage` is going to be the right place without trying it, but it would *need* to be the place if you're going to try what you suggested earlier (not having a `ManagedConnection` class at all, and just putting the extra stuff inside of `ConnectableImage`).

sudo-jarvis: Thanks, got it. Another thing to figure out would be that currently we support restarting the connection if it fails. So currently what `bowtie._containers.Connection` does is as it only supports image connectable, it simply recreates a container, but this would not work when its dealing with an already running container.

Julian: Yes that's another good point -- and one I noticed previously -- the right "long term" solution to that is really to have a `ReconnectingConnection` object which takes *any* object implementing the `Connection` protocol along with a function that implements "restart", and is specific to the specific concrete connection (and then you have a trivial implementation which does nothing at all when used with containers, which cannot be restarted).

But I think it's best if we avoid worrying about that too much right now -- I think I'd try *also* having that logic be only in `ConnectableImage` -- but if that gets complicated we can look at it together.

sudo-jarvis: I have tried writing the code and everything else works but I am getting stuck at figuring out a way to implement the restarting mechanism within `ConnectableImage` itself as its just a wrapper and once it creates the `Connection` it basically has no role so getting back to it such that we can get to the image and restart the container seems to be not feasible in this way.

Julian: Want to push what you have to a branch so I can have a look? Feel free to have it be a different one in case we stick with the code as it is here in this form.

sudo-jarvis: @Julian, I have pushed the code here: (https://github.com/sudo-jarvis/bowtie/blob/test-container-connectable/bowtie/_containers.py#L145)

Please have a look once and share your insights.

Julian:
Merged your changes along with the additional ones we were discussing in the comment thread.

Thanks for getting this most of the way!

PR #1176, opened 2024-05-14
Removing `branch: main` dependence to get the latest frontend-build

adwait-godbole:
@Julian I think this should finally work now as per our needs! :)

----
📚 Documentation preview 📚: https://bowtie-json-schema--1176.org.readthedocs.build/en/1176/


adwait-godbole:
@Julian can you trigger the workflow once so I can check if this worked by seeing if the IMAGE_PREFIX fix reflects on the site or not.

PR #1182, opened 2024-05-17
Refactoring workflows

adwait-godbole:
@Julian @AgniveshChaubey this is my first iteration on this change. Let's discuss this further if this seems like the right way moving forward or not.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1182.org.readthedocs.build/en/1182/


.github/workflows/deploy-frontend.yml
```diff
@@ -4,36 +4,92 @@ on:
...
+
+      # if called from the report workflow and it wasn't a new release event (it was either a cron job or a manual dispatch of the report workflow)
+      - name: Download frontend-build from last release
+        uses: dawidd6/action-download-artifact@v3
+        if: inputs.report_artifact_in_scope == true && inputs.release_published == false
         with:
+          workflow: ci.yml
+          event: release
           name: frontend-build
           path: _site
```
adwait-godbole, line 76: Mostly concerned about this one! Does it seem like the correct way ?

AgniveshChaubey: I am not sure why you used `ci.yml` workflow here. But if you have used the `ci.yml` here, then the `frontend-build` artifact must be available in that workflow else this workflow run will fail. And as far as I can see, no `frontend-build` artifact is present in the `ci.yml`

AgniveshChaubey: Have you tested the workflow? Is it running properly?

.github/workflows/deploy-frontend.yml
```diff
@@ -4,36 +4,92 @@ on:
...
+    paths:
+      - "frontend/**"
+      - ".github/workflows/deploy-frontend.yml"
 
 jobs:
+  # only build if triggered manually or new release published or a PR
   build-frontend:
-    uses: ./.github/workflows/build-frontend.yml
+    runs-on: ubuntu-latest
+    if: inputs.report_artifact_in_scope == false || inputs.release_published == true || github.event_name == 'pull_request'
```
adwait-godbole, line 22: here we build the frontend during a release workflow call chaining.

.github/workflows/deploy-frontend.yml
adwait-godbole: > I am not sure why you used ci.yml workflow here. But if you have used the ci.yml here, then the frontend-build artifact must be available in that workflow else this workflow run will fail. And as far as I can see, no frontend-build artifact is present in the ci.yml.

This was the reason for my question on Slack @AgniveshChaubey . As I mentioned there I had observed that whenever a new release happens only `ci.yml` runs and the chaining of workflow happens within it. So at that time when a new release happens we build the frontend just like [here](https://github.com/bowtie-json-schema/bowtie/pull/1182/files#r1605786645) and because of workflow call chaining starting from the parent `ci.yml` this `deploy-frontend.yml` also gets triggered anyways but under the context of `ci.yml`. So that's the reason above here I wrote `ci.yml` to find the last frontend-build release.

adwait-godbole: > Have you tested the workflow? Is it running properly?

Regarding this I've some issues. For testing this on my fork I wanted some release run of `ci.yml` on my fork to test but since I don't have one by default unlike here in Bowtie repo where multiple release runs of ci.yml are present, can you help me figure out how can I trigger a release on my fork so that `ci.yml` runs ? cc @Julian

.github/workflows/deploy-frontend.yml
```diff
@@ -4,36 +4,92 @@ on:
...
+      - name: Install dependencies
+        run: pnpm install --dir frontend --frozen-lockfile --strict-peer-dependencies
+
+      - name: Build the UI
+        run: pnpm run --dir frontend build --base=${{ github.repository_owner == 'bowtie-json-schema' && '/' || '/bowtie' }}
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: frontend-build
+          path: frontend/build
```
adwait-godbole, line 57: here we upload during workflow call chaining run

Julian:
(Discussed via Slack, we'll revisit this at some point)

PR #1186, opened 2024-05-20
Update dotnet-jsoneverything-net registry lookup for lib v7

gregsdennis:
Fixed an issue with registry lookups.

.Net `Uri` comparisons don't consider the fragment, and v7 now passes the complete URI into the `Fetch` function.  When Bowtie `.ToString()`s the URI to perform a string lookup, it includes the fragment and (rightly) can't find it.

This change updates the registry lookup so that it's a `Uri` key and the fragment will be ignored.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1186.org.readthedocs.build/en/1186/

### DIFF
--- a/implementations/dotnet-jsonschema-net/Program.cs
+++ b/implementations/dotnet-jsonschema-net/Program.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.Json;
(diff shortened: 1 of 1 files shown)


Julian:
(The audit stuff is unrelated obviously, that's a new vulnerable dep presumably, which I'll fix after merge.)

Thanks!

PR #1201, opened 2024-05-21
Added local metaschema

mwadams:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1201.org.readthedocs.build/en/1201/

### DIFF
--- a/implementations/dotnet-corvus-jsonschema/Program.cs
+++ b/implementations/dotnet-corvus-jsonschema/Program.cs
@@ -69,40 +69,40 @@ static JsonSchemaTypeBuilder CreateTypeBuilder()
                         JsonDocument.Parse(File.ReadAllText("./metaschema/draft7/schema.json")));
 
     builder.AddDocument("https://json-schema.org/draft/2019-09/schema",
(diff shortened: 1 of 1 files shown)


Julian:
pre-commit is failing because you keep undoing clang-format's changes :) -- I don't care what formatting is used, but please either:

* tweak the settings so clang-format emits what you want
* leave it how clang-format puts it
* change to some other pre-commit hook that works the way you want

where I presume the above is in ascending order of effort*value.

Julian:
(Normally pre-commit would straight up push a commit to revert what you have, but because your PR comes from an organization it does not have permissions to do so).

mwadams:
> (Normally pre-commit would straight up push a commit to revert what you have, but because your PR comes from an organization it does not have permissions to do so).

Apologies - I realise I keep doing this; clang-format is not part of my toolchain and I have to remember to run it manually.

Julian:
(If you install pre-commit run `pre-commit install` it will run automatically on commit.) Otherwise yeah I don't mind if you wait till push to find out, just making sure you weren't confused by the failures!

Issue #1185, opened 2024-05-20
Track bowtie-found bugs in the docs


Julian:
We should have a page showing implementation bugs found via Bowtie. There are a number we've found already.

adwait-godbole:
Would like to give this a try. How do we approach this one? Is it possible to dynamically render this content on the docs?

Julian:
Actually the tricky bit here is that the few examples I'm aware of are somewhere in my Slack DMs :) -- I filed this to remind me to put them somewhere public, but yeah that's the tricky part about you doing it. There's nothing to dynamically render really, I just wanted a place we could have a "hall of bugs" kind of thing to show we're finding real-world bugs (which we've already found 4 I think).

cbum-dev:
Hi @Julian can I work on this issue?

PR #1205, opened 2024-05-21
Fixed bowtie test errors

mwadams:
1. Support generating code with "do not assert format" for 2020-12
2. Add missing `unevaluated` metaschema for 2020-12
3. Added .vs folder to the .gitignore section for "editor cruft"

----
📚 Documentation preview 📚: https://bowtie-json-schema--1205.org.readthedocs.build/en/1205/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,4 @@ cython_debug/
 # Editor vomit
 .idea/
 .vscode/
+.vs/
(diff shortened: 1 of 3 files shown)


Julian:
> Support generating code with "do not assert format" for 2020-12

Nice! Was this a bug found by Bowtie perchance? (Eventually I'll collect those in #1185)

Feel free to merge when CI passes (you've got the power).

mwadams:
> > Support generating code with "do not assert format" for 2020-12
> 
> Nice! Was this a bug found by Bowtie perchance? (Eventually I'll collect those in #1185)
> 
> Feel free to merge when CI passes (you've got the power).

Originally, it was "by design" - for all our uses we only ever wanted to generate format-asserting code.

But it was a bit rude of us not to support the standard 2020-12 dialect out-of-the-box!

Doing the bowtie integration has forced us to get off our backsides and do it properly.

So this is well worth collecting under #1185. Possibly even more importantly than "bugs" it is "lazy implementers doing things their own way".

PR #1209, opened 2024-05-24
`bowtie statistics` defaulting to retrieving latest dialect's latest report from published website

### LINKED ISSUE #1206: `bowtie statistics` with no stdin input should default to retrieving a report from the published website
I.e. `bowtie statistics` should be equivalent to `http get 'https://bowtie.report/draft2020-12.json' | bowtie statistics`.

adwait-godbole:
Fixes #1206 

Screenshot (158)

Screenshot (159)

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Callable, Iterable, ValuesView
+from collections.abc import Callable, Iterable
 from contextlib import AsyncExitStack, asynccontextmanager
 from fnmatch import fnmatch
 from functools import wraps
@@ -779,6 +779,79 @@ def _validation_results_table_in_markdown(
 
     return final_content
 
+def _get_latest_dialect_report(
+    ctx: click.Context,
+    _,
+    report: _report.Report | None,
+):
+    if report is None:
+        latest_dialect = Dialect.by_short_name()["draft2020-12"]
+        return (
+            latest_dialect.pretty_name,
+            asyncio.run(latest_dialect.latest_report()),
+        )
+    return None, report
+
+@subcommand
+@format_option()
+@click.option(
+    "--quantiles",
+    "n",
+    default=4,
+    type=int,
+    help=(
+        "How many quantiles should be emitted for the compliance numbers? "
+        "Computing quantiles only is sensical if this number is more than the "
+        "number of implementations reported on. By default, we compute "
+        "quartiles."
+    ),
+)
+@click.argument(
+    "report_data",
+    default="-" if not sys.stdin.isatty() else None,
+    required=False,
+    type=_Report(),
+    callback=_get_latest_dialect_report,
+)
+def statistics(
+    report_data: tuple[str | None, _report.Report],
+    n: int,
+    format: _F,
+):
+    """
+    Show summary statistics for a previous report.
+
+    If no report provided, defaults to Bowtie's latest dialect's latest report.
+    """
+    dialect, report = report_data
+    unsuccessful = report.compliance_by_implementation().values()
+    statistics = dict(
+        **(  # latest dialect if no report was provided
+            {"latest_dialect": dialect}
+            if dialect
+            else {}
+        ),
+        median=median(unsuccessful),
+        mean=mean(unsuccessful),
+        **(  # quantiles only make sense for n < len(data)
+            {"quantiles": quantiles(unsuccessful, n=n)}
+            if n < len(unsuccessful)
+            else {}
+        ),
+    )
+    match format:
+        case "json":
+            click.echo(json.dumps(statistics, indent=2))
+        case "pretty":
+            for k, v in statistics.items():
+                click.echo(f"{k}: {v}")
+        case "markdown":
+            markdown = _convert_table_to_markdown(
+                columns=["", ""],
+                rows=[[k, str(v)] for k, v in statistics.items()],
+            )
+            click.echo(markdown)
+
 
 def make_validator():
     validators = validator_registry()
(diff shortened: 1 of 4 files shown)


adwait-godbole:
Also we should likely start showing some helpers in the shell 😅. Maybe something like "Downloding.....100%" or "Testing.....100%". Otherwise for e.g. when a user runs `bowtie statistics --dialect 2020` or any other longer running time command, it gets difficult to know if the command is running in the background or some error has occured.

adwait-godbole:
@Julian can you also review this one.

Julian:
Oh, now I see/remember why I didn't add `--dialect`, it makes no sense when providing a report. Specifically `cat "somereport | bowtie statistics --dialect draft7` makes no sense, the report is in whatever dialect the report is in.

I have to think a bit again, but maybe it instead makes sense to put the `--dialect` argument in a `bowtie latest-report --dialect <DIALECT>` which is a command that simply retrieves the latest report for that dialect from the site.

But yeah think this needs slightly more thought otherwise there are commands which make no sense potentially.

adwait-godbole:
Yes agreed with that! `cat "somereport | bowtie statistics --dialect draft7` doing this definitely doesn't make sense and I initially thought to just show some sort of a `click.UsageError` but discarded it later. Nevertheless doing `bowtie latest-report --dialect <DIALECT>` does make sense but are we displaying the stats in this command then ? How do i proceed further / or put this on hold ?

Julian:
> are we displaying the stats in this command then

No I think then we can simply do `bowtie latest-report ... | bowtie statistics` to get the statistics for a specific dialect maybe (and still default `bowtie statistics` to get the latest report for the latest dialect like it does now maybe).

adwait-godbole:
Cool got it! but then just doing `bowtie latest-report --dialect 2020` just outputs the report onto the console ?

adwait-godbole:
And amidst all this we still default just doing `bowtie statistics` to show stats for 2020-12, right ? just what the issue #1206 says ?

adwait-godbole:
do we take a flag for `bowtie latest-report --dialect 2020` or just do `bowtie latest-report 2020` as an argument ?

Julian:
> And amidst all this we still default just doing bowtie statistics to show stats for 2020-12, right ? just what the issue https://github.com/bowtie-json-schema/bowtie/issues/1206 says ?

Yep! 

> Also do we take a flag for bowtie latest-report --dialect 2020

Yeah may as well for consistency with the other commands

Julian:
Not yet because our report format doesn't yet support combining multiple dialects but once we do we can beef it up!

adwait-godbole:
Cool got it! working on this 👍!

adwait-godbole:
Hi @Julian I've updated the PR, is it fine ?

bowtie/_cli.py
```diff
@@ -780,6 +780,20 @@ def _validation_results_table_in_markdown(
     return final_content
 
 
+def _get_latest_dialect_report(
+    ctx: click.Context,
+    _,
+    report: _report.Report | None,
+):
+    if report is None:
+        latest_dialect = sorted(Dialect.known(), reverse=True)[0]
```
Julian, line 789: This is `max(Dialect.known())`.

Also I think you can probably get away with *just* returning the report (and therefore not needing a callback at all, just the `default`) because I think you can get the dialect from `report.metadata.dialect` if I remember right (I haven't double checked).

adwait-godbole: > Also I think you can probably get away with just returning the report (and therefore not needing a callback at all, just the default) because I think you can get the dialect from report.metadata.dialect if I remember right (I haven't double checked).

Okay @Julian, I tried doing it as you suggested i.e. using `report.metadata.dialect` but it seems to me that during the tests, `envsonschema` and `always_valid`  fake implementations or any other for that matter (e.g. even a real one like `python-jsonschema`) are adding a `Draft 2020-12` default dialect from their own side (reference:- https://github.com/bowtie-json-schema/bowtie/blob/84b56f153711bc001fb31a06abc9082cda5a866f/bowtie/_cli.py#L942) . Because of this `report.metadata.dialect` will always hold a dialect so I will have to change the tests [here](https://github.com/bowtie-json-schema/bowtie/blob/84b56f153711bc001fb31a06abc9082cda5a866f/tests/test_integration.py#L2259) to also include a new dialect line. Is it fine? I was doing this inclusion of a dialect line over here just so that if someone does just `bowtie statistics` then he/she knows that it was for the latest `draft2020` dialect report of the website and not just some random numbers on the console.

adwait-godbole: > Oh, now I see/remember why I didn't add --dialect, it makes no sense when providing a report. Specifically cat "somereport | bowtie statistics --dialect draft7 makes no sense, the report is in whatever dialect the report is in.

Ofcourse I think you saying this will still hold true if the report is in whatever dialect it is (and that one will then appear in the `statistics` output) just that if None found then the implementations will assume / default to `Draft 2020-12` dialect.

adwait-godbole: Also I think you can probably get away with just returning the report (and therefore not needing a callback at all, just the default) because I think you can get the dialect from report.metadata.dialect if I remember right (I haven't double checked).

Okay @Julian, I tried doing it as you suggested i.e. using report.metadata.dialect and it works 👍. However, I just got one query, is there a way we can print/echo on the console that if someone just does bowtie statistics then whatever numbers he/she is seeing on the console are of the latest report of the latest dialect from the website? or should we just leave it be as a helper text on the command?

Julian: > is there a way we can print/echo on the console that if someone just does bowtie statistics then whatever numbers he/she is seeing on the console are of the latest report of the latest dialect from the website? 

I think I like what you were doing here -- you added a column that showed what dialect the report was for, I think you can also add one for the start time of the report and then that serves the purpose you mention.

Maybe we show a header that contains the report date and dialect, and then in the body we show the actual statistics.

adwait-godbole: > Maybe we show a header that contains the report date and dialect, and then in the body we show the actual statistics.

for all formats we show a header ? (json, pretty, markdown)

adwait-godbole: > Maybe we show a header that contains the report date and dialect, and then in the body we show the actual statistics.

@Julian How do we make changes in the tests for this 😅 ?

adwait-godbole: I tried doing `Report.from_serialized(run_stdout)` in the tests to get the report date and dialect from the metadata (for assertion purposes) but I am getting a `JSONDecodeError` 😅.

Julian: `Report.from_serialized` is what you use if you have a report serialized somewhere and want to deserialize it -- we don't have that here, stdout will contain some simple statistics and a small header so you can just assert directly against it I think -- or for the JSON output it should be a JSON output and you assert against the deserialized JSON. Does that help?

adwait-godbole: No I mean the problem is I was trying to do `Report.from_serialized` because in the header we have report date + dialect. So how do we assert against the report date ?

Julian: I again don't see what that function has to do with anything, your output is distinctly not a report.

Are you asking how to handle the fact that the date will not be some fixed time? The easiest way to handle that is probably to parse it manually using the datetime module and assert it's within 5 minutes of `now() `say, and then to just treat it as correct.

adwait-godbole: Yeah sorry I wasn’t able to phrase my question correctly but yeah that’s what I meant. Thanks!

adwait-godbole:
@Julian can you review again please. Also the CI is failing because of some timeout issue in `filter_implementations_no_arguments`. Nothing related to this PR.

bowtie/_cli.py
```diff
@@ -809,13 +828,25 @@ def statistics(report: _report.Report, n: int, format: _F):
             else {}
         ),
     )
+    header = (
+        f"Dialect: {report.metadata.dialect.pretty_name}\n"
+        f"Last ran on: "
+        f"{report.metadata.started.strftime(USER_LOCALE_DT_FMT)}"
```
Julian, line 834: Are you looking for [`%x`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)? There's no need to touch the `locale` module directly for this.

bowtie/_cli.py
```diff
@@ -809,13 +828,25 @@ def statistics(report: _report.Report, n: int, format: _F):
...
+    )
     match format:
         case "json":
+            statistics = {
+                "dialect": str(report.metadata.dialect.uri),
+                "last_ran_on": str(report.metadata.started),
```
Julian, line 840:
```suggestion
                "last_ran_on": report.metadata.started.isoformat(),
```

tests/test_integration.py
```diff
@@ -2294,13 +2344,28 @@ async def test_statistics_json(envsonschema, always_valid):
         json=True,
     )
 
+    last_ran_on_str = jsonout["last_ran_on"]
+    assert last_ran_on_str is not None, "Last ran on date not found in jsonout"
```
Julian, line 2348: This doesn't seem like it will ever be None, if it's missing it'll have raised an exception and errored the test (which is fine).

tests/test_integration.py
```diff
@@ -114,6 +116,37 @@ async def _image(docker):
     return _image
 
 
+async def validate_date_within_mins(
```
Julian, line 119: This doesn't need to be async it can just be a regular function, but also it's pretty simple? I'd just inline it and not have a function at all for either of these two.

adwait-godbole:
@Julian made the suggested changes and again the same test is failing because of timeout exceeded (Nothing related to this PR. )

Julian:
The test error is indeed related, and is caused by how you did `default` in the argument -- specifically, because you didn't make it be a callable, it isn't evaluated lazily, it's evaluated every time Bowtie runs -- meaning literally anytime someone runs any `bowtie` command, you will go out and download a big HTTP report.

You can notice this by literally just running `bowtie` locally :), even with no arguments, but certainly with `bowtie filter-implementations` which is the test that's failing. You'll observe it's instant when run on `main`, and very slow when run on your branch. And if you interrupt it, you'll see it's downloading things.

The reason this fails only on the coverage CI and not others is quite funny, it's because that environment is *already* very slow (because collecting coverage is slow). So only on that environment does making it even *slower* trip the timeout (which pexpect sets to 30 seconds by default). But merging this as-is would have been quite bad! In general you should definitely at least spend a few minutes looking at the test error, trying out what it's saying, re-running the tests a time or two to see if it deterministically fails, and *then* trying to decide whether it's unrelated or not.

adwait-godbole:
Well I am definitely feeling embarrassed now after hearing this reason. 😅 Sorry for causing this trouble 🙏! Will definitely remember next time to contemplate before coming to a conclusion right away.

adwait-godbole:
Also sorry for wasting your time on figuring out a mistake that I made which I should've looked at way earlier on.

Julian:
No problem, it happens!

adwait-godbole:
@Julian, Looks ok now ?

Julian:
Nice! Yep think so!

PR #1218, opened 2024-05-29
Dont poison containers until needed otherwise

### LINKED ISSUE #1187: Validating using the container connectable should not send stop requests / cause the containers to quit
Right now, given a running container via e.g. `podman run --rm -id ghcr.io/bowtie-json-schema/python-jsonschema` with ID `85adf51173fe1eb697cfe7a35842b34bc204ba3ee8118e92bb7b6c31c095e660`, running e.g.:

```
bowtie validate -i container:85adf51173fe1eb697cfe7a35842b34bc204ba3ee8118e92bb7b6c31c095e660 <(printf '{"$schema": "https://json-schema.org/draft/2020-12/schema", "not": {}}') <(printf '12')
```

will, at the end of the run, send a `stop` command, causing the harness to exit.

But the "entire point" of this kind of connectable is that Bowtie isn't managing its lifecycle, so we shouldn't even *request* it to stop, we should simply stop sending it requests.

sudo-jarvis:
Closes #1187.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1218.org.readthedocs.build/en/1218/

### DIFF
--- a/bowtie/_core.py
+++ b/bowtie/_core.py
@@ -490,7 +490,6 @@ class Implementation:
         cls,
         id: ConnectableId,
         reporter: Reporter,
-        should_poison_harness: bool = False,
         **kwargs: Any,
     ) -> AsyncIterator[Self]:
         _harness = HarnessClient(**kwargs)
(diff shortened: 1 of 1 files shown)


Julian:
I'm not a fan of boolean flags -- let's just remove the call entirely, and if it's needed, the connectable can call the method externally.

PR #1221, opened 2024-05-30
`bowtie latest-report` command

adwait-godbole:
@Julian tests for this will rely on the http mocking part i suppose ?

----
📚 Documentation preview 📚: https://bowtie-json-schema--1221.org.readthedocs.build/en/1221/


bowtie/_cli.py
```diff
@@ -1357,6 +1359,32 @@ async def filter_dialects(
             break
 
 
+@subcommand
+@click.option(
```
Julian, line 1363: This is essentially redefining `@DIALECT`.

adwait-godbole: But the `@DIALECT` doesn’t have a default 2020-12 on it I suppose.

bowtie/_cli.py
```diff
@@ -1357,6 +1359,32 @@ async def filter_dialects(
...
+    ),
+)
+def latest_report(dialect: Dialect):
+    """
+    Output latest Bowtie report for a given dialect.
+
+    If no dialect provided, it outputs Bowtie's latest
+    dialect's latest report.
+    """
+    for line in asyncio.run(dialect.latest_report()).iter_lines():
```
Julian, line 1384: You don't really need lines here, you're just writing the contents unchanged -- possibly there's something even more straightforward that isn't occurring to me but at least the below makes fewer assumptions:

```python
def latest_report(...):
    async def write(response):
        async for chunk in (await response).aiter_bytes():
            click.echo(chunk)
    asyncio.run(write(dialect.latest_report()))

```

bowtie/_cli.py
```diff
@@ -816,9 +819,11 @@ def statistics(
     format: _F,
 ):
     """
-    Show summary statistics for a previous report.
+    Show summary statistics for a Bowtie generated report.
 
-    If no report provided, defaults to Bowtie's latest dialect's latest report.
+    If no report provided, it shows statistics for Bowtie's latest dialect's
+    latest report. For getting statistics on other dialect reports, do
+    `bowtie latest-report --dialect {YOUR_DESIRED_DIALECT} | bowtie statistics`
```
adwait-godbole, line 826: @Julian is this fine as a helper text 😅?

bowtie/_cli.py
```diff
@@ -1364,6 +1369,23 @@ async def filter_dialects(
...
 
 
+@subcommand
+@DIALECT
+def latest_report(dialect: Dialect):
+    """
+    Get latest Bowtie generated report for a dialect.
+
+    If no dialect provided, it outputs Bowtie's latest
+    dialect's latest generated report.
```
adwait-godbole, line 1379: Also this one.

Julian:
Merged with some minor changes, thanks again -- though I see the statistics tests we added are broken, specifically they only pass when run from a machine in the UTC timezone (like the GitHub actions runners), but fail locally. The timezone handling is off, we probably should just use dateutil or pendulum which are the usual solutions there, but I'll have to look tomorrow.

Issue #1003, opened 2024-03-10
Show one `unsuccessful` column in the summary table, with a tooltip explaining the 3 possible reasons a test might be unsuccessful


sudo-jarvis:
When hovering over the `unsuccesful` heading of the summary table on https://bowtie.report, the modal that appears seems to have the content overflowing to the left.

https://github.com/bowtie-json-schema/bowtie/assets/75855708/f1afc446-4b78-4f1e-85c2-2d3c9105ec38

sd1p:
Can I take this issue?

sd1p:
I made changes to the hover information so it wouldn't show "error" and "failed" messages together anymore. However, I'm not sure if the current message ("error" or "failed") is displayed correctly based on the actual situation. Can you help me verify that?

After changes.

Hover information for unsuccessful.
![Screenshot 2024-03-11 151250](https://github.com/bowtie-json-schema/bowtie/assets/43758514/67afa061-d5ea-481e-b47c-70a511042994)

Hover information for error.
![Screenshot 2024-03-11 151434](https://github.com/bowtie-json-schema/bowtie/assets/43758514/17037bd7-ae3f-4caa-96fd-56a6d28bce6a)

Julian:
The columns we have right now are a bit of a mess honestly. I've tried a couple of different iterations for what columns we should have but I would probably just have us move to one single "unsuccessful" column at this point in the summary, and then have a tooltip which explains the possible test results in further detail. I'm going to rename this ticket to generalize it to that but if someone wants to help there yes definitely welcome.

sd1p:
After thinking about it, I found that errors and skipped columns aren't related to unsuccessful column data. So we can have a tooltip for the whole row showing errors, skipped, and unsuccessful data, and another tooltip to breakdown unsuccessful column data into failed and errored counts.

sd1p:
I confused myself with error cases and errored tests cases, after going through the CLI code and `parseReportData` code. I found that errored tests, failed tests and skipped tests are corelated and can be put together in the unsuccessful column tooltip.

We can remove the skipped tests column and show the skipped tests data in the unsuccessful column tooltip.

PR #1225, opened 2024-06-01
Start using CSS Modules

adwait-godbole:
Hi @Julian while I was working on fixing #1003, I realized that the repo was using `.css` files while styling React components, which is not really the best approach in modern times. It leads/might lead to class name clashes. So we should start using [CSS Modules](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/) (as the official React docs say),  via `.module.css` which activates that functionality. We can leave out [`global.css`](https://github.com/adwait-godbole/bowtie/blob/main/frontend/src/global.css) as it will just be used to apply global stylings to the DOM elements.

(This is an optional but good feature to have, so even if you don't think we need it I am fine with it :) !)


Julian:
Makes sense I think, thanks!

PR #1227, opened 2024-06-02
Update pnpm 9.1.2 -> 9.1.4

adwait-godbole:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1227.org.readthedocs.build/en/1227/

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,7 +6,7 @@
   "engines": {
     "node": ">=21.0.0"
   },
-  "packageManager": "pnpm@9.1.2",
+  "packageManager": "pnpm@9.1.4",
   "scripts": {
     "start": "vite --no-clearScreen",
     "build": "tsc && vite build",


Julian:
Thanks.

It would be nice if one or both of:

* dependabot did this automatically (but it doesn't seem to be able to)
* we could use a slightly looser range (but unfortunately that's not yet supported according to nodejs/corepack#95)

So for now yeah, I've just been managing this manually :/...

adwait-godbole:
> dependabot did this automatically (but it doesn't seem to be able to)

I think (but not sure) dependabot doesn’t seem to be able to update automatically because it only supports pnpm v7, v8 ig ? that’s what the [official docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pnpm) are saying.

Julian:
That's updating dependencies, I'm not taking about that I'm talking literally about updating this `packageManager` field

adwait-godbole:
Ohh alright! Will have a look at it 👍

PR #1250, opened 2024-06-04
backward compatibility for running `ruby-json_schemer` versions upto `v0.2.17`

adwait-godbole:
@Julian this is the first step that I've taken towards enabling support for multiple versions of the "json_schemer" implementation within the same ruby harness/code file. Here are some things that are yet to be considered. I've taken it to be as backward compatible as possible from my side specifically upto the version `v0.2.17` of "json_schemer". Going back beyond this version fails the Dockerfile build process because of incompatible Ruby versions which is again something that would have to be taken care of (I haven't looked into how to solve this yet but your input into this would be appreciated.)

----
📚 Documentation preview 📚: https://bowtie-json-schema--1250.org.readthedocs.build/en/1250/


implementations/ruby-json_schemer/Dockerfile
```diff
@@ -2,6 +2,10 @@ FROM ruby:3-alpine as builder
 
 # unf_ext wants libc++ to compile itself
 RUN apk add --no-cache g++ make
+
+ARG JSON_SCHEMER_VERSION=2.3.0
+ENV JSON_SCHEMER_VERSION=$JSON_SCHEMER_VERSION
```
adwait-godbole, line 7: The above two lines help us in dynamically setting the `JSON_SCHEMER` version during the docker / podman build process in the below manner:-

` docker build --build-arg JSON_SCHEMER_VERSION=0.2.17 .`

If no argument is provided then it defaults to the latest one (in this case the latest one is `v2.3.0` as of now).

adwait-godbole: and here we set the environment variable with the `JSON_SCHEMER` version that was locked on.

adwait-godbole:
Also with this setup I was able to run the suites on json_schemer versions upto `0.2.17` and was definitely able to see striking differences in the `number of unsuccessful tests`.

implementations/ruby-json_schemer/Dockerfile
Julian: I think probably we should pick a cross-implementation convention here, so that someone can use the same thing across all versions. E.g. `IMPLEMENTATION_VERSION`?

implementations/ruby-json_schemer/Gemfile
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
-gem 'json_schemer', '2.3.0'
+gem 'json_schemer', ENV.fetch('JSON_SCHEMER_VERSION', '2.3.0')
```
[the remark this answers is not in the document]
Julian: I suspect it very well may ruin dependabot, but honestly I don't know how to check.

implementations/ruby-json_schemer/bowtie_json_schemer.rb
```diff
@@ -10,18 +10,34 @@ class UnsupportedCommand < StandardError; end
...
-  JSONSchemer::Draft202012::BASE_URI => JSONSchemer::Draft202012::SCHEMA,
-  JSONSchemer::Draft201909::BASE_URI => JSONSchemer::Draft201909::SCHEMA,
-  JSONSchemer::Draft7::BASE_URI => JSONSchemer::Draft7::SCHEMA,
-  JSONSchemer::Draft6::BASE_URI => JSONSchemer::Draft6::SCHEMA,
-  JSONSchemer::Draft4::BASE_URI => JSONSchemer::Draft4::SCHEMA,
-}
-@meta_schema_refs.merge!(JSONSchemer::Draft202012::Meta::SCHEMAS)
-@meta_schema_refs.merge!(JSONSchemer::Draft201909::Meta::SCHEMAS)
-@meta_schema_refs.transform_keys! { |uri| uri.dup.tap { _1.fragment = nil } }
+additional_dialects = if Gem::Version.new('2.0.0') <= JSON_SCHEMER_VERSION
```
Julian, line 33: I'd try to do all this version stuff in one spot above.

Here you should be able to add `SUPPORTED_DIALECTS = ... above, and in an `else` (for post-2.0) do `SUPPORTED_DIALECTS = ...newerones`

implementations/ruby-json_schemer/bowtie_json_schemer.rb
```diff
@@ -56,23 +70,36 @@ class UnsupportedDialect < StandardError; end
     }
   when 'dialect'
     dialect = request.fetch('dialect')
-    @meta_schema = JSONSchemer::META_SCHEMAS_BY_BASE_URI_STR[dialect]
-    raise UnsupportedDialect, dialect unless @meta_schema
+
+    if Gem::Version.new('2.0.0') <= JSON_SCHEMER_VERSION
```
Julian, line 74: Same here, I'd try to move this logic above to one area (so define some callable there depending on version which gives you what you need.

adwait-godbole:
> After doing all this one thing still remains a question as to how are we exactly going to do / determine docker build --build-arg JSON_SCHEMER_VERSION=<some_version> . What will be some_version here ? Will we have to use GitHub API to fetch the version tags ? Maybe the last five / ten tags ? or what ?

@Julian can you give your review on this as well.

Julian:
> @Julian can you give your review on this as well.

I'm not sure yet how much to automate that. Perhaps we simply have a static file with `matrix_versions.json` or something which just contains all the versions we can build? And then if we can generate that file by calling the GitHub API, great, otherwise we can always manually put stuff in there for each implementation. But obviously not 100% sure yet.

implementations/ruby-json_schemer/Gemfile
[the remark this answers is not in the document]
adwait-godbole: > I suspect it very well may ruin dependabot, but honestly I don't know how to check.

Hmm, idk if I am getting too ahead of myself here but then should we ask this query to the repo maintainer then ? 😅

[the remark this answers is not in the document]
adwait-godbole: > upto the version v0.2.17 of "json_schemer". Going back beyond this version fails the Dockerfile build process because of incompatible Ruby versions which is again something that would have to be taken care of

also this issue arises when trying to go back `v0.2.17`

adwait-godbole:
> I'm not sure yet how much to automate that. Perhaps we simply have a static file with matrix_versions.json or something which just contains all the versions we can build? And then if we can generate that file by calling the GitHub API, great, otherwise we can always manually put stuff in there for each implementation. But obviously not 100% sure yet.

Yup! Okay! Doing `matrix_versions.json` then seems like an easier first step for now. We can go ahead with it then. So should I create this file in this PR then ?

implementations/ruby-json_schemer/matrix_versions.json
```diff
@@ -0,0 +1,21 @@
...
+  "0.2.25",
+  "0.2.24",
+  "0.2.23",
+  "0.2.22",
+  "0.2.21",
+  "0.2.20",
+  "0.2.19",
+  "0.2.18",
+  "0.2.17"
+]
```
adwait-godbole, line 21: here I've only added upto `v0.2.17` as my refactored harness code is only working upto that version. Will definitely add further more afterwards 👍.

adwait-godbole:
@Julian should we get this merged then ?

Julian:
I'll have another look tomorrow morning but yep think so!

Julian:
Merged after tweaking your ruby a bit. Have a look at the commits.

But nice work, what I see in terms of results across 3 versions (which I haven't pushed remotely, just locally) is...

```
⊙  bowtie suite -i ruby-json_schemer -i image:ruby-json_schemer:0.2.25 -i image:ruby-json_schemer:0.2.17 7 | bowtie summary --show failures
Ran 252 test cases.
                      Bowtie                       
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━━━━┓
┃ Implementation      ┃ Skips ┃ Errors ┃ Failures ┃
┡━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━╇━━━━━━━━━━┩
│ json_schemer (ruby) │ 0     │ 0      │ 0        │
│ json_schemer (ruby) │ 0     │ 23     │ 2        │
│ json_schemer (ruby) │ 0     │ 24     │ 3        │
└─────────────────────┴───────┴────────┴──────────┘
```

(side note, clearly we'll have to think about what to do with showing version output in summaries if the same implementation appears multiple times...)

adwait-godbole:
> Merged after tweaking your ruby a bit. Have a look at the commits.

Thanks, yeah I took a look at those commits and will definitely keep the learnings in mind for future.

> (side note, clearly we'll have to think about what to do with showing version output in summaries if the same implementation appears multiple times...)

Yeah definitely! Should I open an issue regarding this ? (displaying version number of an implementation in ` bowtie summary`) ?

Julian:
> Should I open an issue regarding this ? (displaying version number of an implementation in  bowtie summary) ?

Please. Though I think we don't want to do it always, just if the same implementation appears multiple times with different versions...

PR #1266, opened 2024-06-07
JSON Schema for implementation matrix versions

### LINKED ISSUE #1264: Schematize the `implementations/*/matrix_versions.json` file
We should add a simple (private) schema for the matrix_versions.json file we're going to be adding to implementation harness directories.

`bowtie/schemas` (our existing registry of files) seems like a fine enough physical place for it to live.

CC @adwait-godbole

adwait-godbole:
Closes #1264 

@Julian This schema only checks if the version strings are of the format "X.Y.Z". The schema is invalid for "X.Y" or "X". So let me know if we would want to make it valid for them as well. However if we do so then it shouldn't happen that the array had "2.0.0", "2.0" and "2" present all at the same time because in this case even `"uniqueItems": true` won't work ig.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1266.org.readthedocs.build/en/1266/


bowtie/schemas/implementation-matrix-versions.json
```diff
@@ -0,0 +1,17 @@
...
+
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+
+  "$id": "tag:bowtie.report,2024:implementation-matrix-versions",
+
+  "type": "array",
+
+  "items": {
+    "type": "string",
+    "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$"
```
Julian, line 13: Use the field we already have for implementation versions, it's here:

https://github.com/bowtie-json-schema/bowtie/blob/main/bowtie/schemas/models/implementation.json#L27

We don't currently restrict things at all to a regex (it may not even be possible to, I'm not sure what regex covers every version format across all languages).

But yeah use a `$ref` to that (add an `$id` there probably to do so) and then we can independently decide whether to add a `pattern` or not.

adwait-godbole:
@Julian does it look ok now ? I am not really sure of why the `pre-commit.ci` run is failing. It is showing something like below...

```
check lintsonschema for drift..................................................Failed
- hook id: check-lintsonschema
- exit code: 1

lintsonschema's schemas drifted from the real ones. Fixing...
```

Julian:
Would be nice if we had some way of actually validating this but it seems check-jsonschema can't quite do it yet. Will have to file an issue there for it.

PR #1292, opened 2024-06-12
backward compatible code for `js-hyperjump` implementation

adwait-godbole:
Hi @Julian, I've tried to write some backward compatible code in here. However my way of doing it is not able to start the containers for hyperjump versions `>= v1.0.0` (as seen [here](https://github.com/bowtie-json-schema/bowtie/actions/runs/9477051990/job/26110983524?pr=1292) in this failed "Rebuild Bowtie Images" run as well), but the code is able to start the container for `v0.23.4` (which was the last released version before `1.0.0`). Can you help me figure out what am I doing wrong here that the containers aren't starting for versions `>= v1.0.0` 😅 ?


Julian:
Nice.

That's a javascript warning, it's complaining about you using `import()` I believe, instead of the import statement (and then Bowtie is not allowing messages to appear on stderr without failing the smoke test).

I'm not sure what the fix is, do conditional import (statements) not work in the spot you have it there?

I can have a look, but likely not until tomorrow.

Also not sure you have the context for what's happening here (essentially we're trying to add support for running tests across historical versions to build trend reports) -- but @jdesrosiers in case you have any spare time and care to review this, pinging you! If not no worries of course.

adwait-godbole:
> I'm not sure what the fix is, do conditional import (statements) not work in the spot you have it there?

They are working but only in the code condition when the hyperjump version was < 1.0.0 over [here](https://github.com/bowtie-json-schema/bowtie/blob/ba1db3def4ac77bcdfbcd61374f575c58af4577d/implementations/js-hyperjump/bowtie_hyperjump.js#L101). In other conditions (i.e. >= 1.7.0 and  >= 1.0.0) they don't seem to be working according to the code I've written. So I am kind of stuck here but @jdesrosiers if you can guide me if I am doing something wrong over here then that would be a lot helpful. I am guessing you already know this but still if you want to try this out you can simply build a docker/podman (whichever you prefer) image doing 

`docker build --build-arg IMPLEMENTATION_VERSION=0.23.4 -t js-hyperjump:v0.23.4 .`

and then you check out the test suite results by doing 

`bowtie suite -i js-hyperjump:v0.23.4 2020 | bowtie summary --show failures`

The above works according to this PR code but the same is not working for an image version of "1.0.0" or "1.6.7" (the container just fails to start). Also it would be helpful for me if you can confirm/recall if the test suite results are what you've seen in the past for those versions.

Julian:
> They are working but only in the code condition when the hyperjump version was < 1.0.0 over [here](https://github.com/bowtie-json-schema/bowtie/blob/ba1db3def4ac77bcdfbcd61374f575c58af4577d/implementations/js-hyperjump/bowtie_hyperjump.js#L101).

To be clear, what you have here (and I may not know the JS terminology) is not what I was calling an import statement. You're calling a function. Compare the diff.

As I say though I'm no expert, I just know how to cargo cult JS/TS code, but I do recall googling that exact error message when playing with the JS implementations.

Will try to provide more specifics when I get a chance to look at this if you don't get unblocked.

adwait-godbole:
@Julian I tried fixing that experimental error log by following [this discussion](https://github.com/nodejs/node/issues/51347) and using a slightly different way of import `package.json` for getting to know the `hyperjump` version  but still the container is failing to start :(.

jdesrosiers:
I'll look closer later. I don't know what your trying to accomplish, so I'll have to catch up on the context. Without any context, here are my thoughts.

It sounds like you're trying to use v1 in a CommonJS context, which node.js doesn't allow. v0 uses CommonJS (`require`) and v1 uses ESM (`import`). The test runner for v1 needs to have `"type": "module"` in package.json or have extension `.mjs` in order for node to run it as ESM.

jdesrosiers:
Oh, sorry. I see the actual problem has to do with the JSON import. I don't use that syntax so I'm not familiar with the error. If what you changed doesn't help, I usually just read the file and run JSON.parse on the file contents.

implementations/js-hyperjump/bowtie_hyperjump.js
```diff
@@ -24,42 +18,138 @@ function send(data) {
...
-  ...fileSchemeSkippedTests,
-};
+  if (hyperjump_version >= "1.7.0") {
+    await Promise.all([
+      import("@hyperjump/json-schema/draft-04"),
+      import("@hyperjump/json-schema/draft-06"),
+      import("@hyperjump/json-schema/draft-07"),
+      import("@hyperjump/json-schema/draft-2019-09"),
+    ]);
+    const module = await import("@hyperjump/json-schema/draft-2020-12");
```
jdesrosiers, line 46: These imports are needed for all v1, not just >= 1.7.0.

The only difference from 1.7.0 forward is that `addSchema` was replaced with `registerSchema`. Actually, `addSchema` still works, it's just deprecated.

jdesrosiers (commented):
Is there a reason you need to have one test harness that supports both v0 and v1? Even if you can make it work, it's going to be an ugly unmaintainable mess of flags. It would make everyone's lives a lot easier if you just have two versions of the test harness.

I don't think the skipped tests need to be different by version assuming you're always going to use the latest version of the test suite. I'm pretty sure the only changes to the skipped tests over time reflect changes the test suite rather than the implementation.

Julian:
Having separate harness files makes containerizing a bit more complicated. We're playing it by ear though so if it turns out not to be worth it we can of course split, for the other one we've done already it turned out easier this way which was the only reason to try this first.

PR #1293, opened 2024-06-12
Add JSON Toolkit to Bowtie

jviotti:
Still a very WIP, but would like to work on it in the open to make it easier to pair debug if needed (as I never added an implementation before!)

See: https://github.com/sourcemeta/jsontoolkit

----
📚 Documentation preview 📚: https://bowtie-json-schema--1293.org.readthedocs.build/en/1293/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -157,6 +157,11 @@ updates:
       interval: "daily"
 
   - package-ecosystem: "docker"
+    directory: "/implementations/cpp-jsontoolkit"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "docker"
     directory: "/implementations/dotnet-jsonschema-net"
     schedule:
       interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,6 +30,7 @@ updates:
   #     interval: "daily"
 
   # ignore: cpp-valijson - builds its repository's HEAD
+  # ignore: cpp-jsontoolkit - builds its repository's HEAD
 
   - package-ecosystem: "nuget"
     directory: "/implementations/dotnet-jsonschema-net"
--- a/implementations/cpp-jsontoolkit/Dockerfile
+++ b/implementations/cpp-jsontoolkit/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine AS builder
+FROM alpine:3.20 AS builder
 
 RUN apk add --no-cache cmake g++ git make 
 
@@ -18,6 +18,6 @@ RUN g++ -O3 -static -std=c++20 -o /tmp/bowtie /tmp/bowtie.cpp \
   -lsourcemeta_jsontoolkit_uri \
   -luriparser
 
-FROM alpine
+FROM alpine:3.20
 COPY --from=builder /tmp/bowtie /usr/local/bin
 CMD [ "bowtie" ]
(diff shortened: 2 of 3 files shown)


jviotti:
@Julian I think I did it well, but a bit unsure about the exception handling part. Should I return an `errored` object *inside* `results`?

jviotti:
Ah, no. It's a top-level thing. OK. I think I did it well now

implementations/cpp-jsontoolkit/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM alpine AS builder
```
Julian, line 1: We pin these base images so that the builds are reproducible, and then configure dependabot to update them (which is what pre-commit is failing to say).

So

```suggestion
FROM alpine:3.20 AS builder
```

(and below).

implementations/cpp-jsontoolkit/bowtie_jsontoolkit.cpp
```diff
@@ -0,0 +1,113 @@
...
+      assert(started);
+      assert(message.defines("dialect") && message.at("dialect").is_string());
+      default_dialect = message.at("dialect").to_string();
+      auto response{JSON::make_object()};
+      response.assign("ok", JSON{true});
+      stringify(response, std::cout);
+      std::cout << std::endl;
+    } else if (command == "run") {
+      assert(started);
+      assert(message.defines("seq") && message.at("seq").is_integer());
```
Julian, line 59: It's actually not an integer! `bowtie smoke` (which I'm in the middle of beefing up) will soon warn harness authors of that automatically, but yeah `seq` [has no public type](https://github.com/bowtie-json-schema/bowtie/blob/main/bowtie/schemas/report.json#L25-L28), it's an opaque JSON value you're supposed to pass along (and in fact for internal transition reasons it's not an integer when running `bowtie suite`).

implementations/cpp-jsontoolkit/Dockerfile
jviotti: Ah, I copied it from `valijson`, which seems to work without warnings. I'll pin it!

implementations/cpp-jsontoolkit/bowtie_jsontoolkit.cpp
jviotti: Ahhh, got it. Might be worth explaining just that on the "new implementation" guide. Let me fix it now...

.github/dependabot.yml
```diff
@@ -30,6 +30,7 @@ updates:
   #     interval: "daily"
 
   # ignore: cpp-valijson - builds its repository's HEAD
+  # ignore: cpp-jsontoolkit - builds its repository's HEAD
```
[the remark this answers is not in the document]
Julian: Yes, there's 2 things dependabot will update, and this tells it to ignore one.

* Your implementation version (which this disables)
* Your base image (which is what's failing)

jviotti:
Looks like we are still getting some check issues even after my changes:

Screenshot 2024-06-13 at 11 02 02 AM

Julian:
I still don't see you adding a dependabot section. Maybe you didn't push? Or else I'll leave a comment with what you need to add in the file.

Julian:
Never mind I forgot GitHub review is awful and doesn't let you leave comments on arbitrary lines. But you need a block like what's on line 153 of the dependabot.yml:

```
  - package-ecosystem: "docker"
    directory: "/implementations/cpp-jsontoolkit"
    schedule:
      interval: "daily"
```

jviotti:
Done @Julian ! Also, the actual job seems to have passed.

jviotti:
Yay! The check is green now :)

Julian:
Woohoo. Looks great, will merge, congrats/thanks!

Julian:
You're live! https://bowtie.report/#/implementations/cpp-jsontoolkit

PR #1298, opened 2024-06-14
Report JSON Toolkit short git SHA as the version

jviotti:


Julian:
If you don't anyways plan on having version numbers sometime soon, I think it'd be better to pass a commit sha as the version -- the point of the field is to let someone compare the behavior they see locally to what's on the site, but obviously main today might be different from what it is/was when they compiled, or when bowtie is behind, etc.

jviotti:
@Julian The project reports version numbers, but I really like the idea of Bowtie running it from the tip of `main`. I just updated the PR to do what you suggested: report the commit SHA. Please take a look!

Julian:
Awesome thanks!

PR #1301, opened 2024-06-15
Add concurrency group for ci workflow

### LINKED ISSUE #1202: Restrict the CI workflow to a concurrency group that is once per pull request
We currently run multiple jobs if someone repeatedly pushes to a pull request.

Even though it'd be nice to run on every commit at some point, given that we are nearly 100% integration tests which are slow, we should restrict to just running on the latest synchronization and cancel any outstanding jobs using GitHub actions concurrency groups.

We already do this for the image rebuild workflow.

OptimumCode:
Unfortunately, creating a concurrency group that will work only for PR is impossible. It will also affect the pushes to branches and scheduled workflows. If only PRs should be affected, probably, the best option would be to create a reusable workflow from CI workflow and create two separate workflows that use it: one for PRs (with concurrency group) and one for pushes to branches/tags and scheduled runs.

Using `${{ github.event_name }}` should prevent collisions between scheduled executions and pushes to the main branch.

Resolves #1202

----
📚 Documentation preview 📚: https://bowtie-json-schema--1301.org.readthedocs.build/en/1301/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,10 @@ on:
     - cron: "12 5 * * *"
   workflow_dispatch:
 
+concurrency:
+  group: ci-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   list:
     runs-on: ubuntu-latest


Julian:
Thanks again for tackling this kind of ticket, it's very much appreciated, definitely helps distribute the load!

Just a question --

> Unfortunately, creating a concurrency group that will work only for PR is impossible.

I of course haven't looked into this again myself since filing the issue, but isn't this possible by using a concurrency group name containing something which:

* is unique to each PR
* but set to something unique to each commit for non-PRs?

(n.b. I'm honestly not 100% sure what behavior we really care about for normal pushes, I guess it's not the worst thing if we have to live with cancelling for all. But just thinking aloud:)

Specifically if we pick a property present on `pull_request` event types, and then fallback when that's not present to the commit SHA? E.g. in pseudo-YAMLvomit, `group: ci-${{ github.pull_request.number || github.push.after }}-${{ github.ref }}` or something like that?

OptimumCode:
Thank you @Julian for taking a look at this. I always forget that those expressions are kinda JS and you can do things like you suggested `${{ github.pull_request.number || github.event.after }}`)
I think that should work - the group will be bound to PR and for non-PR every push will have a dedicated group.

But I am not sure if you should have a unique group for each push to the branch. In my opinion, it does not make much sense to have several workflow runs for commits pushed one by another to a branch. We are not interested in results for intermediate commits. Only the workflow run for the last one is interesting. And you use tag pushes to create a release - they will always have a unique `github.ref` (unless you republish the tag of course).

If you want to have several workflow runs per branch I will change the group to something similar to your suggestion. If you don't, I think, we can keep the original version.

Julian:
Cool! All good, I'll have a look in the morning, as-is is probably great!

PR #1304, opened 2024-06-17
Add groups to dependabot config

### LINKED ISSUE #1203: Make use of dependabot groups
Dependabot (as of a little while ago) supports [grouping updates within the same PR](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups).

We don't make use of them, mostly because I couldn't figure out the syntax, the docs examples don't seem very helpful/realistic.

At very least, we should batch updates to:

* the same implementation harness all together
* Bowtie's own Python requirements all together

OptimumCode:
Resolves #1203

----
📚 Documentation preview 📚: https://bowtie-json-schema--1304.org.readthedocs.build/en/1304/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,11 +9,19 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      bowtie-dependencies:
+        patterns:
+          - "*"
 
   - package-ecosystem: "npm"
     directory: "frontend"
     schedule:
       interval: "weekly"
+    groups:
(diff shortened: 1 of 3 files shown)


OptimumCode:
I thought it would be more lines per implementation. Probably, we can indeed group all dependencies for each implementation in a single PR

Julian:
> Probably, we can indeed group all dependencies for each implementation in a single PR


Ok will merge this at least and then we can see about anything further. 

Thanks again!

PR #1306, opened 2024-06-17
Remove custom python code for listing implementations and instead use `bowtie filter-implementations` 

adwait-godbole:
@Julian just this minor fix I noticed while skimming through this file to get an idea of how to write a similar one for building multiple images.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1306.org.readthedocs.build/en/1306/

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -35,7 +35,7 @@ jobs:
 
       - id: images-matrix
         run: |
-          echo "images=$(bowtie filter-implementations | jq -R -s -c 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
+          echo "images=$(bowtie filter-implementations --format json" >> $GITHUB_OUTPUT
 
   build:
     needs: list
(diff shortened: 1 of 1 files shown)


.github/workflows/images.yml
```diff
@@ -26,18 +26,16 @@ jobs:
...
-          from pathlib import Path
-          import json
-          paths = [
-              str(path.name)
-              for path in Path("implementations").iterdir()
-              if path.is_dir()
-          ]
-          print(f"images={json.dumps(paths)}")
-          ' >> $GITHUB_OUTPUT
+          echo "images=$(bowtie filter-implementations | jq -R -s -c 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
```
Julian, line 38: I added `bowtie filter-implementations --format json`, so there's no need for ad hoc parsing here.

We don't apparently do the same "magic" when stdout isn't a tty (meaning making format JSON be the default when piping), though we could always change that later.

But yeah don't do text parsing.

PR #1311, opened 2024-06-18
Build and Publish Versioned Images workflow

adwait-godbole:
@Julian here's the first pass on this and you can have a look at the workflow run over [here](https://github.com/adwait-godbole/bowtie/actions/runs/9563828124) on my fork (I dispatched it manually). In this workflow I am only building and publishing versioned images for implementations which have a `matrix-versions.json` file in their directory for now. For those who don't, their latest images will be built and published by the `Rebuild Bowtie Images` workflow anyways.


adwait-godbole:
Although the tags are looking a bit weird over here https://github.com/adwait-godbole/bowtie/pkgs/container/ruby-json_schemer/versions. Am I doing the tagging thing incorrectly?

Julian:
I think the tags look OK from the page you linked (let me know what specifically looks weird to you though).

.github/workflows/versioned-images.yml
```diff
@@ -0,0 +1,87 @@
+name: Build and Publish Versioned Images
+
+on:
+  workflow_dispatch:
```
Julian, line 4: Can we perhaps at least start with taking input the implementation name maybe? So that we're not building all versions for *all* implementations, we just select from the list of implementations containing version files as you have it? Or does that complicate things a lot?

.github/workflows/versioned-images.yml
```diff
@@ -0,0 +1,87 @@
...
+    permissions:
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJson(needs.list.outputs.image-versions) }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install qemu
```
Julian, line 44: I guess it would be nice to call the other workflow with the version passed in rather than copy pasting it. But we can put that off if need be.

Julian:
This looks OK I think. Left a comment or two but we can also choose not to do it now and wait till we use this a bit more. Well done as usual!

adwait-godbole:
@Julian did the changes you requested for. Now the versioned-images.yml workflow is expecting the implementation name from the user and it calls images.yml workflow to perform build and publish tasks. You can have a look at the run on my fork over [here](https://github.com/adwait-godbole/bowtie/actions/runs/9586082360)

.github/workflows/images.yml
```diff
@@ -15,12 +23,9 @@ on:
     tags:
       - "v*"
 
-concurrency:
-  group: images-${{ github.ref }}
-  cancel-in-progress: true
-
```
adwait-godbole, line 21: I had to remove this because while building for multiple versions of an implementation, whenever images.yml was being called from versioned-images.yml, the concurrent jobs were getting cancelled and only one of the version was getting built. Let me know if there's a way I could fix this. For now removing this did not cancel the concurrent jobs for building multiple versions.

adwait-godbole:
Also @Julian rather than having a separate `versioned-images.yml` workflow, should I just put the logic in the existing `images.yml` workflow or is it better to keep them separate ?

adwait-godbole:
@Julian sorry to ping again over here but now I've modified the `images.yml` workflow itself to take an implementation name from the user whose multiple versions he wants to build for. Here's the workflow run on my fork triggered manually for `ruby-json_schemer` --> https://github.com/adwait-godbole/bowtie/actions/runs/9588354256

Julian:
There's a minor merge conflict I caused you again but otherwise I think this looks good enough to try out.

It's unfortunate that GH Actions doesn't support dynamically generating choices for inputs -- that seems like it's https://github.com/orgs/community/discussions/12029. The logic you have for detecting typos is a bit ugly, but it seems like the best we can do I guess, other than hardcoding the list of choices.

adwait-godbole:
@Julian resolved the merge conflicts. Looks ok to merge now ?

PR #1325, opened 2024-06-22
Container runtime pulls version tagged image if available or the latest one as default.

adwait-godbole:
@Julian is this fine ? It works now. Though I still have to write "image:" prefix for it to work.


Julian:
It'd be nice if that function took a `Connectable` instance instead of a string I guess, but yeah this seems fine for now, thanks.

PR #1331, opened 2024-06-25
A public directory structure for serving `implementations.json` to `json-schema-org`

### LINKED ISSUE #1312: Publish a JSON File with all the implementations details suitable for consumption by json-schema.org
Could it be possible to have a JSON file containing the details of all the implementations?

- Name
- Source
- Supported Dialects

We are building the[ new Tooling page for the JSON Schema website](https://github.com/json-schema-org/website/issues/160) and we plan to add a depth integration with Bowtie. The idea is to shaw the shields of the implementations if they are in Bowtie. 

**Our problem is we don't know the bowtie name of each implementation** and without it we can generate the shield urls. As of now the only shared value in both data sources is the repo url (source) so if you guys provide this new file we'll manage to build all the shield urls using the Bowtie name of each tool.

Thanks a lot!

adwait-godbole:
resolves #1312 

@Julian I am not sure if this was the best way of doing this but this is where my initial thoughts inclined towards. Here's how this file would look like once deployed.

https://adwait-godbole.github.io/bowtie/v1/json-schema-org/implementations.json

----
📚 Documentation preview 📚: https://bowtie-json-schema--1331.org.readthedocs.build/en/1331/

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -86,10 +86,25 @@ jobs:
         run: |
           bowtie info $(bowtie filter-implementations | sed 's/^/-i /') --format json > implementations.json
 
+      - name: Generate public-implementations.json file
+        run: |
+          jq 'to_entries | map({
+              (.value.source): (
+                .value + {name: .key}
+                | del(.os, .os_version, .language_version, .source)
+              )
+            })
+            | add' implementations.json > public-implementations.json
+
       - uses: actions/upload-artifact@v4
         with:
           name: implementations
           path: implementations.json
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: public-implementations
+          path: public-implementations.json
 
   site:
     needs:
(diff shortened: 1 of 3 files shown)


.github/workflows/report.yml
```diff
@@ -86,11 +86,26 @@ jobs:
...
 
+      - name: Generate public-implementations.json file
+        run: |
+          jq 'to_entries | map({
+              (.value.source): (
+                .value + {name: .key}
+                | del(.os, .os_version, .language_version, .source)
+              )
+            })
+            | add' implementations.json > public-implementations.json
```
Julian, line 97: This will break if/when we change the format of our own implementations.json, which is the sort of thing we're trying to avoid, so it would be slightly better to generate this from scratch via other public API commands. E.g. you could probably use `bowtie filter-implementations` + `bowtie info` I think?

Julian:
This looks more or less reasonable I think, thanks. Left a comment. Two others are:

* Let's leave off the `.json` I *think* for public API endpoints, I think it's fine to just have it be a bare URL with no extension, which is a bit more common
* Try to write a JSON Schema for that file, and test it's valid under it in the workflow

DarhkVoyd:
Hey, apologies for being late to the discussion,  I am working towards building the new tooling page. Thanks for your work on this issue.  I am somewhat unfamiliar with usage of bowtie, how are we planning the integration? My current idea is that the data model for the tooling page has a slot to store bowtie identifier, so we could possible somehow (maybe linking using repo link) populate that field and use some sort of api which we would fetch during build process to generate the badges. If I am mistaken, I am open to any implementation that is easier for you to implement and maintain. We could also hop on a quick call over slack if you prefer. Let me know!

Julian:
Hey!

The idea from this issue/PR was basically that you / json-schema.org would pull this file from https://bowtie.report/v1/json-schema-org/implementations (as soon as we merge) and then you could use it for the tooling page -- I'll have another look at this since it's easier to describe once live I guess, but you'd get a JSON object where the key was the source URL (which is also present on the tooling page) so you can look up any implementation that way essentially, and then the value would give you all the data you might want from Bowtie.

DarhkVoyd:
Great! Thank you for this. Let me know if anything.

PR #1348, opened 2024-07-06
Start using Implementation class

adwait-godbole:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1348.org.readthedocs.build/en/1348/


frontend/src/index.tsx
```diff
@@ -10,22 +10,20 @@ import { BowtieVersionContextProvider } from "./context/BowtieVersionContext";
...
+import Implementation from "./data/Implementation";
 
 const fetchReportData = async (dialect: Dialect) => {
   document.title = `Bowtie - ${dialect.prettyName}`;
   return dialect.fetchReport();
 };
 
-const fetchAllReportsData = async (langImplementation: string) => {
-  document.title = `Bowtie - ${langImplementation}`;
+const fetchImplementationReportViewData = async (implementationId: string) => {
```
Julian, line 25: One or both of these functions seem like they could be methods on `Implementation` no? Is that easy to do?

PR #1370, opened 2024-07-16
Add `bowtie perf` command

sudo-jarvis:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1370.org.readthedocs.build/en/1370/


bowtie/_cli.py
```diff
@@ -1198,6 +1203,14 @@ def convert(
 @SET_SCHEMA
 @TIMEOUT
 @VALIDATE
+@click.option(
```
Julian, line 1206: If this isn't meant to be exposed, better to have it be an environment variable we check probably, which is a bit more common.

bowtie/_cli.py
```diff
@@ -1271,6 +1298,179 @@ def validate(
...
+@dialect_option()
+@format_option()
+@click.option(
+    "--runs",
+    "-r",
+    "runs",
+    type=int,
+    default=3,
+    show_default=True,
+    help=("Number of runs used to run benchmarks."),
```
Julian, line 1312: Parens aren't needed in all these, they're just needed if you want multiple lines.

bowtie/_cli.py
```diff
@@ -1713,6 +1918,11 @@ async def _run(
             STDERR.print("[bold red]No test cases ran.[/]")
         elif count > 1:  # XXX: Ugh, this should be removed when Reporter dies
             STDERR.print(f"Ran [green]{count}[/] test cases.")
+
+        if time_measurement_path:
+            with time_measurement_path.open("a") as file:
```
Julian, line 1923: Use `Path.write_text`.

Julian: Oh. You're appending here?

bowtie/_core.py
```diff
@@ -859,3 +860,31 @@ def without_expected_results(self) -> Message:
 def registry():
     resources = referencing_loaders.from_traversable(files("bowtie.schemas"))
     return EMPTY_REGISTRY.with_resources(resources).crawl()
+
+
+def convert_table_to_markdown(
```
Julian, line 865: Can you have another look to make sure this function really doesn't exist in some markdown library? It seems a bit absurd for it not to.

bowtie/benchmarks/contains.py
```diff
@@ -0,0 +1,21 @@
+def get_benchmark():
```
Julian, line 1: Might want to split off a separate PR to add this `bowtie/benchmarks` directory with all the benchmarks, that might be easier to immediately add? Up to you though.

bowtie/benchmarks/draft2020_metaschema.json
```diff
@@ -0,0 +1,163 @@
+{
```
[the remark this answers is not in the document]
sudo-jarvis: Yeah actually I felt if we have static files then atleast we would know that we just need to re run the benchmarks during deployment only when there is some change in the benchmarks folder

tests/test_benchmarks.py
```diff
@@ -0,0 +1,433 @@
...
+    async def test_filtering_by_name(self):
+        stdout, stderr = await bowtie(
+            "filter-benchmarks",
+            "-t",
+            "keyword",
+            "-n",
+            uuid.uuid4().hex,
+            exit_code=0,
+        )
+        assert stdout == "", stderr == ""
```
Julian, line 433: This assert probably doesn't do what you expect, the syntax is `assert foo, message` where the latter is a message shown on failure (so here you'd show a message of `True` or `False` but not assert against it). Are you trying to assert 2 things here presumably?

bowtie/_cli.py
sudo-jarvis: Yeah because the same benchmark would run the same test multiple times to avg out, we would need all that value linked to one benchmark only. So one file per benchmark test.

tests/test_benchmarks.py
sudo-jarvis: Oh my bad, I thought it did the same

bowtie/_cli.py
sudo-jarvis: I am not sure how that would happen, then we would have to somehow set the enivornment variable from inside bowtie itself before running the pyperf benchmark. But idts the env var would persist when pyperf then runs its own process and tries running `bowtie run`. So when bowtie run would then check whether the env var exists or not, it would always say it doesnt exist, no ? @Julian

tests/test_benchmarks.py
```diff
@@ -0,0 +1,433 @@
...
+bowtie_dir = Path(__file__).parent.parent.joinpath("bowtie")
+default_benchmarks_dir = bowtie_dir.joinpath("benchmarks")
+keyword_benchmarks_dir = bowtie_dir.joinpath("benchmarks").joinpath("keywords")
+
+DIRECT_CONNECTABLE = "python-jsonschema"
+
+
+@pytest.fixture()
+def valid_single_benchmark():
+    module_name = "tests.benchmarks"
```
Julian, line 42: There's nothing dynamic here, why avoid the `import` statement?

tests/test_benchmarks.py
```diff
@@ -0,0 +1,433 @@
...
+            module="bowtie.benchmarks",
+        )
+        for benchmark_group in benchmark_groups:
+            assert benchmark_validated(benchmark_group.serializable())
+
+
+class TestBenchmarker:
+
+    def test_default_benchmarker(self, benchmarker_run_args):
+        _benchmarks.Benchmarker.from_default_benchmarks(**benchmarker_run_args)
```
Julian, line 219: Is this meant to test it doesn't blow up with those arguments, or what?

bowtie/_core.py
Julian: OK, fair enough then, thanks!

bowtie/benchmarks/draft2020_metaschema.json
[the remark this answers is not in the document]
Julian: Can you elaborate on what you're thinking there -- when/why would we run benchmarks during a deployment?
The things we'd run during a deployment are generally meant to ensure the deployment works correctly -- the only real "exception" is triggering a rebuild of the reports -- but that we do because it's 1) reasonably fast and 2) tests that the images we just rebuilt function.

tests/test_benchmarks.py
sudo-jarvis: Yeah just testing if the benchmarker loads up fine. So that in case something blows up we know where its blowing up.

tests/test_benchmarks.py
sudo-jarvis: Ended up following the same method here too. You are right, it isnt needed here.

bowtie/benchmarks/draft2020_metaschema.json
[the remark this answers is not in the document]
sudo-jarvis: My bad, I was thinking that we would need to recreate the benchmark artifacts each time we deploy. But ig we can simply run the benchmarks say weekly and during deployment it would just pick the latest artifact from the last successful run right ?

bowtie/benchmarks/keywords/draft2020-12/additionalProperties.py
```diff
@@ -0,0 +1,77 @@
...
+                        description="Valid",
+                        instance=_format_properties_as_instance(valid),
+                    ),
+                ],
+            )
+        )
+
+        array_size *= 10
+
+    return dict(
```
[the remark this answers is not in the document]
sudo-jarvis: Hmm that can be done, I already have such a type  - `BenchmarkGroup`. We just create it after loading the data instead of directly doing that in the benchmark file itself. Any particular advantage that it would have over the current thing ?

bowtie/_cli.py
sudo-jarvis: I tried doing this, the env variable didnt persist. Any other idea for the same?

bowtie/benchmarks/draft2020_metaschema.json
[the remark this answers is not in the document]
Julian: Right yeah that was what I was thinking!

bowtie/_cli.py
sudo-jarvis: Yeahh @Julian , ig when pyperf creates a process of its own, it doesnt inherit the env variables as also mentioned in : https://github.com/psf/pyperf/issues/47

sudo-jarvis: Oh wait, they did introduce some flag later on to enable inherting them!

sudo-jarvis: Works now @Julian.

bowtie/_core.py
```diff
@@ -714,12 +714,13 @@ def syntax(self) -> RenderableType:
     @classmethod
     def from_dict(
         cls,
+        instance: Any = None,
```
Julian, line 717: I'm not a fan of this -- I assume you mean to make this optional (by `None`)? But that's extra icky because `None` is a perfectly valid instance. I'll try to see where you're relying on this but it would seem to suggest another type would be better for whatever you're using this for.

tests/benchmarks/invalid_benchmark.py
```diff
@@ -0,0 +1,2 @@
+def get_benchmark():
```
Julian, line 1: The tests are inside the package now as of a week or two ago, so I'm not sure these are even running? Can you move them there and/or double check.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,439 @@
...
+bowtie_dir = Path(__file__).parent.parent.joinpath("bowtie")
+default_benchmarks_dir = bowtie_dir.joinpath("benchmarks")
+keyword_benchmarks_dir = bowtie_dir.joinpath("benchmarks").joinpath("keywords")
+
+DIRECT_CONNECTABLE = "python-jsonschema"
+
+
+@pytest.fixture()
+def valid_single_benchmark():
+    return dict(
```
Julian, line 41: Is this right? Isn't the expected type one of your `BenchmarkGroup` things now?

bowtie/_cli.py
```diff
@@ -1271,6 +1277,179 @@ def validate(
...
+    quiet: bool,
+    test_suite: tuple[Iterable[TestCase], Dialect, dict[str, Any]],
+    benchmark_files: Iterable[str],
+    **kwargs: Any,
+):
+    """
+    Perform performance measurements across supported implementations.
+    """
+    try:
+        if benchmark_files:
```
Julian, line 1375: Generally I think it's nicer in general to avoid `if/elif` chains -- specifically if you want a small exercise, here I think the pattern would be similar to what we do elsewhere, namely when you define each of these options, they shouldn't set boolean flags, they should set a `deserialize` argument (so `keywords` sets it to `Benchmarker.for_keywords` or some wrapper callable for that, `test_suite` to `from_test_cases`, etc.)

bowtie/_cli.py
```diff
@@ -1271,6 +1277,179 @@ def validate(
...
+
+    return 0
+
+
+@subcommand
+@dialect_option()
+@click.option(
+    "-t",
+    "--benchmark-type",
+    type=click.Choice(["default", "keyword"]),
```
Julian, line 1422: Doesn't have to be immediately (i.e. we can leave it and "fix"/"change") but I think `tags` or `markers` makes a bit more sense (i.e. that we have benchmarks have some notion of tags [a la website] or markers [a la pytest test markers]) and then you can filter by a tag/marker, and then right now we just have 1 of those, for keyword benchmarks.

bowtie/schemas/benchmarks.json
```diff
@@ -0,0 +1,66 @@
+{
+  "title": "Bowtie Benchmarks",
+  "description": "A collection of benchmarks used to compare the performance between various JSON Schema Implementations.",
```
Julian, line 3:
```suggestion
  "description": "A collection of benchmarks used to compare the performance between various JSON Schema implementations.",
```

docs/spelling-wordlist.txt
```diff
@@ -18,6 +18,7 @@ implementers
 linter
 linters
 positionally
+perf
```
Julian, line 21: I mean `perf` isn't really a word -- do you see where it's creeping in in the docs? Possibly somewhere should enclose it in ` (which would make this go away)?

tests/test_benchmarks.py
sudo-jarvis: Didnt do the same refactoring for tests properly. They are still loading from a dict. Fixing it now.

tests/benchmarks/invalid_benchmark.py
sudo-jarvis: I am not sure I get it. Could you please elaborate it further ?

Julian: These files belong in `bowtie/tests/benchmarks`, right now they probably don't even run in CI at all.

sudo-jarvis: `bowtie/tests/test_benchmarks.py` is loading the file, so it is being used right? Or else the tests would also fail.

Julian: I'm looking at the GitHub UI not the git branch, so maybe you're saying the UI is wrong, but `bowtie/tests/test_benchmarks.py` doesn't exist, you have `tests/test_benchmarks.py`, which I again suspect isn't running at all, now that the tests are at `bowtie/tests`!

docs/spelling-wordlist.txt
sudo-jarvis: I havent myself included perf anywhere in the docs yet, ig when its building the docs and getting the supported commands it would be adding perf right ? That is what it raising the spelling error to handle which I added the keyword in our wordlist.

bowtie/_core.py
sudo-jarvis: I am not exactly able to recall why I added this :( . It was probably the None instance failing bowtie run or something in some of the benchmark. It was raising the error instance cannot be left empty.

docs/spelling-wordlist.txt
Julian: Oh, that's odd, I guess that's a minor bug in `sphinx-click`'s output that hasn't been obvious given all the other command names are words... I mean this is pretty tiny obviously but perhaps put this instead in `docs/cli.rst` just above the click directive:

```
.. spelling:word-list::
   perf
```

so that at least it only applies in that file?

tests/benchmarks/invalid_benchmark.py
sudo-jarvis: Oh my bad :( I didnt merge the main branch and didnt realize the location of tests has changed. Will change that

bowtie/_cli.py
sudo-jarvis: Hi @Julian , tried doing this. Can you please have a look ?

bowtie/_cli.py
sudo-jarvis: I'll open an issue for this, and if time permits then would add this functionality.

bowtie/_cli.py
sudo-jarvis: The arguements for each are different but, any idea how to handle that?

sudo-jarvis: Could you please have a look now? @Julian

PR #1379, opened 2024-07-22
Add benchmarks

sudo-jarvis:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1379.org.readthedocs.build/en/1379/


sudo-jarvis:
@Julian , added the benchmark files here. Would remove these from the other PR. The tests veryfing these benchmarks and their format would be present in the other PR though

pyproject.toml
```diff
@@ -116,10 +116,16 @@ exclude = [
     "**/tests/conftest.py",
     "**/tests/fauxmplementations/miniatures.py",
     "**/tests/test_*.py",
+    "**/bowtie/benchmarks/*",
+    "**/tests/benchmarks/*",
 ]
 
 [tool.ruff]
 line-length = 79
+exclude = [
```
Julian, line 125: Any reason not to let ruff format these? Perhaps just add specific lint rule exclusions like we do below for tests?

sudo-jarvis: I was thinking that these are anyways static files so wont be of much benefit as well, some errors were cropping up. Ig I'll exclude the non useful ones.

PR #1382, opened 2024-07-24
Only build images for specific version of an implementation

adwait-godbole:
Hi @Julian! Let me know about this and if all good then after merging can you please rebuild for js-hyperjump's latest i.e. version 1.9.4 that we just merged yesterday. Thanks!

----
📚 Documentation preview 📚: https://bowtie-json-schema--1382.org.readthedocs.build/en/1382/


adwait-godbole:
The CI (typing) failure is unrelated to this PR :(

.github/workflows/images.yml
```diff
@@ -4,15 +4,19 @@ on:
   workflow_dispatch:
     inputs:
       implementation:
+        description: The name of an implementation.
+        required: false
+        type: string
+      version:
         description: >
-          The name of an implementation.
+          A specific version of the implementation you want to build.
```
Julian, line 12: Is `implementation` required when providing this `version` option (i.e. it's only optional if this isn't provided)?

Julian:
I've pinned away from the broken pyright version, please merge/rebase, but after that and the question think this is good!

.github/workflows/images.yml
adwait-godbole: I've made it so that either the user provides just the implementation / implementation + version / nothing at all. Each of these scenarios are handled in the if then logic.

adwait-godbole: If the user just provides the version without specifying any implementation then should I throw some error and exit ?

Julian: Probably yes, though I would slightly doubt GH Actions can represent that constraint so yeah you'll have to simply do it in the shell code.

adwait-godbole: Done @Julian.

adwait-godbole:
As mentioned above but yet writing again over here, after merging can you please trigger for js-hyperjump's 1.9.4. Thanks!

Julian:
https://github.com/bowtie-json-schema/bowtie/actions/runs/10081269536

Julian:
https://github.com/bowtie-json-schema/bowtie/actions/runs/10081264256 looks like unfortunate behavior -- specifically, triggering the version build I guess cancelled the normal/other build -- I think that means really we should perhaps include either the event (`workflow_dispatch`) or else the input parameters in the concurrency key. I.e. make it so that GH Actions doesn't consider the manually triggered build (which will build a different output) to be the same as the normal one.

PR #1383, opened 2024-07-24
Include the `matrix-versions.json` file for that implementation along with its versioned reports (better to merge after #1382)

adwait-godbole:
Hi @Julian, sorry to ping yet again here as well but I should've done this earlier. 

This helps both the frontend to know what versions directory of an implementation to look for and also helps `bowtie trend` CLI.

After merging, you will you have to manually trigger the workflow yet again 😅. Sorry for this trouble but before that deploying the latest image of js-hyperjump i.e. 1.9.4 after merging #1382 would also be necessary otherwise building versioned-reports for it would fail in this workflow since we don't have the image for 1.9.4 deployed already.

Thanks again :)!

----
📚 Documentation preview 📚: https://bowtie-json-schema--1383.org.readthedocs.build/en/1383/


adwait-godbole:
We can merge this as well now! Just that I will again need you to manually trigger this one so that the `matrix-versions.json` file reflects in the site directory of Bowtie website.

Thanks again!

PR #1384, opened 2024-07-25
Add `bowtie trend` command

adwait-godbole:
@Julian finally, a first pass on the `bowtie trend` command. Still working upon this but here's a quick demo.

https://github.com/user-attachments/assets/fb250d61-0cbc-4dfe-a70d-33a8392d4b16

----
📚 Documentation preview 📚: https://bowtie-json-schema--1384.org.readthedocs.build/en/1384/


bowtie/tests/fauxmplementations/lintsonschema/schemas/models/dialect.json
```diff
@@ -11,6 +11,8 @@
       "description": "The URI representing this dialect, also known as its dialect identifier",
       "readOnly": true,
 
+      "$id": "tag:bowtie.report,2024:models:dialect:uri",
```
Julian, line 14: Can you split this change of adding an `$id` for dialect URIs off and PR it immediately?

bowtie/_report.py
```diff
@@ -52,6 +54,32 @@ class MissingFooter(InvalidReport):
     """
 
 
+class CaseMismatchInSameSuiteReports(Exception):
```
Julian, line 57: I'll have to read where you use these, but I find the docstrings a bit hard to understand (i.e. to understand what you mean to say will cause these).

bowtie/_report.py
```diff
@@ -52,6 +54,32 @@ class MissingFooter(InvalidReport):
...
+    """
+    Two reports for the same dialect suite had some Case mismatch in them.
+    """
+
+
+class SeqResultMismatchInSameSuiteReports(Exception):
+    """
+    Two reports for the same dialect suite had some SeqResult mismatch.
```
Julian, line 65: More specifically, what does this mean? Isn't it expected that two reports might have different test results?

bowtie/_report.py
```diff
@@ -52,6 +54,32 @@ class MissingFooter(InvalidReport):
...
+
+    **Condition to that it was the same ConnectableId (i.e. same implementation
+    and its same version) in both the reports, yet there was some mismatch in
+    its SeqResults.**
+    """
+
+
+class ImplementationInfoMismatchInSameSuiteReports(Exception):
+    """
+    Two reports for the same dialect suite had an ImplementationInfo mismatch.
```
Julian, line 75: And here -- what makes this an error? Maybe an implementation added a link or something, does that really need to be an error?

In general I think just seeing these makes me think maybe you can have a separate PR (first?) which adds whatever additional validation motivated you to add these?

bowtie/_report.py
```diff
@@ -1,13 +1,15 @@
...
 from datetime import UTC, datetime
+from functools import reduce
 from typing import TYPE_CHECKING, TypedDict
 import importlib.metadata
 import json
 import sys
 
 from attrs import asdict, field, frozen
 from attrs.filters import exclude
+from packaging import version
```
Julian, line 12: What's this used for? It's suspicious to use it -- this implements Python packaging versioning -- that doesn't apply across all languages necessarily.

bowtie/_core.py
```diff
@@ -625,6 +643,144 @@ async def start(
...
+                        advance=content_length,
+                    )
+                    await asyncio.sleep(2)
+
+                    return frozenset(json.loads(content))
+            except httpx.HTTPStatusError:
+                return None
+
+    @classmethod
+    async def download_reports_for(
```
Julian, line 693: This seems like a weird thing to be a classmethod -- it doesn't construct an instance of the class, and doesn't really seem to use `Implementation` (`cls`) at all.

bowtie/_core.py
```diff
@@ -625,6 +643,144 @@ async def start(
 
         yield cls(harness=harness, id=id, info=info, reporter=reporter)
 
+    @classmethod
+    async def download_versions_of(cls, id: ConnectableId) -> Set[str] | None:
```
Julian, line 647: Try to avoid importing and using rich in this file -- rich is about presentation, and specifically presentation to the TUI.

Elsewhere we've tried (although perhaps not fully succeeded) at returning data structures, and then converting to rich only when needed.

So here it seems like maybe this should simply return a deserialized thing from retrieving the HTTP Response. Possibly as a new type.

bowtie/_core.py
```diff
@@ -625,6 +643,144 @@ async def start(
...
+                progress.update(
+                    task,
+                    description=(
+                        "Successfully downloaded all versioned "
+                        f"reports of {id} for {pretty_names_str}!"
+                    ),
+                    completed=actual_downloaded_files,
+                    total=actual_downloaded_files,
+                )
+                await asyncio.sleep(2)
```
Julian, line 780: ??

bowtie/_cli.py
```diff
@@ -20,6 +20,15 @@
 from click.shell_completion import CompletionItem
 from diagnostic import DiagnosticError
 from rich import box, console, panel
+from rich.console import Console
```
Julian, line 23: This is already imported above.

adwait-godbole:
I am making some changes to this PR based on your suggestions. Will ping you soon to review once again.

bowtie/_cli.py
```diff
@@ -1096,6 +1107,19 @@ def _do_nothing(*args: Any, **kwargs: Any) -> CaseTransform:
 POSSIBLE_DIALECT_SHORTNAMES = ", ".join(sorted(Dialect.by_alias()))
 
 
+def pretty_names_str_for(dialects: Iterable[Dialect]) -> str:
```
Julian, line 1110: This is a nice kind of UX function -- but we can use [`inflect`](https://pypi.org/project/inflect/) for this, it has a `.join` function which does this.

Julian: (And your special case will just be `== Dialect.known()` more or less at that point.)

bowtie/_cli.py
```diff
@@ -1671,6 +1695,485 @@ async def info(
...
+    dialects: Iterable[Dialect],
+    format: _F,
+):
+    """
+    Show trend data of an implementation across its multiple versions.
+    """
+    id = connectable.to_terse()
+    versions = asyncio.run(download_versions_of(id))
+
+    downloaded_reports = asyncio.run(
```
Julian, line 2097: You have a small tendency to sprinkle `asyncio.run` in places -- that's a top level function which starts a loop and runs some main thing -- it's not something that should really appear multiple times in a program -- in other functions we start one via `asyncio.run` before actually running a subcommand, you possibly might be looking for that pattern.

In Javascript this function doesn't really exist (AFAIK), there's a single global event loop -- in Python, there can be many (and not even just one event loop framework).

bowtie/schemas/cli/trend.json
```diff
@@ -0,0 +1,41 @@
+{
```
Julian, line 1: Tests? For this and generally?

bowtie/schemas/cli/trend.json
```diff
@@ -0,0 +1,41 @@
...
+  "description": "Trend data of supported implementation(s), emitted by `bowtie trend --format json`.",
+
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+
+  "$id": "tag:bowtie.report,2024:cli:trend",
+
+  "type": "object",
+
+  "$defs": {
+    "count": {
```
Julian, line 12: This looks nearly identical to the `test failures` schema from summary.json.

bowtie/schemas/cli/trend.json
adwait-godbole: I am working on the code for accepting the `.tar` input file so we don't really have to make real **HTTP** calls for `bowtie trend` tests.

bowtie/_core.py
```diff
@@ -5,7 +5,13 @@
 from functools import cache
 from importlib.resources import files
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Protocol, TypeVar, cast
+from typing import (
```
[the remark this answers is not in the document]
adwait-godbole: Sorry. Don't know how this occurred. Probably ruff did something. Will check once again.

bowtie/_cli.py
adwait-godbole: Any suggestions you have for this ? Basically I had to do this way because firstly i need to download the versions for the `ConnectableId` that was passed and then later I want to download reports using that versions metadata that was fetched. So should I do all of this in the `--implementation` option callback itself ? Is that a better place for running `asyncio.run` as you say before actually running a subcommand ?

Julian: I don't really see the issue, this sounds like normal asynchronous code (and looks like it from the body of this function) -- if you need to do 2 async things, call the async code, await if you need to wait for the result in order to continue, then do the other one, await if need be, etc., just do it all within one event loop rather than starting and stopping an event loop each time you want to do an async thing.

Julian:
The demo makes it look like (though I haven't confirmed in the code yet) that you're doing something like "fetch all versions, then download all versions, then parse all versions, then output".

But that's slow and unnecessary -- you should be doing "fetch versions, download each version, don't wait for all responses, as each one comes in, start parsing it immediately" this way you are already starting to do parsing + calculating work while other responses are coming back.

bowtie/_cli.py
```diff
@@ -1671,6 +1697,617 @@ async def info(
...
+
+        async with httpx.AsyncClient(timeout=10) as client:
+
+            async def download_versioned_report_for(
+                version: str,
+                dialect: Dialect,
+            ):
+                try:
+                    url = (
+                        HOMEPAGE
```
Julian, line 1796: These kinds of details (what path the report lives at) should get encapsulated inside `Implementation` -- possibly by having a method that returns an async iterator which yields all the reports, though I think as I mentioned elsewhere maybe that makes progress reporting harder, but some of that I think is due to what I just mentioned -- you "can't" report progress in this way if it means you're uselessly waiting for all responses before getting along with other stuff (parsing).

adwait-godbole:
> The demo makes it look like (though I haven't confirmed in the code yet) that you're doing something like "fetch all versions, then download all versions, then parse all versions, then output".

But that's slow and unnecessary -- you should be doing "fetch versions, download each version, don't wait for all responses, as each one comes in, start parsing it immediately" this way you are already starting to do parsing + calculating work while other responses are coming back.

Fixed this!

Now it looks like this

https://github.com/user-attachments/assets/38449d76-00e4-4cd0-8d4a-d1cc8a55b315

(P.S. It looks slow in this recording because of my recording software (OBS), but it is fast in realtime now since we aren't waiting for all responses and parsing as soon as one comes)

adwait-godbole:
hi @Julian, 

I have added the tests for JSON, markdown and valid markdown formats by using the same `miniatures.foo_v1` and `miniatures.foo_v2` or atleast have tried to so. Can you please have a look once? Thanks. Also, I have made `miniatures.foo_v1` to be always invalid so we get the variation in versions results you were talking about yesterday. 

These two miniatures though they don't have the same `ConnectableId` which is correct as even "real" implementation's versioned images also don't have same `ConnectableId`s since they are treated as "different" implementations altogether just like in this case of miniatures as well. So I just thought of clubbing them together under the same name i.e. "foo" and treat them as versions of that "foo" itself (this is for `bowtie trend` (to create a `foo` directory in the `.tar`) and even for `bowtie summary`, its test of `test_summary_show_failures_markdown_different_versions` also considers `ImplementationInfo.id` ({language}-{name}) as the primary unique name to implement the `Counter` for showing versions in `brackets` in the markdown table as their `ConnectableId`s would be different.) 

I am still trying to figure out the `Direct.from_id` and `kwargs` thing we were discussing yesterday. 
I guess I am just a slow learner :(. I keep re-reading through your messages. Huh. So sorry to let you down.

adwait-godbole:
@Julian any blockers over here to merge now?

Julian:
I was waiting on a response from you on https://github.com/bowtie-json-schema/bowtie/pull/1384/files#r1715404292 I think at least -- did we discuss that elsewhere, I forget

bowtie/_cli.py
adwait-godbole: > You did say earlier to not use rich specific stuff in _core.py

Also this.

Julian: I'm definitely not suggesting you do any rich things in `_core`, which is why I was focused simply on the logic and encapsulation bits, not on presentation.

There is way too much code here to go in the CLI as opposed to somewhere upstream of it -- I'll try to have a look to give more specific suggestions.

bowtie/schemas/cli/summary.json
```diff
@@ -77,32 +77,37 @@
         "unevaluatedItems": false,
         "prefixItems": [
           { "$ref": "tag:bowtie.report,2024:models:implementation:id" },
-          {
-            "description": "Unsuccessful tests for a specific implementation.",
+          { "$ref": "#/$defs/unsuccessfulTests" }
```
Julian, line 80: Please send a PR doing these schema changes separately/immediately, as we can merge those immediately.

bowtie/tests/test_integration.py
```diff
@@ -2995,6 +3031,447 @@ async def test_suite_not_a_suite_directory(tmp_path):
...
+        tar_path,
+        exit_code=EX.DATAERR,
+    )
+
+    assert "Couldn't find a 'foo' directory" in stderr, stderr
+
+
+@pytest.mark.asyncio
+@pytest.mark.containers
+async def test_trend_no_matrix_versions_file(tmp_path):
```
Julian, line 3059: This is a fragile test, especially considering we know we want to remove this file. The file is an implementation detail -- we shouldn't have it appear in the error message, the important bit is that we don't know what versions of an implementation exist without it, but that's the actual important behavior which you should assert against (it's still fine for the filename to appear in the error message whilst we have it, but that's not the assertion you should make).

bowtie/tests/test_integration.py
```diff
@@ -2995,6 +3031,447 @@ async def test_suite_not_a_suite_directory(tmp_path):
...
+                    ],
+                ],
+            ],
+        ],
+    ]
+    assert stderr == ""
+
+
+@pytest.mark.asyncio
+@pytest.mark.containers
```
Julian, line 3278: This marks this test as requiring containers. I don't see anything here requiring containers, you're using a direct connectable (which you should use as often as possible).

bowtie/schemas/cli/summary.json
adwait-godbole: Done!

bowtie/tests/test_integration.py
adwait-godbole: > (it's still fine for the filename to appear in the error message whilst we have it, but that's not the assertion you should make).

So by the above comment, are you saying that I leave this test (`test_trend_no_matrix_versions_file`) as it is and write a new test let's say `test_trend_correct_versions` which makes the assertion you're saying that the versions we passed while constructing the `.tar` i.e. `versions=frozenset(["1.0", "2.0"])` are the same as that after reading `matrix-versions.json` file that the tar utility created ? Should I check it this way ?

bowtie/_cli.py
adwait-godbole: Okay.

bowtie/tests/test_integration.py
Julian: I'm saying make the error message you emit start with something that doesn't specifically mention the filename, to change the name of the test to `test_trend_cannot_detect_versions`, and to make its assertion assert against the beginning of the error message which shouldn't contain the filename, so that in the future this test will not fail just because we switch how we detect versions.

adwait-godbole: I tried fixing this. Is it correct now?

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+                                    ),
+                                )
+                            else:
+                                if report_content:
+                                    versioned_reports.append(
+                                        (
+                                            version,
+                                            dialect,
+                                            _report.Report.from_serialized(
+                                                cast(IO[Any], report_content),
```
Julian, line 2150: This is wrong. You are passing an iterable of bytes to something that claims to take an iterable of str. You calling `cast` just throws away information to make it seem like this type is correct, when it isn't. If you want to "lie" to the type checker, just use ignore. But here, that's not what you should do.

You should have instead either:

* Change the type signature of `from_serialized` to say it accepts an iterable of either, because really it calls json.loads, which does take both.
* Or instead of passing the bytes stream, use `io.TextIOWrapper` to get an iterable of str file-like object instead.

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+                        f"{id} directory of {input.name}.)",
+                    )
+                    ctx.exit(EX.DATAERR)
+
+                versions_dirs = [
+                    member.name
+                    for member in members
+                    if member.isdir()
+                    and (
+                        member.name.removeprefix(f"./{id}/").removeprefix("v")
```
Julian, line 2079: Why are you prepending `v` to the versions -- besides making that a bit awkward that you now have to strip the prefix in a bunch of places, it's also perfectly possible that the version for an implementation *already* starts with `v`, isn't it? What if the implementation claims its version is `v1.2.3` already, and now you added a second `v`?

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+                    description=(
+                        "Successfully downloaded and parsed all versioned "
+                        f"reports of {id} for {pretty_names_str}!"
+                    ),
+                    completed=actual_downloaded_files,
+                    total=actual_downloaded_files,
+                )
+                return responses
+    else:
+        total_files = len(list(dialects))
```
Julian, line 1859: `len(list(...))` is basically always wrong. Just `len`, or else if it's an iterator, you need to make it a list somewhere earlier.

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+        TimeElapsedColumn(),
+        console=console.Console(),
+        transient=True,
+    )
+    task = progress.add_task(
+        description=f"Fetching versions of {id}",
+        total=None,
+    )
+
+    async with httpx.AsyncClient(timeout=10) as client:
```
Julian, line 1731: You repeat more or less the same logic in 3 places here in retrieving a report via HTTPX -- it seems like you would benefit from introducing a `Report.fetch(url: URL, client: AsyncClient | None = None)`.

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+                total_files = len(versions) * len(list(dialects))
+                task = progress.add_task(
+                    description=(
+                        f"Preparing to parse all versioned reports of "
+                        f"{id} for {pretty_names_str} found in {input.name}"
+                    ),
+                    total=total_files,
+                )
+
+                actual_parsed_files = [0]
```
Julian, line 2116: Why is this a list, you don't appear to ever append to it or anything other than dealing with the one int.

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+    for dialect, report in dialects_trend.items():
+        test = "tests" if report.total_tests != 1 else "test"
+        inner_table_rows: list[list[str]] = []
+        if versions:
+            for version, unsuccessful in report.latest_to_oldest():
+                inner_table_rows.append(
+                    create_inner_table_row(version, unsuccessful),
+                )
+        else:
+            implementation = report.implementations[id]
```
Julian, line 1996: This seems like again it should be handled somewhere earlier -- if there's no versions, make wherever you get `versions` from just contain the one version.

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+            str(len(unsuccessful.skipped)),
+            str(len(unsuccessful.errored)),
+            str(len(unsuccessful.failed)),
+        ]
+
+    for dialect, report in dialects_trend.items():
+        test = "tests" if report.total_tests != 1 else "test"
+        inner_table_rows: list[list[str]] = []
+        if versions:
+            for version, unsuccessful in report.latest_to_oldest():
```
Julian, line 1991: Make this a list comprehension, especially once you remove the other branch, which then will let you delete the `create_inner_table_row` function and just do this all inline.

bowtie/_cli.py
```diff
@@ -1684,6 +1708,623 @@ async def info(
...
+        )
+        row_data = [
+            dialect.pretty_name,
+            inner_markdown_table,
+            f"**{report.total_tests} {test} ran**",
+        ]
+        rows_data.append(row_data)
+
+    final_content: Iterable[str] = []
+    for _, row_data in enumerate(rows_data):
```
Julian, line 2015: You are calling enumerate and not using the index? So just don't call enumerate. And this too should just be a list comprehension, don't do `for foo in bar: baz.append(...)` ever basically.

Julian (commented):
I just did a first pass here, sorry for taking so long to get to it. I need to look closer here, but it seems like there are perhaps a few areas you could clean this up a bit. Will try to do a more careful review over the next few days.

Thanks again as usual!

bowtie/_cli.py
adwait-godbole: > Why are you prepending v to the versions -- besides making that a bit awkward that you now have to strip the prefix in a bunch of places, it's also perfectly possible that the version for an implementation already starts with v, isn't it? What if the implementation claims its version is v1.2.3 already, and now you added a second v?

Yeah right! Sorry for this. I will have to fix this in the workflows as well since the problem roots from there.

adwait-godbole: So remove the prepended `v` completely from everywhere right ?

bowtie/_cli.py
adwait-godbole: > This seems like again it should be handled somewhere earlier -- if there's no versions, make wherever you get versions from just contain the one version.

I think the solution for this would be better served once we manage to get PR #1487 merged. Since after that `get_versions` would simply be available on `Implementation` itself and then that will be single handedly responsible for returning the correct `versions` data back and then I won't have to deal with just just `ConnectableId`s for `bowtie trend` (cause currently that's how I have set this up (`bowtie trend` is not really starting an implementation at all right now)).

bowtie/_cli.py
adwait-godbole: I think the fix for this would also get easier and the reasoning for that is the same as the above comment.

bowtie/_cli.py
adwait-godbole: and what about the progress reporting?

bowtie/_cli.py
adwait-godbole:
```python
                )
                return responses
    else:
        total_files = len(list(dialects))
```

here for the lines you've pointed out, the code is only entering the else block if versions were empty (again something that would get fixed by the other PR as pointed out below). So list(dialects) is only happening once at this place.

Julian:
Let's merge these so we see your work live. We can always tidy it later. Thanks again for this and all your hard work this GSoc!

PR #1397, opened 2024-07-31
Chore: dev.harrel implementation - add support for draft7

harrel56:
Hopefully, for any new spec draft support it will happen automatically from now on

----
📚 Documentation preview 📚: https://bowtie-json-schema--1397.org.readthedocs.build/en/1397/

### DIFF
--- a/implementations/java-json-schema/BowtieJsonSchema.java
+++ b/implementations/java-json-schema/BowtieJsonSchema.java
@@ -21,7 +21,9 @@ public class BowtieJsonSchema {
 
   private static final String RECOGNIZING_IDENTIFIERS =
       "Determining if a specific location is a schema or not is not supported.";
-  private static final Map<String, String> UNSUPPORTED = Map.of("$ref prevents a sibling $id from changing the base uri", RECOGNIZING_IDENTIFIERS);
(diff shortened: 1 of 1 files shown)


harrel56:
@Julian well, I don't remember what was wrong with the impl images - please let me know if I should correct something?

Julian:
Thanks! (The CI failure is some new version change that I haven't gotten a chance to look at...)

Julian:
You tripped arguably a minor bug, though a slightly nuanced one which you may not have intended so I'll share it here -- you're returning that the Draft 07 dialect ID is http://json-schema.org/draft-07/schema -- but it isn't, it's http://json-schema.org/draft-07/schema# with an empty fragment. As URIs, those are generally *not* equivalent, so Bowtie is left a bit confused. However as a media type, *JSON Schema* considers those URIs equivalent, so really Bowtie should as well. But yeah you may want to consider fixing the URI in your implementation regardless of that I'll put that fix in here.

harrel56:
@Julian I actually had some problems already regarding this trailing fragment :D it indeed should return the URI with an empty fragment, I totally missed that, gonna fix it soon. Thanks for pointing that out

PR #1401, opened 2024-08-03
Graphing trend on UI

adwait-godbole:
@Julian

----
📚 Documentation preview 📚: https://bowtie-json-schema--1401.org.readthedocs.build/en/1401/


AgniveshChaubey:
Hey @adwait-godbole, when raising a PR, it's always a good practice to write a short description of the changes/features/improvements you wanna achieve through the PR. This helps reviewers a lot by making it easier to understand the context at first glance. Can you do that for this one as well?

adwait-godbole:
Hi @AgniveshChaubey this PR is a part of my GSoC project (`bowtie trend`). It introduces a visualization aspect using a line graph on the UI to better visualize a trend of how different versions of some particular implementation have differed in their failed tests, errored tests, etc. on different dialect test suites.

adwait-godbole:
If you can run this on your local once it will surely give you a better understanding.

adwait-godbole:
Meanwhile @Julian, can you please also have a look at this PR so I have something to show at https://bowtie.report.

adwait-godbole:
Here's the demo:


https://github.com/user-attachments/assets/e05f9a6f-8a0f-491f-bede-8910fc628d12

AgniveshChaubey:
Hey @adwait-godbole, the graph looks good in the first iteration already! I haven't had much time to look at the codebase thoroughly, but here are the few things I noticed at first glance...

1. The dialect dropdown (at the top right side) is not visible in the dark mode.

image
 
 2. The version specific tooltip seems to have visibility issues in the dark mode due to faded texts.
 
image

AgniveshChaubey:
One more thing... how are you collecting the version specific test results? The loading time for implementation-specific pages is very slow (nearly 5-7 seconds) in my case.

adwait-godbole:
> The dialect dropdown (at the top right side) is not visible in the dark mode.
> The version specific tooltip seems to have visibility issues in the dark mode due to faded texts.

I am yet to fix it in the dark mode 😅. For now, preoccupied with other stuff, will at least try to fix it soon enough but if not then will have to go ahead with this in the light mode itself. Sorry.

> One more thing... how are you collecting the version specific test results? The loading time for implementation-specific pages is very slow (nearly 5-7 seconds) in my case.

Well that's a complex process, the version specific test results are collected from such urls:

https://bowtie.report/implementations/python-jsonschema/v4.23.0/draft7.json
https://bowtie.report/implementations/python-jsonschema/v4.22.0/draft7.json
https://bowtie.report/implementations/python-jsonschema/v4.21.1/draft7.json
https://bowtie.report/implementations/python-jsonschema/v4.21.0/draft7.json
..
..
..

etc. etc. and the implementations versions metadata is collected from: 
https://bowtie.report/implementations/python-jsonschema/matrix-versions.json

I have designed the frontend logic in such a way that when the user navigates to an implementation page let's say `python-jsonschema` then on the route async loader, it loads the very latest dialect's trend data (currently meaning 2020-12) and passes it on to the component. Then later only when the user selects the dropdown to change the dialect, only then that particular dialect's versioned reports data gets fetched and gets stored in a `Map` so we don't have to re-fetch it again when user navigates to some other dialect in the dropdown and comes back to it.

adwait-godbole:
> One more thing... how are you collecting the version specific test results? The loading time for implementation-specific pages is very slow (nearly 5-7 seconds) in my case.

Also not really sure why it's taking so long on your device. On mine it takes only roughly about 2-3 seconds or so. Maybe some internet speeds mismatch ig.

adwait-godbole:
> I have designed the frontend logic in such a way that when the user navigates to an implementation page let's say python-jsonschema then on the route async loader, it loads the very latest dialect's trend data (currently meaning 2020-12) and passes it on to the component. Then later only when the user selects the dropdown to change the dialect, only then that particular dialect's versioned reports data gets fetched and gets stored in a Map so we don't have to re-fetch it again when user navigates to some other dialect in the dropdown and comes back to it.

Now that I retrospect on this decision, ig I should just remove any async route loader and let the user first select a dialect from the dropdown and only then show the trend data. This will also reduce the implementation page's load time (the issue you were facing).

Will do this change by tomorrow.

adwait-godbole:
> 1. The dialect dropdown (at the top right side) is not visible in the dark mode.

> 2. The version specific tooltip seems to have visibility issues in the dark mode due to faded texts.

> Now that I retrospect on this decision, ig I should just remove any async route loader and let the user first select a dialect from the dropdown and only then show the trend data. This will also reduce the implementation page's load time (the issue you were facing). Will do this change by tomorrow.

@AgniveshChaubey @Julian fixed all of the above. Can we get this merged ?

adwait-godbole:
@AgniveshChaubey @Julian here's a demo in both dark and light modes


https://github.com/user-attachments/assets/f886a867-fa5b-4141-941f-c93f015a2eb7

AgniveshChaubey:
Hi @adwait-godbole, replying via email as I'm not on my PC right now. Will have a closer look at the code tonight. UI looks good now but instead of not showing anything initially, it'd be better to show the compliance trend for the latest dialect i.e., 2020-12

adwait-godbole:
> Hi @adwait-godbole, replying via email as I'm not on my PC right now. Will have a closer look at the code tonight. UI looks good now but instead of not showing anything initially, it'd be better to show the compliance trend for the latest dialect i.e., 2020-12

I did have it like this way! 🙂 until yesterday when you said that it was taking about 5-7 seconds on your machine to load. So I shifted to this design where we don't really load / fetch the data until the user himself doesn't explicitly select the dialect for which he wants to see the trend data for and now I think that this might be a better way since there is a lot of data to be loaded for showing the trend which unnecessarily makes the routing slower. Still would also like hear out @Julian's thoughts on this.

Julian:
Not showing anything immediately doesn't look good to me. I didn't read super carefully but I wouldn't expect it to be slow to load all data? It's not that much data. It should be roughly the same amount of time as what you show there to load just 1 dialect as long as you do it in parallel -- and it will feel like even less, because the user will possibly have to scroll before they see the graph.

adwait-godbole:
> Not showing anything immediately doesn't look good to me. I didn't read super carefully but I wouldn't expect it to be slow to load all data? It's not that much data. It should be roughly the same amount of time as what you show there to load just 1 dialect as long as you do it in parallel -- and it will feel like even less, because the user will possibly have to scroll before they see the graph.

Cool. Okay. Will revert back to what I had.

adwait-godbole:
> Not showing anything immediately doesn't look good to me. I didn't read super carefully but I wouldn't expect it to be slow to load all data? It's not that much data. It should be roughly the same amount of time as what you show there to load just 1 dialect as long as you do it in parallel -- and it will feel like even less, because the user will possibly have to scroll before they see the graph.

Fixed this! @Julian

adwait-godbole:
@AgniveshChaubey hope you find some time today to review this. Thanks.

AgniveshChaubey:
Seems better now! 
I saw you have used `FC` in many components to define the type of Props. But why import a new component (i.e., `FC `) if you can directly specify the type of the prop just by doing this:

`const DialectCompliance = ({implementation, dialectsCompliance} : Props) => {...}`

After React 18, using `FC` is discouraged. 
Check [this](https://github.com/facebook/create-react-app/pull/8177) for a better understanding.

Also, I found a lot of inconsistencies in code formatting and couple of snake case variables. In js we usually prefer camelCase variables. can you fix those as well?

adwait-godbole:
> I saw you have used FC in many components to define the type of Props. But why import a new component (i.e., FC ) if you can directly specify the type of the prop just by doing this:

> const DialectCompliance = ({implementation, dialectsCompliance} : Props) => {...}

> After React 18, using FC is discouraged.
> Check https://github.com/facebook/create-react-app/pull/8177 for a better understanding.

Fixed this!

> Also, I found a lot of inconsistencies in code formatting and couple of snake case variables. In js we usually prefer camelCase variables. can you fix those as well?

Regarding the snake casing you're talking about, I could only find `Dialect.newest_to_oldest()` which I now fixed it to `Dialect.newestToOldest()`. Lmk what others you saw atleast in this PR for now.

@AgniveshChaubey 
Thanks!

adwait-godbole:
Hi @AgniveshChaubey, any updates over here?

AgniveshChaubey:
HI Adwait, I have gone through the code and rest of the things seems fine to me. The only thing I am concerned about is the n^3 time complexity due to `implementation.fetchVersionedReportsFor` and `prepareVersionsComplianceReportFor` functions in the `VersionsTrend.tsx` component. Let me see if I can optimize it.

adwait-godbole:
> `implementation.fetchVersionedReportsFor`

I don't see this having an n^3 time complexity. All the JSON files are being fetched concurrently using `Promise.all`.

>  `prepareVersionsComplianceReportFor`

I don't see this as well having an n^3 time complexity. Though I do see it having an n^2 time complexity in it. I will optimize that.

AgniveshChaubey:
the `versions.map(...)` loop inside `fetchVersionedReportsFor` and the two nested loops inside `prepareVersionsComplianceReportFor` collectively makes the time complexity n^3.

adwait-godbole:
> the versions.map(...) loop inside fetchVersionedReportsFor and the two nested loops inside prepareVersionsComplianceReportFor collectively makes the time complexity n^3.

I just optimized it a bit. Can you have a look?

AgniveshChaubey:
I think it's possible to reduce the time complexity to the order of O(n). I'm working on it. I'll update you as soon as I find a way!

adwait-godbole:
> I think it's possible to reduce the time complexity to the order of O(n). I'm working on it. I'll update you as soon as I find a way!

Okay. Cool. While doing your optimizations, can you just make sure one thing that we don't change `fetchVersionedReportsFor`'s return type i.e. `Map<string, ReportData>`. I am saying this because later on I also want to add `diffing` between two versions reports to see what results changed from one version to another version. If you change it's return type from `Map<string, Partial<Totals>>` then later on will again have to change the code to get the actual entire `ReportData` and not just the `Partial<Totals>` corresponding to a version.

AgniveshChaubey:
>Okay. Cool. While doing your optimizations, can you just make sure one thing that we don't change the return type of fetchVersionedReportsFor function i.e. Map<string, ReportData>.

Sure, I'll take care of that.

Right now you have used `fetchVersionedReportsFor` function in the `VersionsTrens.tsx` component only, right?

adwait-godbole:
> Sure, I'll take care of that.

> Right now you have used fetchVersionedReportsFor function in the VersionsTrens.tsx component only, right?

I just optimized it even further @AgniveshChaubey. Can you have a look? :)

adwait-godbole:
Just a sec.

adwait-godbole:
Any further optimizations that you seem to notice?

AgniveshChaubey:
I'll have a look at it shortly. BTW, whenever you update anything in the PR, always try to mention the specific updates or changes you made in the conversation. :)

AgniveshChaubey:
The updated one seems a bit more slower. 😅 Previously, the processing time was around 4.0 to 4.2 seconds, but now it is around 4.6-4.7 seconds.

adwait-godbole:
> The updated one seems a bit more slower. 😅 Previously, the processing time was around 4.0 to 4.2 seconds, but now it is around 4.6-4.7 seconds.

Well. But now the time complexity doesn't really look like being n^3 though as you were saying earlier or does it still?

AgniveshChaubey:
> Well. But now the time complexity doesn't really look like being n^3 though as you were saying earlier or does it still?

Theoretically, it is of n^3 complexity, but technically it is just n^2. and that too n^2 doesn't create a major difference. Let me explain...

Previously, the `fetchVersionedReportsFor` function was fetching the data for all the versions, and the `prepareVersionsComplianceReportFor` function was preparing the final data to be consumed by the UI. (taking up n^2 complexity). However as we don't have an extremely large set of data, the n^2 time taken to prepare the data is relatively very less as compared to the time taken by the `fetchVersionedReportsFor` function to fetch the raw data over the network. 

I have optimized it a bit to bring down the final loading time to 3.7-3.9 seconds, and I don't think we can do anything much for the data fetching part as you have already used the promises to do the parallel operations.

adwait-godbole:
> The updated one seems a bit more slower. 😅 Previously, the processing time was around 4.0 to 4.2 seconds, but now it is around 4.6-4.7 seconds.

Also one more thing is that although we are currently fetching the data from our localhosts, by default it's a bit slower already but when it gets deployed, at that time it will be fetched from the deployed assets/artifacts itself. Ik this is not a good reason to justify the slower time but still a factor to consider.

adwait-godbole:
> I have optimized it a bit to bring down the final loading time to 3.7-3.9 seconds, and I don't think we can do anything much for the data fetching part as you have already used the promises to do the parallel operations.

Curious to know how you're profiling the performance

adwait-godbole:
Btw I have removed the `prepareVersionsComplianceReportFor`. Did you see that?

AgniveshChaubey:
> Also one more thing is that although we are currently fetching the data from our localhosts, by default it's a bit slower already but when it gets deployed, at that time it will be fetched from the deployed assets/artifacts itself. Ik this is not a good reason to justify the slower time but still a factor to consider.

makes sense.

> Curious to know how you're profiling the performance

Initially, you were using `prepareVersionsComplianceReportFor` to prepare the final data to be consumed by the UI. We can simply just prepare the final data within the `fetchDialectTrendData` itself to avoid unnecessary looping to prepare the final data.

AgniveshChaubey:
> Btw I have removed the prepareVersionsComplianceReportFor. Did you see that?

Oh! I didn't see that part!

adwait-godbole:
> Initially, you were using prepareVersionsComplianceReportFor to prepare the final data to be consumed by the UI. We can simply just prepare the final data within the fetchDialectTrendData itself to avoid unnecessary looping to prepare the final data.

did you fetch/pull my latest commit?

AgniveshChaubey:
Okay, you are preparing the final data inside the `setDialectsTrendData((prev) =>...` fn, which I did in the `fetchVersionedReportsFor` function itself (however, I have changed the return type here). 

But given that you need the return type of `fetchVersionedReportsFor` to remain unchanged, we are good to go with your latest changes.

adwait-godbole:
> Okay, you are preparing the final data inside the setDialectsTrendData((prev) =>... fn, which I did in the fetchVersionedReportsFor function itself (however, I have changed the return type here).

> But given that you need the return type of fetchVersionedReportsFor to remain unchanged, we are good to go with your latest changes.

Yeah. Thanks a lot for pointing out all of this. Did one more minor optimization of removing `getTotals` as well. But overall can we then tell @Julian to merge this one then?

AgniveshChaubey:
> Yeah. Thanks a lot for pointing out all of this. Did one more minor optimization of removing getTotals as well :). But overall can we tell @Julian to merge this one then?

Let Julian also have a look at it, then yeah we can proceed further.

adwait-godbole:
@Julian this is also ready for a final review by you now that I have fixed issues that were pointed out by Agnivesh.

PR #1408, opened 2024-08-06
Publish an implementation metadata file suitable for use by json-schema.org

### LINKED ISSUE #1312: Publish a JSON File with all the implementations details suitable for consumption by json-schema.org

adwait-godbole:
Fixes #1312 

@Julian can you review this one so that we can get this file available to @benjagm ASAP. Also earlier you had told me to remove the `.json` extension from the file, are you still ok with that ? Removing that extension directly downloads the file onto your desktop rather than opening on the Bowtie website.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1408.org.readthedocs.build/en/1408/


adwait-godbole:
Also @benjagm what all keys do you expect to be included in this API corresponding to an implementation?

Julian:
> Removing that extension directly downloads the file onto your desktop rather than opening on the Bowtie website.

Only if you view it in a browser, for API consumption it doesn't matter either way. (So yes let's remove it).

I'm confused though., didn't we merge this already? Having a look.

bowtie/schemas/public-implementations.json
```diff
@@ -0,0 +1,77 @@
+{
+  "title": "Public Implementations",
```
Julian, line 2: You're focused a bit too much on the word "public". Public simply means here' we're really providing an API for others to consume, as opposed to just something we (Bowtie, the people who work on Bowtie) use ourselves.

But "public" shouldn't really appear anywhere. Here let's put this file in `bowtie/schemas/api/v1/json-schema-org/implementations.json`.

.github/workflows/report.yml
```diff
@@ -82,15 +82,48 @@ jobs:
...
           version: ${{ inputs.bowtie-version }}
 
-      - name: Generate implementations.json file
+      - name: Generate implementations.json (Private API) file
         run: |
           bowtie info $(bowtie filter-implementations | sed 's/^/-i /') --format json > implementations.json
 
+      - name: Generate public-implementations (Public API) file
+        run: |
+          bowtie filter-implementations | \
```
Julian, line 91: This is going to essentially immediately bite us, as this code is essentially incomprehensible. But really we need to get `bowtie site` to be a way of generating all of what we need from Python rather than writing YAML shell code. So I guess for right this second it's ok.

bowtie/schemas/public-implementations.json
```diff
@@ -0,0 +1,77 @@
...
+  "description": "A Public API providing metadata about implementations of JSON Schema supported by Bowtie",
+
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+
+  "$id": "tag:bowtie.report,2024:models:public-implementations",
```
Julian, line 7: This isn't a model (a model is like "a concept", so e.g. an implementation class' schema). Let's make its ID be similar to the above -- `tag:bowtie.report,2024:api:v1:json-schema-org:implementations`.

It should really also share as much as possible with the definitions -- i.e. rather than redefining all the properties we should share them. That though requires moving some models to a common place, so we can perhaps do that after as well.

bowtie/schemas/public-implementations.json
```diff
@@ -0,0 +1,77 @@
...
+    "properties": {
+      "name": { "type": "string" },
+      "language": {
+        "description": "The implementation language (e.g. C++, Python, etc.)",
+
+        "type": "string",
+        "pattern": "^[a-z][a-z0-9-+_]*$"
+      },
+      "dialects": {
+        "description": "A list of JSON Schema dialects (URIs) which the implementation understands. When running test cases, this list will be consulted before sending them to the implementation (and any unsupported dialects will be skipped).",
```
Julian, line 28: The end of this description isn't relevant to this use case (there's no running of test cases as far as json-schema-org is concerned).

Julian:
> Also @benjagm what all keys do you expect to be included in this API corresponding to an implementation?

Looks like Benja shared what's expected [here](https://github.com/bowtie-json-schema/bowtie/issues/1312#issue-2360229179) -- so all we need is a mapping where the keys are source URLs for implementations, and the values should have:

* bowtie's ID (e.g. `python-jsonschema`)
* the `dialects` key.

We can skip everything else for now.

adwait-godbole:
Doing the changes.

adwait-godbole:
> the badges URLs

Is there a way of generating those URLs programmatically using `bowtie` 😅?

Julian:
I think that logic lives in the UI code unfortunately right now which I think you've seen last -- so not directly -- you could have a small typescript script maybe which imports the UI file, but not sure that's simpler than doing the hairy thing again :/

.github/workflows/report.yml
```diff
@@ -82,15 +82,58 @@ jobs:
...
+                  }
+                )}
+              '
+            done | \
+          jq -s 'add' > public-implementations
+
+      - name: Validate public-implementations against its corresponding JSON Schema
+        run: |
+          VALIDATION=$(bowtie validate --expect valid -i python-jsonschema bowtie/schemas/api/v1/json-schema-org/implementations.json public-implementations)
+          SUMMARY=$(echo "$VALIDATION" | bowtie summary --show failures --format json)
```
Julian, line 120: This seems like a bug, `bowtie summary --show failures` should probably exit with non-zero status when there are failures, and then all of this woud go away and you can just use `--format markdown` probably?

Can you at least file that issue perhaps.

.github/workflows/report.yml
```diff
@@ -114,12 +157,24 @@ jobs:
...
 
-      - name: Include Implementation Metadata
+      - name: Include Implementations Metadata (Private API)
         uses: actions/download-artifact@v4
         with:
           name: implementations
           path: site/
 
+      - name: Create a Public Directory Structure for JSON Schema Org
+        run: mkdir -p site/v1/json-schema-org
```
Julian, line 167: Possibly you're missing an `api/`? But can't you just make stuff above output the right directory structure so that there's no need to rename things here?

test.json
```diff
@@ -0,0 +1,431 @@
+{
```
Julian, line 1: You have a local testing file committed.

(Ideally please force push it out of the history.)

.github/workflows/report.yml
adwait-godbole: Still working on the PR 😅. Was going to ping you once pushing the commits was done.

adwait-godbole: > But can't you just make stuff above output the right directory structure so that there's no need to rename things here?

Have to do the renaming because we want this to be accessible --> https://bowtie.report/api/v1/json-schema-org/implementations and not https://bowtie.report/api/v1/json-schema-org/implementations-file-for-json-schema-org

At this line --> https://github.com/bowtie-json-schema/bowtie/blob/fa9f61e4c47aba3eaea1eaa8bd7c8f1100768bd9/.github/workflows/report.yml#L91

there's already an `implementations` artifact getting created for our private API usage so again a same named artifact can't be created as it will cause conflict while downloading.

adwait-godbole: So I have to essentially create a different named file and then download it and rename it just to avoid conflict. Can't think of a workaround right now for this. Will come back to it if i find something.

Julian: You'd create an artifact called `api` instead, and put *all* of the nested files (exactly 1 for now) in it :)!

adwait-godbole: Right! Thanks :)

adwait-godbole:
@Julian made the final changes. Tested on my local works fine.

You can have a look here --> https://adwait-godbole.github.io/bowtie/api/v1/json-schema-org/implementations

adwait-godbole:
good to merge ?

adwait-godbole:
I am not sure why the tests are failing even after merging with main.

Julian:
Thanks! Let's give it a shot. Well done.

adwait-godbole:
It would be better if you could manually trigger this workflow once the CI passes, so we know if it works or not.

Julian:
I kicked a release!

adwait-godbole:
I hope for e.g. the below output format is fine right ?

```
  "https://github.com/tdegrunt/jsonschema": {
    "id": "js-jsonschema",
    "dialects": [
      "http://json-schema.org/draft-07/schema#",
      "http://json-schema.org/draft-06/schema#",
      "http://json-schema.org/draft-04/schema#",
      "http://json-schema.org/draft-03/schema#"
    ],
    "badges_urls": {
      "supported_versions": "https://bowtie.report/badges/javascript-jsonschema/supported_versions.json",
      "dialects_compliance": {
        "draft7": "https://bowtie.report/badges/javascript-jsonschema/compliance/draft7.json",
        "draft6": "https://bowtie.report/badges/javascript-jsonschema/compliance/draft6.json",
        "draft4": "https://bowtie.report/badges/javascript-jsonschema/compliance/draft4.json",
        "draft3": "https://bowtie.report/badges/javascript-jsonschema/compliance/draft3.json"
      }
    }
  }
```

Julian:
Ah, close, I think ideally the keys in the `dialects_compliance` badges should be dialect URIs, otherwise there's no way for someone to know the short names of dialects (we "invented" them even though they're basically what everyone uses). Can you make that tweak?

Julian:
Probably you can also shorten the key name to `compliance` to match the URI, rather than `dialects_compliance`.

adwait-godbole:
Yup makes sense!

PR #1424, opened 2024-08-08
Combine versioned reports

adwait-godbole:
@Julian the idea here basically was to achieve the following:

If someone does `bowtie trend -i ruby-json_schemer --dialect 7` then in this case, I wanted to combine the below reports:

https://bowtie.report/implementations/ruby-json_schemer/v2.3.0/draft7.json
https://bowtie.report/implementations/ruby-json_schemer/v2.2.1/draft7.json
https://bowtie.report/implementations/ruby-json_schemer/v2.2.0/draft7.json
..
..
..

and so on into one single report so I am able to achieve the below output

Screenshot (180)

----
📚 Documentation preview 📚: https://bowtie-json-schema--1424.org.readthedocs.build/en/1424/


adwait-godbole:
And I am using `semver` here to sort the versions from latest to oldest while displaying.

adwait-godbole:
@Julian can you review this one as well please. Thanks.

bowtie/_report.py
```diff
@@ -9,6 +10,7 @@
 from attrs import asdict, field, frozen
 from attrs.filters import exclude
 from rpds import HashTrieMap
+from semver import VersionInfo
```
Julian, line 13: An arbitrary implementation is not really guaranteed to be using semver (or any semver-resembling scheme) -- so we anyways would need to potentially handle arbitrary version schemes, which makes me think unlike in general (where we certainly should use rich types whenever possible) it's not really worthwhile to use some package with a specific versioning format -- e.g. what will this do for an implementation with version `20+foo`? Will it blow up? If so it seems like we should likely just roll our own simple thing which splits on `.` + `int` but otherwise falls back to simple lexicographic, and doesn't really do anything else (any other validation), unless this package or some other really does do exactly that.

bowtie/_report.py
```diff
@@ -288,6 +296,71 @@ def empty(
...
+    ) -> Report | None:
+        versions_reports = [
+            version_report
+            for version_report in versions_reports
+            if version_report.metadata.dialect == dialect
+        ]
+
+        if versions_reports:
+
+            def combine_cases(
```
Julian, line 313: I'm not sure this seems any clearer or shorter using `reduce` over just doing 2 loops, no? You still have a for loop anyhow.

bowtie/_report.py
adwait-godbole: So in that case, before I start writing the custom thing, should we consider `4.18.0a10` to be greater than `4.18.0` ? and also `20+foo` to be greater than `20` ?

bowtie/_report.py
adwait-godbole: Will do. Also wanted to get one more thing cleared, is CaseMismatchInReportsForSameDialect the correct way I am handling it? It can happen that the same dialect test suite for e.g. draft 7 suite had two different test cases at some point of time.

Is it correct if I just consider `cases=report[0]._cases` ? 

Again just for context all of this is for combining

https://bowtie.report/implementations/ruby-json_schemer/v2.3.0/draft7.json
https://bowtie.report/implementations/ruby-json_schemer/v2.2.1/draft7.json
https://bowtie.report/implementations/ruby-json_schemer/v2.2.0/draft7.json
..
..
..

bowtie/_report.py
Julian: I think it's probably fine for those to sort lexicographically, we likely won't really test prerelease versions anyhow.

bowtie/_report.py
adwait-godbole: Can you answer this one as well? Have made changes to tackle the other review of yours.

adwait-godbole:
Can we merge this then ?

bowtie/_report.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 from datetime import UTC, datetime
+from functools import cmp_to_key
```
Julian, line 4: This function should essentially never be used (it's there to support legacy `cmp` functions).
The version comparison below looks super complicated compared to simply splitting and trying to call int. I'll try to have a look in a bit and give you tips, but I'd expect this to be ~4 lines long essentially.

adwait-godbole:
@Julian is it fine now ?

bowtie/_report.py
```diff
@@ -288,6 +294,57 @@ def empty(
...
+            return cls(
+                cases=cases,
+                results=results,
+                metadata=RunMetadata(
+                    implementations=implementations,
+                    dialect=dialect,
+                ),
+                did_fail_fast=False,
+            )
+        else:
```
Julian, line 345: Can you invert the order of the if blocks (to `if not versioned_reports`) here which will let you dedent the larger block

bowtie/_report.py
```diff
@@ -288,6 +294,57 @@ def empty(
...
+
+            for versioned_report in versioned_reports:
+                for seq, case in versioned_report._cases.items():
+                    existing_case = cases.get(seq)
+                    if not existing_case:
+                        cases = cases.insert(seq, case)
+                    elif case != existing_case:
+                        raise CaseMismatchInReportsForSameDialect()
+
+                version_id, version_results = next(
```
Julian, line 326: This line is a bit strange -- I'm not 100% sure what it is meant to do yet without trying to run this -- you're trying to get the first -- and just the first -- element from `_results`? Why?

adwait-godbole: Because there would be just one in for e.g. https://bowtie.report/implementations/ruby-json_schemer/v2.3.0/draft7.json
while trying to combine in the for loop 😅.

Julian: So is the length *always* 1?

adwait-godbole: This method is essentially just trying to combine

https://bowtie.report/implementations/ruby-json_schemer/v2.3.0/draft7.json
https://bowtie.report/implementations/ruby-json_schemer/v2.2.1/draft7.json
https://bowtie.report/implementations/ruby-json_schemer/v2.2.0/draft7.json
..
..
..

into a single report, so yes, in the for loop for every report we always just have a single  _versioned_ implementation and results for it. And the cases are essentially the same for all of them i.e. of `draft7`. So I might as well just do `cls(cases=reports[0]._cases)` and entirely avoid `CaseMismatchInReportsForSameDialect`. Is it fine this way ?

adwait-godbole:
@Julian I have simplified the logic as much as possible now or at least have tried to do so 😅.

bowtie/_report.py
Julian: It's still hard to really follow here without dropping into a debugger -- please add a test or two for this functionality, which at least will make it a bit clearer, even though I think I follow what this is doing (it's assuming that all combined reports are for the same implementation, different versions, and that each input has one and only one version?)

adwait-godbole: > it's assuming that all combined reports are for the same implementation, different versions, and that each input has one and only one version?

Yes!

Julian: (the tests should/can probably go to `test_report.py` as I think they won't need to be integration tests?)

adwait-godbole: Added. Can you have a look.

bowtie/_report.py
```diff
@@ -288,6 +288,45 @@ def empty(
...
+            return None
+
+        results: HashTrieMap[
+            ConnectableId,
+            HashTrieMap[Seq, SeqResult],
+        ] = HashTrieMap()
+        implementations: dict[ConnectableId, ImplementationInfo] = {}
+
+        for versioned_report in versioned_reports:
+            version_id, version_info = next(
```
Julian, line 313:
```suggestion
            (version_id, version_info), = versioned_report.metadata.implementations.items()
```

bowtie/_report.py
```diff
@@ -288,6 +288,45 @@ def empty(
...
+                iter(versioned_report.metadata.implementations.items()),
+            )
+            implementations[version_id] = version_info
+            version_results = next(iter(versioned_report._results.values()))
```
Julian, line 317:
```suggestion
            version_results, = versioned_report._results.values())
```

bowtie/_report.py
```diff
@@ -288,6 +288,45 @@ def empty(
...
+        dialect: Dialect,
+    ) -> Report | None:
+        versioned_reports = [
+            versioned_report
+            for versioned_report in versioned_reports
+            if versioned_report.metadata.dialect == dialect
+        ]
+
+        if not versioned_reports:
+            return None
```
Julian, line 304: Why None? This seems like it should return an empty report.

adwait-godbole:
Did the changes.

Issue #1388, opened 2024-07-28
Benchmarks should have the notion of `tags` or `markers`


sudo-jarvis:
Benchmarks should have the notion of having `tags` or `markers` like pytest test markers which would allow filtering them in a very convenient method as compared to the current one by a tag/marker. Right now we can have 2 of them, `keyword` for keyword benchmarks and `default` for default benchmarks. Later when more type of benchmarks are added, this would extend to them as well.

PR #1425, opened 2024-08-08
Add some other fields to Benchmarks, Add Multi Dialect support to a benchmark, Improve Benchmark Output & simplify filtering benchmarks by tags

sudo-jarvis:
1. Add some other fields to benchmarks

- varying_parameter: Helps group the benchmark_results.
- benchmark_type: Acts as a tag for filtering benchmarks.
- dialects_supported: Store the dialects that a benchmark supports to be run on.

2. Add multi dialect support to a benchmark.

3. Improve Benchmark Output for pretty and markdown output by grouping according to varying_parameter if possible.

4. Simplify filtering benchmarks according to benchmark_type.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1425.org.readthedocs.build/en/1425/


sudo-jarvis:
@Julian, the tests wont pass in the benchmarks PR if I included it separately so I included their changes as well in this PR. If its too much then I can put it in a separate one and then that one could be merged later then the tests would pass. Do let me know if you want that.

bowtie/_cli.py
```diff
@@ -1442,11 +1442,14 @@ def filter_benchmarks(
...
     Output benchmarks matching the specified criteria.
     """
-    _benchmarks.get_benchmark_filenames(
+    files: list[Path] = _benchmarks.get_benchmark_files(
         benchmark_type,
         benchmarks=benchmark_names,
         dialect=dialect,
     )
+    for file in files:
+        console.Console().file.write(str(file))
```
Julian, line 1451:
```suggestion
        console.Console().file.write(f"{file}\n")
```

bowtie/schemas/benchmark_report.json
```diff
@@ -120,16 +121,30 @@
...
+        "varying_parameter",
+        "benchmark_results"
+      ],
       "properties": {
         "name": {
           "description": "Benchmark Group's Name.",
           "type": "string"
         },
+        "benchmark_type": {
+          "description": "The type of the Benchmark Group (Eg. keyword, default, etc.)",
```
Julian, line 138: Should this be an enum? What does `etc.` mean here? Do we use arbitrary values, or is this basically a preliminary implementation of #1388 ?

bowtie/schemas/benchmark_report.json
```diff
@@ -120,16 +121,30 @@
...
+          "type": "string"
+        },
         "description": {
           "description": "Benchmark Group's Description.",
           "type": "string"
         },
+        "varying_parameter": {
+          "description": "The parameter which is changing in the various benchmarks present in the Benchmark Group (Eg. array_size, object_size, etc.)"
```
Julian, line 146: `type` at least presumably?

bowtie/schemas/benchmarks.json
```diff
@@ -10,16 +10,33 @@
...
+        },
         "description": {
           "type": "string",
           "description": "The description of what the benchmark does."
         },
+        "dialects_supported": {
+          "type": "array",
+          "description": "An array of supported dialects by the benchmark.",
+          "items": {
+            "type": "string"
```
Julian, line 37: This should presumably $reference the dialect ID schema URI.

bowtie/schemas/benchmark_report.json
sudo-jarvis: Yeah, a preliminary implementation of #1388, this should improve going on further. etc. because currently we just have 2 types but going further we could have others too.

Julian: Got it, sounds good!

bowtie/tests/test_benchmarks.py
```diff
@@ -359,6 +350,59 @@ async def test_benchmark_run_markdown_output(
...
+            "-i",
+            DIRECT_CONNECTABLE,
+            "-q",
+            "--format",
+            "markdown",
+            tmp_path / "benchmark.json",
+            exit_code=0,
+        )
+
+        expected_data1 = """
```
Julian, line 376: Can you use `textwrap.dedent` to make these a bit less "ugly" indented in a follow-up PR? Will merge for now.

PR #1428, opened 2024-08-09
Backward compatibility for `python-jsonschema`

adwait-godbole:
@Julian, I haven't included any pre-release versions in the `matrix-versions.json`. That's fine right?

----
📚 Documentation preview 📚: https://bowtie-json-schema--1428.org.readthedocs.build/en/1428/


implementations/python-jsonschema/bowtie_jsonschema.py
```diff
@@ -3,14 +3,46 @@
...
 import sys
 import traceback
 
 from jsonschema.validators import validator_for
-import referencing.jsonschema
+
+jsonschema_version = metadata.version("jsonschema")
+
+
+def compare_versions(v1: str, v2: str) -> int:
```
Julian, line 18: Here jsonschema *does* use Python versioning (and semver) so we have a single known implementation and version scheme -- so here we definitely should use an existing parser (probably the one you had from the packaging repo).

adwait-godbole:
Ok to merge now ? After merge you will have to trigger the `images.yml` for just `python-jsonschema` without specifying any version and it would build all of it and later the `versioned-report.yml` 😅

Julian:
> I haven't included any pre-release versions in the matrix-versions.json. That's fine right?

Yep -- though I think you could have gone back further :) -- but we can always do that after.

adwait-godbole:
> Yep -- though I think you could have gone back further :) -- but we can always do that after.

Further ? 

https://github.com/python-jsonschema/jsonschema/releases?page=7 is the last page right ? Here I only see upto `v4.0.0`

Julian:
Seems like GitHub may not keep all releases or something, but here's the full release history: https://pypi.org/project/jsonschema/#history

adwait-godbole:
Ohh wow. Okay will cover these versions later then. Thanks!

adwait-godbole:
Just a reminder, can you also trigger the `verioned-report.yml`  please 😅?

Julian:
Done!

adwait-godbole:
Thanks!

PR #1430, opened 2024-08-10
Remove deprecated `react-test-renderer`

adwait-godbole:
@Julian as per over [here](https://github.com/facebook/react/blob/main/packages/react-test-renderer/README.md) and also over [here](https://react.dev/warnings/react-test-renderer), `react-test-renderer` seems to have been deprecated so ig its better to use `@testing-library/react` as recommended over there.

Thanks.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1430.org.readthedocs.build/en/1430/

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -17,6 +17,7 @@
     "watch-test": "vitest watch"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


Julian:
Thanks!

PR #1431, opened 2024-08-11
`bowtie summary --show failures` should return nonzero status on failures

### LINKED ISSUE #1414: `bowtie summary --show failures` should exit with non-zero status when there are failures
[Reference](https://github.com/bowtie-json-schema/bowtie/pull/1408#discussion_r1705928170)

adwait-godbole:
Fixes #1414 

----
📚 Documentation preview 📚: https://bowtie-json-schema--1431.org.readthedocs.build/en/1431/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -616,7 +616,8 @@ def summary(report: _report.Report, format: _F, show: str):
             if any(
                 unsuccessful.failed or unsuccessful.errored
                 for _, __, unsuccessful in results
-            ) else 0
+            )
+            else 0
         )
         to_table = _failure_table
         to_markdown_table = _failure_table_in_markdown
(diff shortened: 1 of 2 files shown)


bowtie/_cli.py
```diff
@@ -611,6 +611,14 @@ def summary(report: _report.Report, format: _F, show: str):
     """
     if show == "failures":
         results = report.worst_to_best()
+        exit_code = (
```
Julian, line 614: This is a bit nonideal given it loops over the results multiple times (once here, once eventually when we're processing them). I haven't looked to see whether that's avoidable yet. Hopefully you did first?

Julian: True. OK, let's keep it as is for now then, I want to replace lots of this presentation logic with something a bit more generic at some point. Thanks.

Issue #1332, opened 2024-06-25
`bowtie filter-dialects` should have options to print out other dialect properties instead of just the URIs


adwait-godbole:
For e.g. instead of needing to do `jq -r '.[].shortName' data/dialects.json`, it would've been better if `bowtie filter-dialects` itself had maybe a `--show shortNames` option?

Julian:
This seems reasonable though a better design is probably to allow for format strings, e.g. `filter-dialects --format {.short_name}: {.pretty_name}` rather than as you say needing to add lots of different options.

PR #1458, opened 2024-08-16
Add Benchmarks workflow

sudo-jarvis:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1458.org.readthedocs.build/en/1458/

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -5,8 +5,8 @@ name: Collect New Benchmark Results
 on:
   workflow_dispatch:
   schedule:
-    # Every Monday at 00:00 UTC
-    - cron: "0 0 * * 1"
+    # Every Monday at 08:00 UTC
+    - cron: "0 8 * * 1"
 
 permissions:
   contents: read
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -106,33 +106,21 @@ jobs:
           path: benchmarks-${{ matrix.dialect }}/
           merge-multiple: true
 
+      - name: Install Bowtie
+        uses: ./
+
       - name: Merge Benchmark Reports for a dialect
         run: |
           python - <<EOF
-          import os
-          import json
+          from pathlib import Path
+          from bowtie._benchmarks import combine_benchmark_reports_to_serialized as combine
 
-          directory_path = f'benchmarks-${{ matrix.dialect }}'
-          output_file = f'${{ matrix.dialect }}.json'
-          metadata = None
-          combined_results = []
-
-          for filename in sorted(os.listdir(directory_path)):
-              file_path = os.path.join(directory_path, filename)
-              with open(file_path, 'r') as file:
-                  data = json.load(file)
-                  if metadata is None:
-                      metadata = data.get('metadata', {})
-                  results = data.get('results', [])
-                  combined_results.extend(results)
-
-          combined_json = {
-              'metadata': metadata,
-              'results': combined_results
-          }
-
-          with open(output_file, 'w') as file:
-              json.dump(combined_json, file, indent=4)
+          Path("${{ matrix.dialect }}.json").write_text(
+              combine(
+                  Path("benchmarks-${{ matrix.dialect }}").iterdir()
+              )
+          )
+          
           EOF
 
       - name: Upload final Benchmark Report for dialect
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -79,10 +79,10 @@ jobs:
       - name: Generate Benchmark Report
         run: |
           bowtie perf $(bowtie filter-implementations | sed 's/^/-i /') -b ${{ matrix.benchmark }} -D ${{ matrix.dialect }} -q > ${{ matrix.benchmark }}-file.json
-        
+
       - name: Store Unique Id for distinction
         run: echo "unique_id=$(uuidgen)" >> $GITHUB_ENV
-        
+
       - name: Upload Benchmark file as artifact
         uses: actions/upload-artifact@v4
         with:
@@ -144,12 +144,11 @@ jobs:
         with:
           name: benchmark-report-${{ matrix.dialect }}
           path: ${{ matrix.dialect }}.json
-    
-          
+
   upload_benchmark_artifact:
     needs:
       - merge_benchmarks_into_single_report
-    
+
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,177 @@
+# This workflow is the main workflow for regenerating the benchmarks data needed for Bowtie's UI.
+# It runs all benchmarks over Bowtie's supported implementations, publishing the benchmark reports for use in the frontend.
+name: Collect New Benchmark Results
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every Monday at 00:00 UTC
+    - cron: "0 0 * * 1"
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  dialects:
+    runs-on: ubuntu-latest
+    outputs:
+      dialects: ${{ steps.dialects-matrix.outputs.dialects }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Collect supported dialects
+        id: dialects-matrix
+        run: |
+          printf 'dialects=%s\n' "$(jq -c '[.[].shortName]' data/dialects.json)" >> $GITHUB_OUTPUT
+
+  benchmark_files:
+    needs: dialects
+    runs-on: ubuntu-latest
+    outputs:
+      dialect_benchmarks: ${{ steps.benchmarks.outputs.dialect_benchmarks }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Bowtie
+        uses: ./
+
+      - name: Collect Benchmark Files
+        id: benchmarks
+        run: |
+          results=()
+          dialects='${{ needs.dialects.outputs.dialects }}'
+          dialects=$(echo $dialects | jq -r '.[]')
+          results=()
+
+          for dialect in $dialects; do
+            output=$(bowtie filter-benchmarks -D "$dialect")
+
+            if [ -n "$output" ]; then
+              while IFS= read -r line; do
+                json_result=$(jq -nc --arg p "$dialect" --arg o "$line" '{ dialect: $p, benchmark: $o }')
+                results+=("$json_result")
+              done <<< "$output"
+            fi
+          done
+          final_json="$(jq -sc '.' <<< "${results[@]}")"
+          final_json=$(echo "$final_json" | jq -c '{ "include": . }')
+          echo $final_json >> $GITHUB_STEP_SUMMARY
+          echo "dialect_benchmarks=$final_json" >> $GITHUB_OUTPUT
+
+  run_benchmarks:
+    needs: benchmark_files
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,175 @@
+# This workflow is the main workflow for regenerating the benchmarks data needed for Bowtie's UI.
+# It runs all benchmarks over Bowtie's supported implementations, publishing the benchmark reports for use in the frontend.
+name: Collect New Benchmark Results
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every Monday at 00:00 UTC
+    - cron: "0 0 * * 1"
```
Julian, line 9: Try to avoid running anything at midnight that really doesn't need to run then. Think about all the people on the planet who've chosen that time of day to run stuff thinking it's "arbitrary". So even though GitHub may try to prevent degradation at that time, perhaps knowing how many people have chosen it, there's still no reason to take the risk.

sudo-jarvis: Configured it to run at 8 in the morning

sudo-jarvis:
@Julian Looks like there is a vulnerability in the npm package `micromatch` because of which the audit test is failing .

Julian:
It looks like from https://github.com/advisories/GHSA-952p-6rrq-rcjv that the CVE is misdeclared (it considers 4.0.6 vulnerable) so we can't even downgrade while making it go away (so there's nothing to do until that's fixed in the database, which it looks like is in progress).

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,175 @@
+# This workflow is the main workflow for regenerating the benchmarks data needed for Bowtie's UI.
+# It runs all benchmarks over Bowtie's supported implementations, publishing the benchmark reports for use in the frontend.
+name: Collect New Benchmark Results
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every Monday at 08:00 UTC
+    - cron: "0 8 * * 1"
```
Julian, line 9: Running at 0 on the hour is not smart for similar reasons -- you need to imagine the distribution of jobs over the day -- it's going to be very high at midnight, and then have little bumps on the hour every hour, and then littler bumps every hour on multiples of 10 or 5, or etc.

But OK, we can leave this.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,175 @@
...
+  dialects:
+    runs-on: ubuntu-latest
+    outputs:
+      dialects: ${{ steps.dialects-matrix.outputs.dialects }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Collect supported dialects
+        id: dialects-matrix
+        run: |
+          printf 'dialects=%s\n' "$(jq -c '[.[].shortName]' data/dialects.json)" >> $GITHUB_OUTPUT
```
Julian, line 26: The reason we use this (already inscrutible) code is because in the CI job if we didn't (and used the alternative, `bowtie filter-dialects`) we would have to wait some amount of time to install bowtie and run it before doing useful work (running the suite).

Here, that seems anyways the case -- your next job waits for this one, and then the job after that waits for the middle one -- and the middle one installs bowtie and runs some inscrutible shell code.

Point being, if you anyways are doing some stuff installing bowtie before doing anything useful, you should just do this in one job and using bowtie's CLI.

(Ideally your entire shell script should be even doable using Bowtie's CLI in some trivial way without needing to write 12 lines of stuff, but that's possibly less trivial to fix.)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,175 @@
...
+          done
+          final_json="$(jq -sc '.' <<< "${results[@]}")"
+          final_json=$(echo "$final_json" | jq -c '{ "include": . }')
+          echo $final_json >> $GITHUB_STEP_SUMMARY
+          echo "dialect_benchmarks=$final_json" >> $GITHUB_OUTPUT
+
+  run_benchmarks:
+    needs: benchmark_files
+    runs-on: ubuntu-latest
+    timeout-minutes: 720
```
Julian, line 65: What's the point of this? GitHub will kill the job anyways.

.github/workflows/benchmark.yml
sudo-jarvis: Anyways our benchmarks are gonna require about 2.5 hrs from what I have tested, so thats why this shouldnt make much of a difference.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,175 @@
...
+      - name: Install pyperf dependency
+        run: |
+          python -m pip install pyperf
+
+      - name: Generate Benchmark Report
+        run: |
+          bowtie perf $(bowtie filter-implementations | sed 's/^/-i /') -b ${{ matrix.benchmark }} -D ${{ matrix.dialect }} -q > ${{ matrix.benchmark }}-file.json
+
+      - name: Store Unique Id for distinction
+        run: echo "unique_id=$(uuidgen)" >> $GITHUB_ENV
```
Julian, line 84: I'm not sure I understand distinction from what -- but surely github workflow runs already have some unique ID in their environment that you can use (and that presumably would match some value one can see in the UI).

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,175 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: benchmark-file-${{ matrix.dialect }}-*
+          path: benchmarks-${{ matrix.dialect }}/
+          merge-multiple: true
+
+      - name: Merge Benchmark Reports for a dialect
+        run: |
+          python - <<EOF
+          import os
```
Julian, line 116: This *definitely* seems like it should live in Bowtie and use Bowtie's benchmark types rather than some ad hoc code here.

.github/workflows/benchmark.yml
sudo-jarvis: Ah I was using this during testing to increase the timeout as the jobs were getting killed. But I later found out that 720 is the default timeout itself and that they were getting killed due to some other reason. Forgot to remove it. Will do.

.github/workflows/benchmark.yml
sudo-jarvis: Hmm you are right, here `bowtie filter-dialects` itself can be used.

.github/workflows/benchmark.yml
sudo-jarvis: I had thought about it but figured that maybe having a command like `bowtie merge-benchmarks` can do this but would be have any usecase for it other than this or not was the case. I decided since we arent probably gonna have such a usecase. It was better to do it here itself for the time being. 

What do you feel ? @Julian

Julian: It's possible -- but then it still would be better as some code living in Bowtie which used Bowtie's types (and had tests) than just ad hoc Python code which has no tests and no assurance it'll work if we change the output format ever.

.github/workflows/benchmark.yml
sudo-jarvis: I tried testing with this, it seems like `bowtie filter-dialects` supports only URI format in output. I am using this dialect name to  differentiate different dialect benchmarks as well. The URI contains characters which github doesnt allow to be present in an artifact name.

Currently I am using: `pattern: benchmark-file-${{ matrix.dialect }}-*` to filter all benchmark reports of a particular dialect.

Shall I continue using this or switch to using `bowtie filter-dialects` and then apply some string manipulation to the URI while naming the file to avoid this error ? @Julian

Julian: (If we do end up leaving as is, leave a comment then that changing this depends on #1332)

So wait do you not use this for anything else other than naming the artifact, which you then ignore and use just to collect the artifact? Or do you really use the short name itself somehow too?

If it's literally just to be able to upload multiple artifacts I think again you should have a look, surely there's a per-matrix identifier that github makes available for every subjob.

But if you don't find one or if you're using it somehow, then yeah leave as is. Thanks for checking.

sudo-jarvis: Its basically just helping group artifacts

Say we have:

benchmark-dialect2020-1.json
benchmark-dialect2020-2.json
benchmark-dialect2019-1.json
benchmark-dialect2019-2.json


The shortname is just being used to then filter the benchmark

sudo-jarvis: The shortname is also being used to get the benchmarks supported for each dialect but for that the URI works too so there is no issue in that.

sudo-jarvis: Also to keep the naming convention of benchmark reports similar to test-reports, the shortname is required.

sudo-jarvis: So ig it would be better to do this once #1332 is done.

Julian: > Also to keep the naming convention of benchmark reports similar to test-reports, the shortname is required.

That doesn't seem the same to me -- test reports are by their nature specific to a dialect, and compliance is always relative to a version of the spec.

A benchmark really isn't, the fact that dialects come into the picture at all is simply to know whether a particular dialect is supported so we know whether we can run a specific schema against an implementation, but I would prefer the notion of a dialect be nowhere near the real "structure" of benchmark reports and just be a kind of 'tag' like any other that we can maybe filter by at some point, but the main structure of benchmarks I think should just be "flat" if that makes sense in an imprecise sense.

.github/workflows/benchmark.yml
Julian: That seems reasonably OK, but if `combine_benchmark_reports` is already going to take paths (and not reports) then it seems reasonable that it returns a str and not a `BenchmarkReport` (and so it should just json.dumps + serializable its return value probably). Making it take an iterable and not a list also seems easily doable to avoid needing the `list` call.

.github/workflows/benchmark.yml
sudo-jarvis: Ig we had discussed this earlier that it would be great that at sometime we support say running benchmark over the same implementation and same schema but diff dialects and see the difference. So in a way benchmark reports are tied to dialects.

sudo-jarvis: Maybe once multi dialect supported within the same benchmark is added then this would be possible

Julian: OK, I'm good with whatever here to start!

sudo-jarvis: Yeah once that is supported , displaying it on the UI would also require some thinking so as to how to show those comparisons. Because now its easy and similar to tests that just show how dialect2020 test of `implementation1`  performs against other implementations on the same dialect. But when multi dialects are supported we would need to think on what would `implementation1` performing better than `implementation2` mean if its not tied to a particular dialect.  Would it mean on avg of all tests of all dialects or what. This would need to be thought of. Idk if I made much sense :(

Julian: Unless we have a good reason to be more complicated, I1 faster than I2 should mean that the benchmarks run by I1 are a superset of the benchmarks run by I2 (not necessarily a strict superset) and that the geometric average of runtime over that subset of benchmarks is faster in I1 than I2 -- but yeah again in my opinion with no reference to dialects whatsoever, you just flatten all of them and consider them individually.

sudo-jarvis: So currently in the frontend we have separate pages for each dialect, I should combine them all into one then ?

sudo-jarvis: Ok then ig we can let this be like this for now and once the above is supported we can flatten it all.

sudo-jarvis:
Hey @Julian , can you do a release for bowtie and then run the benchmarks workflow for the first time? Or else the reports workflow would fail as currently we have no artifact as this workflow has never been run.

PR #1459, opened 2024-08-16
Taking arbitrary params in Direct connectable strings

adwait-godbole:
@Julian I have probably failed miserably at this :( but yeah here’s something atleast.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1459.org.readthedocs.build/en/1459/

### DIFF
--- a/bowtie/_connectables.py
+++ b/bowtie/_connectables.py
@@ -122,8 +122,7 @@ class Connectable:
                 if len(args) == 1 and not kwargs:
                     connector = Connector(id=f"{id}:{args[0]}")
                 else:
-                    if "id" in kwargs:
-                        id = kwargs.pop("id")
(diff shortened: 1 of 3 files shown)


adwait-godbole:
The tests are failing because now everytime `make_connect` creates a new lambda callable so the the object addresses are not the same while comparing.

Julian:
This indeed turned out to be less straightforward than I implied, basically because to solve the problem you mentioned (about equality of lambda functions) requires inserting an additional type. You were certainly on the right track! So props for that. You can see what I did for this in 7ea8c3077275ad10b4d5ef47fe66230dd1a404e3 -- which still can be improved, but hopefully unblocks this. Thanks again for giving these kinds of things a real shot, well done!

PR #1461, opened 2024-08-17
UI changes to visualize Benchmark Reports

sudo-jarvis:
Here are some UI samples:

1. Benchmarking System Information
<p>
Screenshot 2024-08-17 at 11 12 03 PM
<p>
2. Added filtering on benchmark type as well
<p>
Screenshot 2024-08-17 at 11 12 07 PM
<p>
3. Chart visualizing the Benchmark Results
<p>
Screenshot 2024-08-17 at 11 12 09 PM
<p>
4. Each Benchmark present as a dropdown
<p>
Screenshot 2024-08-17 at 11 12 13 PM
<p>
5. Chart visualization per benchmark as well
<p>
Screenshot 2024-08-17 at 11 12 18 PM
<p>
6. Option to view detailed report as well for both types: same size variation as well as diff size
<p>
Screenshot 2024-08-17 at 11 12 20 PM
<p>
7. Detailed Report View in tabular form
<p>
Screenshot 2024-08-17 at 11 12 34 PM
<p>
8. Removed the chart from the mobile view as it wont be possible to visualize so much on such a small screen. The detailed results would be only present on the mobile screen.
<p>
Screenshot 2024-08-17 at 11 22 27 PM
<p>

https://github.com/user-attachments/assets/c4aaf9a3-b0b7-4041-b11f-410312b2dacb

<p>

----
📚 Documentation preview 📚: https://bowtie-json-schema--1461.org.readthedocs.build/en/1461/


sudo-jarvis:
@Julian , please have a look

frontend/src/components/BenchmarkResults/BenchmarkSummarySection.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const meanForImplementation: Record<string, number> = {};
+
+    Object.entries(resultsForImplementation).map(
+      ([implementationId, implementationResults]) => {
+        const means: number[] = implementationResults.map(
+          (implementationResult) => {
+            return mean(implementationResult);
+          },
+        );
+        const prodOfMeans = prod(means);
```
Julian, line 58: This can use your `geometricMean` helper right?

frontend/src/components/BenchmarkResults/BenchmarkTypeResultSection.tsx
```diff
@@ -0,0 +1,40 @@
...
+}) => {
+  return (
+    <ListGroup.Item
+      as="li"
+      className="d-flex justify-content-between align-items-start"
+    >
+      <div className="ms-2 me-auto">
+        <div className="fw-bold mb-4 mt-2 ">
+          Benchmark Type:{" "}
+          {benchmarkType.charAt(0).toUpperCase() + benchmarkType.slice(1)}
```
Julian, line 22: I think we do this in a few places, is there some helper lib we can use with a title function?

frontend/src/index.tsx
```diff
@@ -48,6 +49,22 @@ const dialectReportViewDataLoader = async ({
   return { reportData, allImplementationsData };
 };
 
+const benchmarkReportViewDataLoader = async ({
+  params,
+}: {
+  params: Params<string>;
+}) => {
+  const draftName = params?.draftName ?? "draft2020-12";
```
Julian, line 57: Rather than hardcoding `"draft2020-12"` see if we already have a `LATEST_DIALECT` constant somewhere, or if not possibly we should.

adwait-godbole: Not really my place to comment here but I did do a minor hackaround to fix this over [here](https://github.com/bowtie-json-schema/bowtie/blob/4ae905b04eb452eb3542e31ef2b1f6e6b9431392/frontend/src/index.tsx#L47) in the trend chart PR.

frontend/src/components/BenchmarkResults/BenchmarkTypeResultSection.tsx
sudo-jarvis: Will look for it

frontend/src/index.tsx
sudo-jarvis: Yeah I think we can have a Dialect.latest()

Julian:
I tried to merge this after fixing the conflicts, but then when going to the benchmarks page I get a

```

Unexpected Application Error!
A "java-networknt-json-schema-validator" implementation already exists.

ImplementationError@http://localhost:5173/src/data/Implementation.ts:134:5
Implementation@http://localhost:5173/src/data/Implementation.ts:24:13
parseBenchmarkMetadata/implementations<@http://localhost:5173/src/data/parseBenchmarkData.ts:21:32
parseBenchmarkMetadata@http://localhost:5173/src/data/parseBenchmarkData.ts:21:7
fromJSON@http://localhost:5173/src/data/parseBenchmarkData.ts:10:15
fetchBenchmarkReport@http://localhost:5173/src/data/Dialect.ts:46:12

💿 Hey developer 👋

You can provide a way better UX than this when your app throws errors by providing your own ErrorBoundary or errorElement prop on your route.
```

Any idea what that's about?

adwait-godbole:
> I tried to merge this after fixing the conflicts, but then when going to the benchmarks page I get a
> 
> ```
> 
> Unexpected Application Error!
> A "java-networknt-json-schema-validator" implementation already exists.
> 
> ImplementationError@http://localhost:5173/src/data/Implementation.ts:134:5
> Implementation@http://localhost:5173/src/data/Implementation.ts:24:13
> parseBenchmarkMetadata/implementations<@http://localhost:5173/src/data/parseBenchmarkData.ts:21:32
> parseBenchmarkMetadata@http://localhost:5173/src/data/parseBenchmarkData.ts:21:7
> fromJSON@http://localhost:5173/src/data/parseBenchmarkData.ts:10:15
> fetchBenchmarkReport@http://localhost:5173/src/data/Dialect.ts:46:12
> 
> 💿 Hey developer 👋
> 
> You can provide a way better UX than this when your app throws errors by providing your own ErrorBoundary or errorElement prop on your route.
> ```
> 
> Any idea what that's about?

@Julian I think you will have to blame me for this. I've added the below way of handling implementations after introducing the `Implementation` class.

https://github.com/bowtie-json-schema/bowtie/blob/1a33a04b5b13cd4095caf0398f9df095dc266f93/frontend/src/data/parseReportData.ts#L51

But I think it would be better if `new Implementation()` did not throw an error and just returned the already existing `Implementation` ig?

Julian:
Probably? The way we do dialects makes no sense really either, nothing should ever raise errors when constructing things, the only reason dialects work that way is because we have crazy global state there -- i.e. we do `new Dialect(...)` to add dialects to some list, rather than simply having an array of dialects we create as part of initializing the module. If we had that it would make no sense to complain about creating the same dialect twice either.

adwait-godbole:
@sudo-jarvis if you got any ideas over here then that would be great!

sudo-jarvis:
Ig to me too raising an Error while initializing a new Implementation object doesnt make sense. Better to just return the object if it already exists.

PR #1470, opened 2024-08-20
Backward compatibility for `java-json-schema`

adwait-godbole:
Hi @harrel56! Hope you're doing well. Needed a bit of your help over here.

To give you the context, this is a part of my [GSoC project](https://github.com/json-schema-org/community/issues/607). So I was trying to make the harness code of your implementation to be backward compatible so as to make it possible to build container images out of it, tagging them multiple releases and going as backward as possible in time with them and eventually all this gets reported via a **Bowtie** report on the official website via a trend graph for now.

Although I have tried tweaking some of your code over here, my first aim was to make sure that after making the changes, the results on the latest version which right now is 1.7.1 remain the same giving some sort of an assurance that atleast I've started on the right path. Though when I tried building an image for 1.7.1 out of this current scenario of the code, the results were same for `Draft2020-12` and `Draft2019-09` but something seems to be going wrong with `Draft7`. The results are coming out to be `Errors: 68, Failures: 16, Skips: 2`. The `Skips: 2` part was as expected but the other two are too much off from the current scenario. Can you please help me in figuring out what have I done wrong over here. Your help would be much appreciated.

Ohh and btw, the command I am using to build the Docker image locally on my machine right now looks like this

```bash
docker build --build-arg IMPLEMENTATION_VERSION=1.7.1 -t foo/java-json-schema:v1.7.1 .
```

Thanks.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1470.org.readthedocs.build/en/1470/


harrel56:
@adwait-godbole well, got to say this idea for a backward compatibility is very ambitious :D I'm gonna take a closer look at the code changes in a few days but in the meanwhile I have some questions:
1. What's the earliest version we want to check? ngl it'd be easiest to just start from 1.7.0 :D There probably is no good approach for handling arbitrarily old version - I guess the approach you took with magics of java reflection is the only one that could do the job, but it's extremely prone to errors and certainly will break in the future. And I'm not sure if I really want the maintenance burden here.
2. Is it actually worth it? From this implementation past experience there is not many changes in spec compliance. The only changes that happened are when support for draf2019-09 and draft7 were added. idk if such compliance data would make a nice chart or whatever it's gonna get presented as :)
3. Any other implementation already went through this process?

adwait-godbole:
Hi @harrel56, here are the answers to your questions above.

> 1. What's the earliest version we want to check? ngl it'd be easiest to just start from 1.7.0 :D There probably is no good approach for handling arbitrarily old version - I guess the approach you took with magics of java reflection is the only one that could do the job, but it's extremely prone to errors and certainly will break in the future. And I'm not sure if I really want the maintenance burden here.

I just added a `matrix-versions.json` file on this PR and have successfully managed to build images of your harness for all the versions that are listed in that JSON file. Please have a look. Also now the harness is working fine on all the dialects that a version is claiming to support and giving out correct results for the same.

> 2. Is it actually worth it? From this implementation past experience there is not many changes in spec compliance. The only changes that happened are when support for draf2019-09 and draft7 were added. idk if such compliance data would make a nice chart or whatever it's gonna get presented as :)

That's fine. Not an issue as such. Me and @Julian have discussed this earlier on as well and he is also fine even if the implementations don't really have drastic changes in the spec compliance as you're saying and nonetheless as part of my GSoC project, getting your implementation up on the website for showing the trend is just part of my job even if there's not really anything eye catching per se in fact there would just be a flat line on the trend graph in your case
(P.S. All credits to you for writing such a great implementation! :)).

> 3. Any other implementation already went through this process?

Yes! There's [ruby-json_schemer](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/ruby-json_schemer/bowtie_json_schemer.rb), [js-hyperjump](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/js-hyperjump/bowtie_hyperjump.js) and [python-jsonschema](https://github.com/bowtie-json-schema/bowtie/blob/main/implementations/python-jsonschema/bowtie_jsonschema.py) (Julian's implementation :)) that have already went through this process and you can even have a look at their trend graphs below (the PR for this is not merged yet, so it's not yet live on https://bowtie.report but will be soon enough) :-

https://github.com/user-attachments/assets/c7c4f190-d284-455c-a395-cbc6cb9cfa42

implementations/java-json-schema/BowtieJsonSchema.java
```diff
@@ -115,18 +138,55 @@ private void dialect(JsonNode node) throws JsonProcessingException {
...
+                                            .getClass()
+                                            .getMethod(
+                                              "withDefaultDialect",
+                                              Dialect.class
+                                            );
+      withDefaultDialectMethod.invoke(validatorFactory, dialectInstance);
+    } catch (NoSuchMethodException e) {
+        Method withDialectMethod = validatorFactory
+                                      .getClass()
+                                      .getMethod(
```
harrel56, line 171: I really dislike calling methods like this through reflection - for the `withDefaultDialect` it's necessary but here I think you can just call `validatorFactory.withDialect(...)` (this method was always present if I remember correctly)

implementations/java-json-schema/BowtieJsonSchema.java
```diff
@@ -115,18 +138,55 @@ private void dialect(JsonNode node) throws JsonProcessingException {
...
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to set dialect", e);
     }
+
     DialectResponse dialectResponse = new DialectResponse(true);
     output.println(objectMapper.writeValueAsString(dialectResponse));
   }
 
+  private void setDialect(Class<?> dialectClass) throws Exception {
+    Object dialectInstance = dialectClass.getDeclaredConstructor().newInstance();
```
harrel56, line 158: Maybe instead of calling `getDeclaredConstructor()` (which can return also non-public constructors and run into some security/access issues) you could call `getConstructor()` (which returns only public ctors, and these classes should have public ctors)

Julian:
"We" definitely aren't settled on whether to use one file or many to do this, so push back if you suspect this makes the harness uglier to maintain. In general there's two use cases to this: the main one Adwait showed off is we want to show for *all* implementations some notion of how they changed over time -- which yeah I think is nice even if it's flat! that's a good thing maybe! And the second is that it enables a user of Bowtie to answer questions about behavior they see for any version, even if in their codebase they're on an old one, but if they want to check whether they see is reproducible, they'd be able to (as well as check whether it's fixed by later versions). So if they're using some schema X and they see unexpected behavior, they can check whether the behavior they see is reproducible on various other versions without needing to futz with their own codebase.

This isn't a "huge" win -- so as I say feel free to push back on structure, it may be there's an even better way to do this (e.g. the obvious one is to have harnesses which don't simultaneously live in the Bowtie main branch codebase, but this makes rebuilding them if there's a bug that *isn't* related to the implementation harder). So yeah a lot isn't fully figured out here and could be improved on, but that's the general thinking so far!

implementations/java-json-schema/BowtieJsonSchema.java
```diff
@@ -115,18 +138,55 @@ private void dialect(JsonNode node) throws JsonProcessingException {
...
 
-    // well, setting default dialect shouldn't be necessary but why not
-    if (SpecificationVersion.DRAFT2020_12.getId().equals(dialectRequest.dialect())) {
-        validatorFactory.withDefaultDialect(new Dialects.Draft2020Dialect());
-    } else if (SpecificationVersion.DRAFT2019_09.getId().equals(dialectRequest.dialect())) {
-        validatorFactory.withDefaultDialect(new Dialects.Draft2019Dialect());
-    } else if (SpecificationVersion.DRAFT7.getId().equals(dialectRequest.dialect())) {
-      validatorFactory.withDefaultDialect(new Dialects.Draft7Dialect());
+    try {
+      if (getSpecificationVersionFor("DRAFT2020_12").getId().equals(dialectRequest.dialect())) {
```
harrel56, line 142: For this section I have a different idea :)
Maybe collect available dialects like this:
```java
Map<String, Dialect> dialectsMap = Arrays.stream(Dialects.class.getClasses())
            .filter(Dialect.class::isAssignableFrom)
            .map(Dialect.class::cast)
            .collect(Collectors.toMap(dialect -> dialect.getSpecificationVersion().getId(), Function.identity()));
```
store this map as a field and query it later for specific dialect:
```java
dialectsMap.get(dialectRequest.dialect())
```

harrel56: This way hopefully I won't have to edit the harness when I start supporting another draft version

implementations/java-json-schema/BowtieJsonSchema.java
```diff
@@ -32,6 +34,30 @@ public class BowtieJsonSchema {
   private final PrintStream output;
   private boolean started;
 
+  static {
+    MANIFEST_ATTRIBUTES = readManifestAttributes();
+    UNSUPPORTED =
+        "1.7.0".compareTo(
```
harrel56, line 40: Just use the first map:
```java 
Map.of("$ref prevents a sibling $id from changing the base uri",
                     RECOGNIZING_IDENTIFIERS)
```

The second one is taking into account an older version of test suite so it isn't really related to my impl version

harrel56 (commented):
Overall this looks very good, I just left a few comments that might help with maintenance in the future.

Followup question: if I release a new version how is the `matrix-versions.json` updated? Is it somehow handled by dependabot? Same question about that `gradle` snippet with a default version (I'm not sure if dependabot can handle bumping such version notation)

implementations/java-json-schema/BowtieJsonSchema.java
```diff
@@ -115,18 +120,42 @@ private void dialect(JsonNode node) throws JsonProcessingException {
...
 
-    // well, setting default dialect shouldn't be necessary but why not
-    if (SpecificationVersion.DRAFT2020_12.getId().equals(dialectRequest.dialect())) {
-        validatorFactory.withDefaultDialect(new Dialects.Draft2020Dialect());
-    } else if (SpecificationVersion.DRAFT2019_09.getId().equals(dialectRequest.dialect())) {
-        validatorFactory.withDefaultDialect(new Dialects.Draft2019Dialect());
-    } else if (SpecificationVersion.DRAFT7.getId().equals(dialectRequest.dialect())) {
-      validatorFactory.withDefaultDialect(new Dialects.Draft7Dialect());
+    ConcurrentHashMap<String, Dialect> dialectsMap = Arrays.stream(Dialects.class.getClasses())
+            .filter(Dialect.class::isAssignableFrom)
```
harrel56, line 124: I think this filter always evaluates to `false` - that's my bad (I also forgot about actually instantiating dialect objects but I see you figured it out already :). And it seems that filtering classes to be of a right type is a bit trick if you have to operate on `Class<?>` instances. So I would just get rid of the whole filtering, I don't think we gain much from it (worst case scenario it will throw exception and at least there will some indication what went wrong)

implementations/java-json-schema/BowtieJsonSchema.java
```diff
@@ -115,18 +120,42 @@ private void dialect(JsonNode node) throws JsonProcessingException {
...
+            .map(clazz -> {
+                try {
+                    return (Dialect) clazz.getConstructor().newInstance();
+                } catch (Exception e) {
+                    throw new IllegalStateException("Failed to instantiate Dialect", e);
+                }
+            })
+            .collect(
+              Collectors.toConcurrentMap(
```
harrel56, line 133: `ConcurrentHashMap` is not really necessary here:
- the whole harness code is sequential and single-threaded
- this is just a local variable in a method so even if there were multiple threads they'd cannot possibly share this map instance (each thread would create its own map instance)
- this map is created once and never changed, so it's effectively immutable. And immutable maps are thread-safe

And the other thing is that we can instantiate this map only once per harness lifetime. So I'd suggest making this map a final field in `BowtieJsonSchema` class and instantiate it in constructor (line 46). After this change it still wouldn't be necessary to change it to `ConcurrentHashMap`. But you can call `Collections.unmodifiableMap(...)` on a final result to be extra safe and actually enforce immutability :)

implementations/java-json-schema/BowtieJsonSchema.java
harrel56: > I think this filter always evaluates to false 

Well, turns out I don't know what I'm saying, it's all good, don't mind the whole previous comment :)

implementations/java-json-schema/BowtieJsonSchema.java
adwait-godbole: https://results.pre-commit.ci/run/github/531839193/1724347296.qqr9ylU_Tz-p-HGmDO30nQ

harrel56: Ah, so that's why you changed it. So I would just disable this rule in PMD because it doesn't make sense (this tool was really noisy from the beginning, reports a lot of false positives). I don't remember where the config was located though and cannot check it right now

adwait-godbole: > And the other thing is that we can instantiate this map only once per harness lifetime. So I'd suggest making this map a final field in BowtieJsonSchema class and instantiate it in constructor (line 46). After this change it still wouldn't be necessary to change it to ConcurrentHashMap. But you can call Collections.unmodifiableMap(...) on a final result to be extra safe and actually enforce immutability :)

The pre-commit error just went away after doing this :). Thanks!

harrel56 (approved):
No more comments, happy to merge. @Julian @adwait-godbole Still have one question pending - is dependabot able to publish new version to matrix. json or should it be updated manually?

adwait-godbole:
> No more comments, happy to merge. @Julian @adwait-godbole Still have one question pending - is dependabot able to publish new version to matrix. json or should it be updated manually?

@harrel56, Currently I am still working on figuring out how the `matrix-versions.json` should be updated whenever a `dependabot` update happens but regardless, I have updated this PR to **NOT** do any changes in the `build.gradle` file. Now, I am inserting the `IMPLEMENTATION_VERSION` from the Dockerfile itself using `sed` utility and replacing the version in `build.gradle` dynamically. So given this change, `dependabot` won't be confused  while updating your library version over there.

harrel56:
@adwait-godbole this sed looks pretty complicated - check out my approach: https://github.com/bowtie-json-schema/bowtie/commit/90b0c7bf414c1736a23472d0b99c2f0493032d49 dependabot handles it well

Julian:
> No more comments, happy to merge. @Julian @adwait-godbole Still have one question pending - is dependabot able to publish new version to matrix. json or should it be updated manually?

@harrel56 I think ideally we'd get rid of this file! The better way I think will be we literally use the tags we have in our own GitHub container registry -- in other words, the set of versions of an implementation that we support is simply "all the ones we built once".

And then all we have to solve is how to bootstrap that list the first time we add backwards compat for an implementation -- but that's a problem we don't automate even today.

Julian:
(And then I think with ^ everything "works" -- specifically dependabot will trigger *new* versions to get built, and old versions will simply stick around)

adwait-godbole:
Can you please trigger the `images.yml` workflow for `java-json-schema`? Thanks.

PR #1473, opened 2024-08-22
Add bowtie merge benchmarks, Remove duplicate benchmark

sudo-jarvis:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1473.org.readthedocs.build/en/1473/


bowtie/_benchmarks.py
```diff
@@ -105,6 +105,32 @@ def get_benchmark_files(
     return files
 
 
+def merge_benchmark_report_jsons(
```
Julian, line 108: You have too much of an affinity for dicts again it seems. Can't this be an operation on `Benchmark`s? Or `BenchmarkGroup`s, whichever is appropriate?

bowtie/_cli.py
```diff
@@ -1467,6 +1467,28 @@ def filter_benchmarks(
...
+@click.option(
+    "-r",
+    "--report",
+    "benchmark_report_files",
+    type=str,
+    multiple=True,
+    required=True,
+    help="Merge multiple benchmark reports into one.",
+)
+def merge_benchmark_reports(
```
Julian, line 1480: Didn't you say you thought we shouldn't add a CLI command because it didn't seem useful for anything else?

bowtie/_benchmarks.py
sudo-jarvis: I dont understand how we would do that, since to reduce time we are running each benchmark on separate jobs in the github workflow, it would create multiple reports which would be in json. So now when we want to merge them we have to load them in a dict right ? Thats why I did this

bowtie/_cli.py
sudo-jarvis: But then how else are you suggesting that we use this ? Just have a function in _benchmarks.py which we can directly call using the python shell in the CI workflow ?

bowtie/_benchmarks.py
Julian: Somewhere must have all the data, because somewhere you're running this code somewhere to do the merge -- so that place should be doing "load dicts into objects, call some method on that object which knows how to combine things correctly as Python objects" and then "write that result out to JSON", not "do arbitrary manipulation on dicts", essentially. But I admit I still don't have a mental picture of what the specific types are in that sentence, so let me know if I'm leading you astray.

bowtie/_cli.py
Julian: Sure, `python -c 'from bowtie._benchmarks import combine; combine(sys.stdin)` is way better than 15 lines of untested python code written into a workflow!

bowtie/_benchmarks.py
sudo-jarvis: Okay so if I get you right:

In the workflow itself we can use the python shell to load the report into a `BenchmarkReport` using the from_dict method.

Now we simply call the `merge_benchmark_reports` function which takes objects as input and merges them to again produce an object.

We can then serialize this object and write it to json then.

bowtie/_cli.py
sudo-jarvis: Got it, makes sense. Sorry for making things unnecessarily complicated.

bowtie/_benchmarks.py
Julian: That sounds like it has the right shape at least, yeah. Possibly as `BenchmarkReport.merge(*paths).serialize()`

bowtie/_cli.py
Julian: No no worries, I was terse!

bowtie/_benchmarks.py
sudo-jarvis: > BenchmarkReport.merge(*paths).serializable()

BenchmarkReport.merge(*report_objects) right ?

Julian: Yeah -- but some place has to be the place that actually *does* load from paths, so yeah I think that's right, at this layer take report objects, and then I guess in your `combine()` function you load paths into reports, yeah.

sudo-jarvis: But then if we have the combine function as well within bowtie itself then wouldnt it become the same thing as above. combine() + merge() would basically be doing the same thing, no?

sudo-jarvis: And here I am thinking, the combine and merge function would exist like this:

```
def combine_benchmark_reports(
    benchmark_report_files: list[Path],
) -> BenchmarkReport:
    return BenchmarkReport.merge([
        BenchmarkReport.from_dict(
            json.loads(report_file.read_text())
        )
        for report_file in benchmark_report_files
    ])
```
    
```
@classmethod
    def merge(
        cls, benchmark_reports: list[BenchmarkReport]
    ) -> BenchmarkReport:
        return BenchmarkReport(
            metadata=benchmark_reports[0].metadata,
            results={
                k: v for report in benchmark_reports
                for k, v in report.results.items()
            },
        )
```

Julian: combine would call merge with the loaded paths, so it would just be your entry point, but that operation is now testable without file paths.

Julian: That looks right -- though yeah presumably at some point there should be some validation that the metadata is mutually compatible, and not just arbitrarily take the first one.

sudo-jarvis: Thats true, but yeah then ig this looks right for an inital version.

Julian: Yep.

sudo-jarvis:
Strangely precommit is showing errors even after this:

`Found 2 errors (2 fixed, 0 remaining).`

Julian:
The issues are real, and somehow pre-commit.ci is confused (presumably this is a known bug). But it's just autofixing them that's not working. Let's see if this fixes it -- otherwise you'll have to fix them by running ruff yourself.

Julian:
pre-commit.ci autofix

sudo-jarvis:
@Julian , the test looks fine or should I write a more exact one comparing all the fields

bowtie/_benchmarks.py
```diff
@@ -420,6 +475,26 @@ def serializable(self):
         )
         return as_dict
 
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> BenchmarkMetadata:
+        return cls(
```
Julian, line 480: This is doing too much manual `dict` access -- it should be doing `**data` (so that extra fields cause errors), and then the only thing it should do is coerce types of things in `data` which don't start as the right type.

bowtie/_benchmarks.py
```diff
@@ -420,6 +475,26 @@ def serializable(self):
...
+                for id, info in data["implementations"].items()
+            },
+            num_runs=data["num_runs"],
+            num_values=data["num_values"],
+            num_warmups=data["num_warmups"],
+            num_loops=data["num_loops"],
+            system_metadata=data.get("system_metadata", {}),
+            bowtie_version=data.get(
+                "bowtie_version",
+                importlib.metadata.version("bowtie-json-schema"),
```
Julian, line 493: This default also doesn't seem like it belongs here.

bowtie/_benchmarks.py
```diff
@@ -436,6 +511,33 @@ def serializable(self):
...
 
+    @classmethod
+    def from_dict(
+        cls,
+        data: dict[str, Any],
+    ) -> BenchmarkReport:
+        return cls(
+            metadata=BenchmarkMetadata.from_dict(data["metadata"]),
+            results={
+                r.get("name"): BenchmarkGroupResult.from_dict(r)
```
Julian, line 522: Why would `name` not be present? `None` is going to be a really poor key for this dict.

bowtie/tests/test_benchmarks.py
```diff
@@ -53,6 +53,26 @@ def invalid_benchmark():
     return invalid_benchmark.get_benchmark()
 
 
+@pytest.fixture()
+def benchmark_report1():
```
Julian, line 57: Fixtures should be for "obviously" reusable things. Here (and in the existing tests) you basically create fragility more than anything -- because it's not clear what someone could use these fixtures for *other* than the very specific test you wrote for merging -- that means you should just do this inline rather than making it "seem" reusable.

bowtie/_benchmarks.py
sudo-jarvis: `name` would be present always. I just used get here. I'll change it to r["name"].

bowtie/_benchmarks.py
```diff
@@ -105,6 +105,20 @@ def get_benchmark_files(
     return files
 
 
+def combine_benchmark_reports_to_serialized(
+    benchmark_report_files: Iterable[Path],
+) -> str:
+    merged_report = BenchmarkReport.merge(
+        [
```
Julian, line 112: Seems like `BenchmarkReport.merge` should really have taken an iterable too.

Julian:
OK let's merge for now this looks good enough, we can come back for some small cleanups. Thanks again!

PR #1476, opened 2024-08-25
Bug Fix: Somehow the former gives error in terms of some url library missing during the workflow

sudo-jarvis:
I had tested the workflow using, since the latest benchmark changes were not there in the latest release:

This works fine
```
     - name: Set up Python
        uses: actions/setup-python@v4
        with:
          python-version: '3.12'

      - name: Install Bowtie
         run: |
          pip install -e .
```

However now using, 

This doesnt work
```
- name: Install Bowtie
        uses: ./
```

Since the latest release has been done, the workflow is failing on the python script step saying `url` library not found. I dont understand why though @Julian . Both code should work the same functionality wise right ?

ERROR:

```
Run python - <<EOF
  python - <<EOF
  from pathlib import Path
  from bowtie._benchmarks import combine_benchmark_reports_to_serialized as combine
  
  Path("draft[2](https://github.com/sudo-jarvis/bowtie/actions/runs/10547927252/job/29221319098#step:5:2)020-12.json").write_text(
      combine(
          Path("benchmarks-draft2020-12").iterdir()
      )
  )
  
  EOF
  shell: /usr/bin/bash -e {0}
  
  File "<stdin>", line 2, in <module>
  File "/home/runner/work/bowtie/bowtie/bowtie/__init__.py", line [5](https://github.com/sudo-jarvis/bowtie/actions/runs/10547927252/job/29221319098#step:5:5), in <module>
    from url import URL
ModuleNotFoundError: No module named 'url'
Error: Process completed with exit code 1.
```

Failed job link: https://github.com/sudo-jarvis/bowtie/actions/runs/10547927252/job/29221319098#step:5:19

----
📚 Documentation preview 📚: https://bowtie-json-schema--1476.org.readthedocs.build/en/1476/

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -109,7 +109,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.12'
+          python-version: "3.12"
 
       - name: Install Bowtie
         run: |
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -106,8 +106,14 @@ jobs:
           path: benchmarks-${{ matrix.dialect }}/
           merge-multiple: true
 
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.12'
+
       - name: Install Bowtie
-        uses: ./
+        run: |
+          pip install -e .
 
       - name: Merge Benchmark Reports for a dialect
         run: |
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -107,13 +107,16 @@ jobs:
           merge-multiple: true
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: "3.12"
 
+      - name: Set up uv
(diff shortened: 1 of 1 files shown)


Julian:
You need to use the `python` executable from the virtualenv bowtie was installed into, which is why it doesn't work. We (in the action) happen to install using pipx -- so this is quite hacky -- but we're doing internal things here.
The fix here isn't right, if you're going to install the package yourself you should just not use the action to install Bowtie -- but probably for now given we accepted the temporary hacky python script inlined we should possibly just check if pipx has a way to tell us where the virtualenv is / where that python lives, and you run using that one.

sudo-jarvis:
Hmm this worked @Julian . Had to use --sytem flag or else it was giving error that venv not found.

```
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: "3.12"

      - name: Set up uv
        uses: hynek/setup-cached-uv@v2

      - name: Install Bowtie
        run: | 
          uv pip install --system .
```

sudo-jarvis:
I'll push this then.

sudo-jarvis:
Done & tested it @Julian.

sudo-jarvis:
@Julian, sorry for bothering but please cancel the old workflow run and rerun it again

Julian:
Done!

PR #1487, opened 2024-08-29
Add `--versions` option to `bowtie info` for fetching versions from ghcr registry

adwait-godbole:
Screenshot (184)

@Julian I tried getting this to work.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1487.org.readthedocs.build/en/1487/

### DIFF
--- a/bowtie/__init__.py
+++ b/bowtie/__init__.py
@@ -10,3 +10,8 @@ DOCS = URL.parse("https://docs.bowtie.report/")
 GITHUB = URL.parse("https://github.com/")
 ORG = GITHUB / "bowtie-json-schema"
 REPO = ORG / "bowtie"
+
+GITHUB_API = URL.parse("https://api.github.com/")
+ORG_API = GITHUB_API / "orgs" / "bowtie-json-schema"
+PACKAGES_API = ORG_API / "packages"
+CONTAINER_PACKAGES_API = PACKAGES_API / "container"
--- a/bowtie/__init__.py
+++ b/bowtie/__init__.py
@@ -10,8 +10,3 @@ DOCS = URL.parse("https://docs.bowtie.report/")
 GITHUB = URL.parse("https://github.com/")
 ORG = GITHUB / "bowtie-json-schema"
 REPO = ORG / "bowtie"
-
-GITHUB_API = URL.parse("https://api.github.com/")
-ORG_API = GITHUB_API / "orgs" / "bowtie-json-schema"
-PACKAGES_API = ORG_API / "packages"
-CONTAINER_PACKAGES_API = PACKAGES_API / "container"
--- a/bowtie/__init__.py
+++ b/bowtie/__init__.py
@@ -8,5 +8,6 @@ HOMEPAGE = URL.parse("https://bowtie.report/")
 DOCS = URL.parse("https://docs.bowtie.report/")
 
 GITHUB = URL.parse("https://github.com/")
-ORG = GITHUB / "bowtie-json-schema"
+ORG_NAME = "bowtie-json-schema"
+ORG = GITHUB / ORG_NAME
 REPO = ORG / "bowtie"
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1641,7 +1641,9 @@ def latest_report(dialect: Dialect):
     "--versions",
     "show_versions",
     is_flag=True,
-    help="Show all versions of an implementation that its harness can build.",
+    help=(
+        "Also show all versions of this implementation which Bowtie supports."
+    ),
 )
 async def info(
     start: Starter,
@@ -1654,18 +1656,16 @@ async def info(
     serializable: dict[ConnectableId, dict[str, Any]] = {}
 
     async for _, each in start():
-        metadata = [
-            (k, v)
-            for k, v in {
-                **each.info.serializable(),
-                **(
-                    {"versions": list(await each.get_versions())}
-                    if show_versions
-                    else {}
+        metadata = [(k, v) for k, v in each.info.serializable().items() if v]
+        if show_versions:
+            versions = await each.get_versions()
+            metadata.append(
+                (
+                    "versions",
+                    list(versions) if versions else [each.info.version],
                 ),
-            }.items()
-            if v
-        ]
+            )
+
         metadata.sort(
             key=lambda kv: (
                 kv[0] != "name",
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1660,7 +1660,8 @@ async def info(
                 **each.info.serializable(),
                 **(
                     {"versions": list(await each.get_versions())}
-                    if show_versions else {}
+                    if show_versions
+                    else {}
                 ),
             }.items()
             if v
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1637,9 +1637,16 @@ def latest_report(dialect: Dialect):
 
 @implementation_subcommand()  # type: ignore[reportArgumentType]
 @format_option()
+@click.option(
+    "--versions",
+    "show_versions",
+    is_flag=True,
+    help="Show all versions of an implementation that its harness can build.",
+)
 async def info(
     start: Starter,
     format: _F,
+    show_versions: bool,
 ):
     """
     Show information about a supported implementation.
@@ -1647,13 +1654,24 @@ async def info(
     serializable: dict[ConnectableId, dict[str, Any]] = {}
 
     async for _, each in start():
-        metadata = [(k, v) for k, v in each.info.serializable().items() if v]
+        metadata = [
+            (k, v)
+            for k, v in {
+                **each.info.serializable(),
+                **(
+                    {"versions": list(await each.get_versions())}
+                    if show_versions else {}
+                ),
+            }.items()
+            if v
+        ]
         metadata.sort(
             key=lambda kv: (
                 kv[0] != "name",
                 kv[0] != "language",
                 kv[0] != "version",
                 kv[0] == "links",
+                kv[0] == "versions",
                 kv[0] == "dialects",
                 kv[0],
             ),
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1660,8 +1660,7 @@ async def info(
                 **each.info.serializable(),
                 **(
                     {"versions": list(await each.get_versions())}
-                    if show_versions
-                    else {}
+                    if show_versions else {}
                 ),
             }.items()
             if v
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1658,13 +1658,7 @@ async def info(
     async for _, each in start():
         metadata = [(k, v) for k, v in each.info.serializable().items() if v]
         if show_versions:
-            versions = await each.get_versions()
-            metadata.append(
-                (
-                    "versions",
-                    list(versions) if versions else [each.info.version],
-                ),
-            )
+            metadata.append(("versions", list(await each.get_versions())))
 
         metadata.sort(
             key=lambda kv: (
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1658,7 +1658,9 @@ async def info(
     async for _, each in start():
         metadata = [(k, v) for k, v in each.info.serializable().items() if v]
         if show_versions:
-            metadata.append(("versions", list(await each.get_versions())))
+            metadata.append(
+                ("versions", next(iter(await each.get_versions()))),
+            )
 
         metadata.sort(
             key=lambda kv: (
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1638,12 +1638,9 @@ def latest_report(dialect: Dialect):
 @implementation_subcommand()  # type: ignore[reportArgumentType]
 @format_option()
 @click.option(
-    "--versions",
+    "--versions/--no-versions",
     "show_versions",
-    is_flag=True,
-    help=(
-        "Also show all versions of this implementation which Bowtie supports."
-    ),
+    help="Also show all implementation versions which Bowtie supports.",
 )
 async def info(
     start: Starter,
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1658,9 +1658,7 @@ async def info(
     async for _, each in start():
         metadata = [(k, v) for k, v in each.info.serializable().items() if v]
         if show_versions:
-            metadata.append(
-                ("versions", next(iter(await each.get_versions()))),
-            )
+            metadata.append(("versions", await each.get_versions()))
 
         metadata.sort(
             key=lambda kv: (
--- a/bowtie/_core.py
+++ b/bowtie/_core.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager, suppress
 from datetime import date
 from functools import cache
 from importlib.resources import files
@@ -555,10 +555,8 @@ class Implementation:
 
         gh = _github()
         pages: list[GitHubCore] = []
-        try:
+        with suppress(GitHubError):
             pages = gh._iter(count=-1, url=str(url), cls=GitHubCore)  # type: ignore[reportPrivateUsage]
-        except GitHubError:
-            pass
 
         versions: Set[str] = (
             {self.info.version} if self.info.version else set()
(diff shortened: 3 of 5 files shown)


bowtie/__init__.py
```diff
@@ -10,3 +10,8 @@
 GITHUB = URL.parse("https://github.com/")
 ORG = GITHUB / "bowtie-json-schema"
 REPO = ORG / "bowtie"
+
+GITHUB_API = URL.parse("https://api.github.com/")
```
Julian, line 14: These can go in the CLI module I think, they're unlikely to be used anywhere outside it (whereas the rest of the stuff here are used in multiple places).

Also it's fine to just have one variable, I don't think anything but `CONTAINER_PACKAGES_API` is likely to be used anywhere either.

bowtie/_cli.py
```diff
@@ -1637,23 +1637,42 @@ async def write(response: Awaitable[Response]):
 
 @implementation_subcommand()  # type: ignore[reportArgumentType]
 @format_option()
+@click.option(
+    "--versions",
+    "show_versions",
+    is_flag=True,
+    help="Show all versions of an implementation that its harness can build.",
```
Julian, line 1644:
```suggestion
    help="Also show all versions of this implementation which Bowtie supports.",
```

bowtie/_cli.py
```diff
@@ -1637,23 +1637,42 @@ async def write(response: Awaitable[Response]):
...
+    show_versions: bool,
 ):
     """
     Show information about a supported implementation.
     """
     serializable: dict[ConnectableId, dict[str, Any]] = {}
 
     async for _, each in start():
-        metadata = [(k, v) for k, v in each.info.serializable().items() if v]
+        metadata = [
```
Julian, line 1657: Just leave this as is and do:

```
if show_versions:
  metadata.append((...))
```

bowtie/_core.py
```diff
@@ -538,6 +544,74 @@ def supports(self, *dialects: Dialect) -> bool:
...
+                    -int(part) if part.isdigit() else part
+                    for part in cast(str, tag).split(".")
+                ]
+            ),
+        )
+
+        try:
+            gh = GitHub(token=os.environ.get("GITHUB_TOKEN", ""))
+
+            class _TagsWrapper(GitHubCore):
```
Julian, line 575: This is a bunch of wrapping and unwrapping, I'm not sure it does much good -- does this not work if you try simply using `GitHubCore._iter()` -- i.e. not really defining anything new, just calling that method with your parameters, which it looks like already returns a `GitHubIterator`?

bowtie/__init__.py
adwait-godbole: > These can go in the CLI module I think

Meaning should I put them in `_core.py` itself? since I have added the `get_versions` method on `Implementation` itself.

bowtie/_core.py
```diff
@@ -538,6 +544,74 @@ def supports(self, *dialects: Dialect) -> bool:
...
+        )
+
+        url = CONTAINER_PACKAGES_API / self.id / "versions"
+        versions = SortedSet(
+            key=lambda tag: (  # type: ignore[reportUnknownLambdaType]
+                [
+                    -int(part) if part.isdigit() else part
+                    for part in cast(str, tag).split(".")
+                ]
+            ),
```
adwait-godbole, line 569: Also @Julian, can you point me somewhere where I can put the above as a `util` itself since this logic is repeated over [here](https://github.com/bowtie-json-schema/bowtie/blob/02c8ea066a2b7931fd97b55ea34c148ad40c5dc8/bowtie/_report.py#L381) as well.

bowtie/__init__.py
Julian: Yes, there is fine too.

bowtie/_core.py
Julian: I guess here in `_core` considering it's higher up than _report is (and the latter will anyhow import this here).

adwait-godbole:
@Julian I have tried to fix some of your suggested changes. Please have a look and let me know. Meanwhile, the discussion we had regarding adding test for this functionality, I later recalled that, is it possible to really add a test for this given that our dummy implementation (`envsonschema`) is not really published to any registry in real life. Or are we just mocking the GitHub API call so it doesn't really matter if `envsonschema` is published or not?

adwait-godbole:
> @Julian I have tried to fix some of your suggested changes. Please have a look and let me know. Meanwhile, the discussion we had regarding adding a test for this functionality, I later recalled, is it possible to really add a test for this given that our dummy implementation (envsonschema) is not really published to any registry in real life. Or are we just mocking the GitHub API call so it doesn't really matter if envsonschema is published or not?

Also, since I've added the `get_versions` method on `Implementation` itself, will that pose as a problem for writing the test?

bowtie/_core.py
```diff
@@ -538,6 +543,47 @@ def supports(self, *dialects: Dialect) -> bool:
...
 
+    async def get_versions(self):
+        from github3 import GitHub  # type: ignore[reportMissingTypeStubs]
+        from github3.exceptions import (  # type: ignore[reportMissingTypeStubs]
+            GitHubError,
+        )
+        from github3.models import (  # type: ignore[reportMissingTypeStubs]
+            GitHubCore,
+        )
+        from sortedcontainers import (  # type: ignore[reportMissingTypeStubs]
```
Julian, line 554: You're adding a new dependency here, but haven't add it to the requirements in `pyproject.toml`.

bowtie/_core.py
```diff
@@ -538,6 +543,47 @@ def supports(self, *dialects: Dialect) -> bool:
...
+            SortedSet,
+        )
+
+        url = CONTAINER_PACKAGES_API / self.id / "versions"
+        versions = SortedSet(key=version_key(reverse=True))
+
+        try:
```
Julian, line 561: A try block should generally have 1 line in it. so you catch only the exception on the line you intend to. Very rarely it's ok for it to have more, but veeeery rarely.

This seems definitely possible to reduce.

bowtie/_cli.py
```diff
@@ -1648,12 +1657,22 @@ async def info(
 
     async for _, each in start():
         metadata = [(k, v) for k, v in each.info.serializable().items() if v]
+        if show_versions:
+            versions = await each.get_versions()
+            metadata.append(
+                (
+                    "versions",
+                    list(versions) if versions else [each.info.version],
```
Julian, line 1665: This fallback logic (to return just the one version) seems like it belongs in `get_versions`?

bowtie/_core.py
Julian: I'm not sure it's even super important for this to be sorted by this method, you probably could just not use this library and return a normal set, and sort it when/if you want to print it.

bowtie/_core.py
```diff
@@ -803,3 +849,20 @@ def convert_table_to_markdown(
     body = "\n".join(body)
 
     return f"\n{header}\n{separator}\n{body}"
+
+
+def version_key(reverse: bool):
```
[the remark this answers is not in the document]
adwait-godbole: Fixed this!

bowtie/_cli.py
adwait-godbole: Fixed this!

adwait-godbole:
@Julian I tried fixing all your suggested changes. Please have a look and let me know. Thanks!

bowtie/_core.py
```diff
@@ -538,6 +543,43 @@ def supports(self, *dialects: Dialect) -> bool:
...
+                continue
+            else:
+                versions.update([tag for tag in tags if "." in tag])
+
+        if not versions and self.info.version:
+            versions.add(self.info.version)
+
+        return frozenset(
+            sorted(versions, key=sortable_version_key, reverse=True),
+        )
```
adwait-godbole, line 581: Also @Julian I am facing an issue over here in these lines above, when we try to do `frozenset(sorted(...))`, th e sorted order is not maintained after converting it to a `frozenset`. Is there a way of fixing this?

Julian: `frozenset` indeed isn't ordered. You have to pick one -- either return a set and let the person using this sort, or just return the iterable.

adwait-godbole: I tried a bit of a hack by following a discussion that I got on Stack Overflow and the below way seemed to work and also maintained the sorted order while returning a `frozenset`

```python
return frozenset(
    zip(
        *zip(sorted(versions, key=sortable_version_key, reverse=True)),
    ),
)
```

just that now I have to do 

```python
metadata.append(("versions", next(iter(await each.get_versions()))))
```

adwait-godbole: So now `get_versions` is returning the below

```python
async def get_versions(self) -> frozenset[zip[tuple[str, ...]]]
```

adwait-godbole:
Screenshot (185)


Also @Julian can you check if the above is enabled for the `GITHUB_TOKEN` settings of Bowtie? Otherwise it gives a `403` saying that read:packages scope is not enabled on this token.

bowtie/_core.py
Julian: I don't understand why you need/want to do this though. Why do you need both here?

Julian: That kind of StackOverflow answer is absolutely horrid. If you wanted both, use the dependency you found.

Julian:
> Also @Julian can you check if the above is enabled for the GITHUB_TOKEN settings of Bowtie? Otherwise it gives the following error 403 You need at least read:packages scope to get a package's versions.

That's controlled by the `packages: read` permission in any workflow where you want to use this I believe. GITHUB_TOKEN is just the default token provisioned by GitHub Actions.

bowtie/_core.py
```diff
@@ -538,6 +543,45 @@ def supports(self, *dialects: Dialect) -> bool:
...
+                tags = cast(
+                    Iterable[str],
+                    page.as_dict()["metadata"]["container"]["tags"],
+                )
+            except KeyError:
+                continue
+            else:
+                versions.update([tag for tag in tags if "." in tag])
+
+        if not versions and self.info.version:
```
Julian, line 576: This seems like we should do it always, rather than just if we didn't get a response -- if the current container is for a version not in the list, obviously that version still exists right :)?

bowtie/_core.py
adwait-godbole: > frozenset indeed isn't ordered. You have to pick one -- either return a set and let the person using this sort, or just return the iterable.

Okay I guess I will go with just returning a set then. Using `frozenset` with `SortedSet` also doesn't seem to be maintaing the order so no point in adding that dependency either.

adwait-godbole:
Made the fix. Can you have a look again? Sorry for the trouble.

adwait-godbole:
@Julian any updates over here?

Julian:
Does this work for you?

Here it doesn't seem to, I get:

```
⊙  bowtie info -i python-jsonschema --versions                                                                                                                                                                                                                                                                    julian@Airm ●
  File "/Users/julian/.local/bin/bowtie", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/rich_click/rich_command.py", line 367, in __call__
    return super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/rich_click/rich_command.py", line 152, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 268, in run
    exit_code = fn(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 379, in cmd
    return asyncio.run(run(connectables=connectables, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 350, in run
    fn_exit_code = await fn(start=start, **kw)  # type: ignore[reportArgumentType]
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/bowtie/_cli.py", line 1661, in info
    metadata.append(("versions", await each.get_versions()))
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/Development/bowtie/bowtie/_core.py", line 566, in get_versions
    for page in pages:
                ^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/github3/structs.py", line 98, in __iter__
    json = self._get_json(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/github3/structs.py", line 145, in _get_json
    return self._json(response, 200)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.12/site-packages/github3/models.py", line 161, in _json
    raise exceptions.error_for(response)
github3.exceptions.AuthenticationFailed: 401 Requires authentication
```

Julian:
Oh, the issue is an old GitHub API issue I filed that still exists unfortunately, namely that the packages API isn't readable publicly :/ even when an org is public. It's a really silly API bug but I guess no one else really cares about it at the minute.

But that's pretty poor behavior at the CLI -- essentially authentication is required for that CLI option. We need to at least have some error message indicating that's the case, and to have that present in the help message. I've pushed another few tiny tweaks but yeah please add some better error handling for when someone uses this at the CLI and then we can merge.

Julian:
It should immediately error out if you pass `--versions` and aren't authenticated without starting a container I think.

adwait-godbole:
> It should immediately error out if you pass --versions and aren't authenticated without starting a container I think.

Which means we should likely try to log in to `GitHub` even before calling `each.get_versions()` and possibly pass it the GitHub client that logged in. But should I do this in `_cli.py` and pass the client down to that method?

Julian:
Hm. Then you could catch it when it happens and just present a better error message, that's also fine if it's not easy to exit immediately.

adwait-godbole:
> Hm. Then you could catch it when it happens and just present a better error message, that's also fine if it's not easy to exit immediately.

@Julian I was yet to fix this :(. But I see that you merged already.

Julian:
It's fine to PR the improvement as a follow-up!

PR #1512, opened 2024-09-10
Add PHP justinrainbow/json-schema implementation

DannyvdSluijs:
This PR adds an implementation for [justinrainbow/json-schema](https://github.com/jsonrainbow/json-schema) (written in PHP).

The PR was drafted using the [instructions](https://docs.bowtie.report/en/stable/implementers/) if anything is missing or overlooked please feel free to comment.

I'm currently one of the maintainers of `justinrainbow/json-schema` which we are reviving. First order of business is getting the library back into an actively maintained state. Which is one of the main reasons of opening this PR.


Julian:
Very nice, thanks for this! Will do a release later tonight or tomorrow. 

Good luck with the implementation (and stay in touch)!

implementations/php-justinrainbow-json-schema/composer.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "justinrainbow/bowtie-json-schema-test-harness",
+  "description": "These sources contains the test harness implementation for Bowtie",
+  "license": "MIT",
+  "require": {
+    "justinrainbow/json-schema": "dev-master"
```
Julian, line 6: If perchance you're interested in a follow up PR, changing this to instead be a fixed version, (perhaps as soon as you publish one if that's not yet happened?) would be nice, and then relying on Dependabot to update it.

As is, right now the version information shown at https://bowtie.report/#/implementations/php-justinrainbow-json-schema (which is live, congrats!) won't be useful to someone, since it just says `dev-master`, but someone won't know what commit that corresponds to even if it was HEAD at some point in time.

PR #1593, opened 2024-10-15
Rename JSON Toolkit to Blaze

jviotti:
We extracted the JSON Schema validator out of the JSON Toolkit project into a project of its own, called "Blaze", which depends on JSON Toolkit.

See: https://github.com/sourcemeta-research/jsonschema-benchmark/pull/67

----
📚 Documentation preview 📚: https://bowtie-json-schema--1593.org.readthedocs.build/en/1593/


jviotti:
@Julian Looks like CI is still trying to build `cpp-jsontoolkit`?

Julian:
Nice! Congrats. I think your build has been failing though, I meant to mention it to you -- I think (but haven't looked) that that's happening on this build as well, can you have a look?

(This is another reason possibly to not allow implementations to use `master`/`main` here as their build target, as it becomes a burden here to diagnose a moving target, but perhaps that's something worth deciding "formally" on later on if this happens consistently with implementations.)

jviotti:
@Julian Where do you see the build error? I only see this error, due to CI not picking up the rename?

Screenshot 2024-10-16 at 10 29 44 am

> Nice! Congrats. I think your build has been failing though, I meant to mention it to you -- I think (but haven't looked) that that's happening on this build as well, can you have a look?

Yeah, we plan to start releasing more stable versions soon. Right now, because of all of the work going in, it feels we making breaking changes every day in between 😅 Is there a way implementation authors can at least be notified of build failures? Other than that, pinning to specific commit hashes probably works for now, until we start releasing more stable versions.

Julian:
(The rename bug obviously does seem to be a bug, but I don't know offhand why it's happening given I think we do use the `HEAD` commit to list implementations. If you see something let me know, otherwise I'll try to make some time to have a look, it should be something simple.)

> @Julian Where do you see the build error? I only see this error, due to CI not picking up the rename?

Here's the runs from the last 2 Bowtie releases:

https://github.com/bowtie-json-schema/bowtie/actions/runs/11309482319/job/31453388548
https://github.com/bowtie-json-schema/bowtie/actions/runs/11205331918/job/31145241357

> Is there a way implementation authors can at least be notified of build failures? 

I didn't want to do this "without asking" but perhaps a good idea is for any implementers who want them we could auto-open issues on their repos saying "building your image failed here". I can't remember the last time it was badly "Bowtie's fault" there on the actual build being done incorrectly, though of course it's possible (and likely) it'll happen at some point, but at least if an implementer is interested perhaps that would be a good way of passing the notification along... Thoughts definitely welcome!

jviotti:
Awesome thanks! Let me tweak my implementation to pin the commit for now and fix the current errors.

> I didn't want to do this "without asking" but perhaps a good idea is for any implementers who want them we could auto-open issues on their repos saying "building your image failed here". I can't remember the last time it was badly "Bowtie's fault" there on the actual build being done incorrectly, though of course it's possible (and likely) it'll happen at some point, but at least if an implementer is interested perhaps that would be a good way of passing the notification along... Thoughts definitely welcome!

Auto-opening issues, or even pinging on JSON Schema Slack sounds great to me.

jviotti:
@Julian This seems to work for me. Can you confirm?

jviotti:
Awesome! @Julian Please let me know when its up, as I'll need to update my README badges :D

Julian:
Looks like there are some errors:

https://github.com/bowtie-json-schema/bowtie/actions/runs/11374775964/job/31644183016#step:8:11

jviotti:
Whoops, it's missing the standard C++ shared library, etc as its no longer a fully static binary after my `Dockerfile` revamp. Sending a PR in a bit

jviotti:
This should do: https://github.com/bowtie-json-schema/bowtie/pull/1607

PR #1634, opened 2024-10-30
Add jsoncons implementation

michaelmior:
----
📚 Documentation preview 📚: https://bowtie-json-schema--1634.org.readthedocs.build/en/1634/

### DIFF
--- a/implementations/cpp-jsoncons/bowtie_jsoncons.cpp
+++ b/implementations/cpp-jsoncons/bowtie_jsoncons.cpp
@@ -1,0 +1,129 @@
+#include <jsoncons/json.hpp>
+#include <jsoncons_ext/jsonschema/jsonschema.hpp>
+
+#include <cassert>
+#include <cstdlib>
+#include <iostream>
+#include <optional>
+#include <string>
+#include <utility>
+
+using namespace jsoncons;
+
+int main() {
+  bool started{false};
+  std::optional<std::string> default_dialect{std::nullopt};
+
+  std::string line;
+  while (std::getline(std::cin, line)) {
+    const json message{json::parse(line)};
+
+    assert(message.contains("cmd"));
+    assert(message.at("cmd").is_string());
+
+    const std::string command{message.at("cmd").as<std::string>()};
+
+    if (command == "start") {
+      // Validate protocol version
+      started = true;
+      assert(message.contains("version") && message.at("version").is_integer());
+      assert(message.at("version").as<int64_t>() == 1);
+
+      // Respond with implementation details
+      json response;
+      response["version"] = 1;
+
+      json implementation;
+      implementation["language"] = "c++";
+      implementation["version"] = std::to_string(JSONCONS_VERSION_MAJOR) + "." +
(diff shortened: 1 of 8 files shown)


michaelmior:
@Julian Wondering if you could help me figure out what's going on here. @jviotti and I are working on Blaze together and we'd like to make sure we can compare correctness with a wide variety of implementations. For some reason, the new implementation doesn't seem to be picked up.

Julian:
Seems like a CI bug but not sure what's up. Will have a look in an hour or two.

Julian:
@michaelmior think you should be good to go now with the commit I pushed. Glancing over the contents it looks good to me I think. Let me know if you had anything else you know is missing here (for the draft status) otherwise let's see if CI is happy now.

michaelmior:
@Julian I think it's ready for review. CI seems to be running although I have no idea why it's taking slow long. It's already been going for over 25 minutes. In my local environment it was only a couple minutes. I know GHA can be slower, but it seems like something may be up. At least once it's built and pushed it won't be in the critical path.

michaelmior:
Ohh, I see for some reason it's trying to build vcpkg from source. It hasn't even gotten to building jsoncons yet. That wasn't necessary in my local environment. I would just use NuGet like the other C++ implementations, but the jsoncons packages there are very old.

Julian:
Is vcpkg itself something that's in the alpine repos?

michaelmior:
@Julian It's not, but I have a possible fix. I converted back to draft now. Let's see what happens with CI this time.

michaelmior:
So I checked the vcpkg bootstrap script. It has some logic to explicitly detect Alpine Linux and configure vcpkg to use musl. For some reason, this logic isn't working correctly on GitHub Actions. Passing in an explicit argument to enable musl seems to have resolved the issue. Just waiting for CI to finish, but it's looking good now.

implementations/cpp-jsoncons/bowtie_jsoncons.cpp
```diff
@@ -0,0 +1,129 @@
...
+          json test_result;
+          test_result["valid"] = std::move(valid);
+          response["results"].push_back(std::move(test_result));
+        }
+
+        std::cout << print(response) << std::endl;
+      } catch (const std::exception &error) {
+        // Report any errors during compilation or validation
+        json response;
+        response["errored"] = true;
```
Julian, line 112: You're missing including `seq` here (which is the CI error now).

michaelmior:
Okay, I think this should be all set now. Note that I don't believe jsoncons supports any sort of registry, so any tests using `registry` unfortunately must be skipped.

implementations/cpp-jsoncons/bowtie_jsoncons.cpp
michaelmior: Yes, I had fixed that and pushed before but I think it got overwritten when you force pushed your change earlier. I added that back.

Julian:
Awesome, looks good, thanks again!

michaelmior:
@Julian Thanks for the merge!

PR #1747, opened 2025-01-01
Fix bug in hyperjump harness

jdesrosiers:
String comparison is not sufficient for version checks. The current code broke when I added v1.10.0.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1747.org.readthedocs.build/en/1747/


Julian:
Thanks!!

Issue #1752, opened 2025-01-05
Missing Custom 404 Page, Leading to Unclear Errors


Adi-204:
Currently, when users navigate to a non-existent or undefined route, the application throws an unclear error message instead of displaying a user-friendly 404 error page. This issue results in a poor user experience as users are not informed that the page they are looking for does not exist.

Step to Reproduce:
1. Try to access an invalid path in URL which doesn't exits.
![image](https://github.com/user-attachments/assets/d3438533-a261-45f7-9019-8b57f5bd1894)

Expected Behavior:
When a user tries to access an undefined route, the application should display a custom 404 error page with a clear message like:
1. 404: Page Not Found
2. A friendly message explaining that the requested page doesn't exist.
3. A button or link to redirect the user to the homepage.

Adi-204:
Please assign me this issue I would like to contribute.

Julian:
The existing page doesn't bother me a ton honestly, it does say "does not contain the requested file". So even though there's a bit of irrelevant info it's not something that's been top of mind. 

But certainly willing to see what you have in mind, (so sure, happy to assign) -- the offer to contribute is appreciated!

Adi-204:
Proposal 1
![image](https://github.com/user-attachments/assets/10284d50-6d51-4954-af25-81dcb3e2f1ff)

Proposal 2
![image](https://github.com/user-attachments/assets/b1479b8a-897d-4f38-816a-d3933b133191)

@Julian please provide feedback and any suggestions from your side.

Julian:
Is that red color something custom (it looks maybe like it?) or is it a color in our existing color palette / bootstrap theme?

It's a bit jarring. If it isn't in the theme try to use a color (CSS class) which is. Hopefully something a tiny bit less saturated.

Adi-204:
> Is that red color something custom (it looks maybe like it?) or is it a color in our existing color palette / bootstrap theme?
   It is "text-danger". 
> It's a bit jarring. If it isn't in the theme try to use a color (CSS class) which is. Hopefully something a tiny bit less saturated.
   Ok.

Julian:
A 404 is not the end of the world error-wise, I think, so we could either follow GitHub's general style (and just use a heading) or perhaps `text-info` then if we want some color.

Adi-204:
text-info looks good when user is in dark mode 
![image](https://github.com/user-attachments/assets/85f6c7b9-1824-412d-ad00-0976be9016f7)

[...]

PR #1753, opened 2025-01-05
Added Custom 404 Page.

Adi-204:
Closes: #1752 

----
📚 Documentation preview 📚: https://bowtie-json-schema--1753.org.readthedocs.build/en/1753/

### DIFF
--- a/frontend/src/components/NotFound.test.tsx
+++ b/frontend/src/components/NotFound.test.tsx
@@ -4,30 +4,27 @@ import { MemoryRouter, Route, Routes } from "react-router";
 import NotFound from "./NotFound";
 
 interface NotFoundTestParams {
-    path: string,
-    expectedText: string
+  path: string;
+  expectedText: string;
 }
 
 const testCases: NotFoundTestParams[] = [
-    {
-        path: "/random-path",
-        expectedText: "404 Not Found",
-    }
-]
+  {
+    path: "/random-path",
+    expectedText: "404 Not Found",
+  },
+];
 
-describe.each(testCases)(
-    "NotFound Component",
-    ({ path, expectedText }) => {
-        const testName = `Render correctly when navigating to '${path}'`;
-        test(testName, () => {
-        render(
-            <MemoryRouter initialEntries={[path]}>
-                <Routes>
-                    <Route path="*" element={<NotFound />} />
-                </Routes>
-            </MemoryRouter>
-        );
-        expect(screen.getByText(expectedText)).not.toBeNull();
-    });
+describe.each(testCases)("NotFound Component", ({ path, expectedText }) => {
+  const testName = `Render correctly when navigating to '${path}'`;
+  test(testName, () => {
+    render(
+      <MemoryRouter initialEntries={[path]}>
+        <Routes>
+          <Route path="*" element={<NotFound />} />
+        </Routes>
+      </MemoryRouter>,
+    );
+    expect(screen.getByText(expectedText)).not.toBeNull();
+  });
 });
-
--- a/frontend/src/components/NotFound.test.tsx
+++ b/frontend/src/components/NotFound.test.tsx
@@ -1,0 +1,33 @@
+import { describe, expect, test } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter, Route, Routes } from "react-router";
+import NotFound from "./NotFound";
+
+interface NotFoundTestParams {
+    path: string,
+    expectedText: string
+}
+
+const testCases: NotFoundTestParams[] = [
+    {
+        path: "/random-path",
+        expectedText: "404 Not Found",
+    }
+]
+
+describe.each(testCases)(
+    "NotFound Component",
+    ({ path, expectedText }) => {
+        const testName = `Render correctly when navigating to '${path}'`;
+        test(testName, () => {
+        render(
+            <MemoryRouter initialEntries={[path]}>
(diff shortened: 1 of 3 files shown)


Julian:
Thanks. Any chance you know how to add a test for this? Something which just directly browses to a route that doesn't exist and then asserts the component being shown is the not found component?

Adi-204:
I have not previously worked on writing test cases in React, but I am willing to study and implement them as needed. If the implementation does not go as expected, I will remove the test cases and provide a report accordingly.

frontend/src/components/NotFound.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const testCases: NotFoundTestParams[] = [
+  {
+    path: "/random-path",
+    expectedText: "404 Not Found",
+  },
+];
+
+describe.each(testCases)("NotFound Component", ({ path, expectedText }) => {
```
Julian, line 18: I'm not sure I see a reason for the `describe.each` -- I can't think of a reason we'd need a second test here, there's one single behavior (visiting a non-existing route) that we're testing.
So I'd just inline the "array" and create a single normal test.

frontend/src/components/NotFound.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const testName = `Render correctly when navigating to '${path}'`;
+  test(testName, () => {
+    render(
+      <MemoryRouter initialEntries={[path]}>
```
Julian, line 22: I don't think this test is good. If I understand it correctly, what you've done is set up a *new* router which always renders the `NotFound` component, and then browsed to some path (`/random-path` -- but even `/` would have the same behavior!) and asserted the 404 text is shown.

That isn't useful.

You should test *our* application renders the `NotFound` component on a route not known by *our* application, otherwise we're testing something completely unconnected to our application, which could be entirely broken (say if you forgot to make the change to `index.tsx` -- this test would still pass).

Adi-204:
Thank you for guidance.

Should I make following changes - 
import { expect, test } from "vitest";
import { render, screen } from "@testing-library/react";
import { MemoryRouter, RouterProvider } from "react-router";
import { router } from "..";

test('landing on a bad page', () => {
    const badRoute = '/random-path';
    render(
        <MemoryRouter initialEntries={[badRoute]}>
            <RouterProvider router={router} />
        </MemoryRouter>,
    )
    expect(screen.getByText("404 Not Found")).not.toBeNull();
});


Source  - https://testing-library.com/docs/example-react-router/

Julian:
That looks like it has the same issue as your current test, you create a new (memory) router instead of using our real one.
(I'm not familiar with that page, but it seems a bit strange as a recommendation.)
The "real" solution here is likely we should be testing through [playwright](https://playwright.dev/) as an end-to-end test.
But to keep the work minimal, I haven't tried it, but I would assume your test should work just fine if you import our [real hash router](https://github.com/bowtie-json-schema/bowtie/blob/40c34ff6915bde1336946e46908271dc60b0200f/frontend/src/index.tsx#L75) which is what we care about the behavior of, and request against it? At least that should be the first thing tried.

Adi-204:
I have imported the router from index.tsx (line no. 4) and used the spread operator to copy all paths from there, setting the entry point as /random-path.
If the following code is missing in the router, the test case will fail:
{
    path: "*",
    Component: NotFound,
  },
However, if this code is present, the test case will pass.

Are you telling that we should test app on all other existing paths as well ?

Julian:
> I have imported the router from index.tsx (line no. 4) and used the spread operator to copy all paths from there, setting the entry point as /random-path. So new (memory) router is used to set initialEntry point of website when all routes are present in route.

I see, sorry, it's hard to review screenshots.

I don't yet understand what the point of that is though, what goes wrong if you literally pass `RouterProvider router={ourHashRouter}`?

Adi-204:
If we render <RouterProvider router={ourHashRouter} /> without specifying an initial path, the test framework will not know which route to check. To set the initial route, we need to use createMemoryRouter, which allows us to define initialEntries.
This ensures that when testing for /random-path, the router will not find a matching route, triggering the wildcard (*) path and rendering the NotFound component. The test can then verify that the expected content is present in the rendered output.

Julian:
I don't follow why the test framework needs an initial route? The route we check should be one you browse to, i.e. `/random-path`. https://test-utils.vuejs.org/guide/advanced/vue-router perhaps seems like a relevant page, what I'm talking about is the "test with the real router" bits described there.

Julian:
If after reading that page it isn't obvious how to write this test I'll make some time in the next few days to figure it out.

Adi-204:
I explored alternative ways to test routing in React but found that createMemoryRouter is the most suitable approach. However, I couldn't find an equivalent method similar to Vue. Given this, should I remove the test implementation and proceed with committing only the addition of the 404 Not Found page? Also, could you please merge only this change?

Julian:
I don't plan to merge this without tests (all PRs merged should have tests!). But I'm happy to try to make time to figure out how to best test this if you think what you have is the best you've seen.

Adi-204:
Thanks, I will also try to find a better solution.

Adi-204:
@Julian just wanted to ask the status of the tests for the above. Let me know if you need any help or input from my side.

Julian:
I made some time to do the work to add support for Playwright tests and moved your test over to that. Thanks for the PR, appreciated.

PR #1765, opened 2025-01-08
Work in progress: Harness for elixir-jsv

lud:
I have not implemented the "errored" part yet, but I'm open to suggestions!

TODO
- [x] Multistage Dockerfile
- [ ] remove `test.sh`

----
📚 Documentation preview 📚: https://bowtie-json-schema--1765.org.readthedocs.build/en/1765/


implementations/elixir-jsv/lib/bowtie_jsv.ex
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  defp handle_command(%{"cmd" => "start", "version" => 1}, state) do
+    start_response = %{
+      version: 1,
+      implementation: %{
+        language: "elixir",
+        name: "jsv",
+        source: "https://github.com/lud/jsv",
+        homepage: "https://github.com/lud/jsv",
```
Julian, line 48: Including the version of your library here (via some Elixir-specific way to lookup what version of your library has been installed) is nice and will make that show up in the UI eventually.

implementations/elixir-jsv/lib/bowtie_jsv.ex
```diff
@@ -0,0 +1,92 @@
...
+
+    {:reply, start_response, state}
+  end
+
+  defp handle_command(%{"cmd" => "stop"}, _state) do
+    System.halt()
+  end
+
+  defp handle_command(%{"cmd" => "dialect", "dialect" => dialect}, state) do
+    true = dialect in @dialects
```
Julian, line 62: (Again with apologies for not knowing any Elixir), what does this line do?

Is `true` not a keyword in Elixir (i.e. is this assigning a variable called true)?

It's a bug if Bowtie ever sends you a dialect you don't support, so generally there's no need to check this in the harness. What this response is meant to mean is simply "does your implementation support a schema like `{"type": "integer"}` if I tell you externally "that's a draft 7 schema, interpret it as such".

(So I'm suspicious if `true` isn't literally always "return `true` statically", since from below it would seem you do support externally identifying schema dialects).

Julian:
This looks like a great start, thanks again! Left some initial comments!

implementations/elixir-jsv/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN mix local.hex --force && mix local.rebar --force
+
+COPY mix.exs mix.exs
+
+RUN mix deps.get && mix deps.compile
+
+COPY lib lib
+
+RUN mix compile && mix escript.build
```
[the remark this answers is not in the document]
lud: Hello,
No elixir releases are not binaries, it's a way of packaging the app and its dependencies, with optionally an Erlang runtime, so it is independent. But it will contain the bytecode of each module independently.

And unfortunately when starting a realease there may be some output to stdout, and dealing with stdio is not so fun.

What I have done is build the app as an [escript](https://www.erlang.org/docs/18/man/escript) which is actually everything in a single file and plays nice with stdio. But I still have to build that in docker to have a proper elixir environment.

What we could do is to `rm` the rest of the files once the escript is built. An if we embed the erlang runtime in the script we could just use a 2-stage Dockerfile with a bare alpine linux image as the second image, just copying the escript.

implementations/elixir-jsv/lib/bowtie_jsv.ex
lud: Yeah I should remove this line by now, it is indeed a check to see if the given dialect is in the list. I did this because I do URL normalization, and that trailing `#` in the draft 7 metaschema was causing problems. But it's fixed by now.

(For technical details, `true` is an atom, not a variable indeed, so we are not assigning a value to `true` here (though that sounds like a lot of fun haha), it is meant to crash if the right-hand expression is not equal to `true`.)

implementations/elixir-jsv/Dockerfile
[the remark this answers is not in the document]
Julian: Got it!

> An if we embed the erlang runtime in the script we could just use a 2-stage Dockerfile with a bare alpine linux image as the second image, just copying the escript.

This sounds like the sort of thing we do in similar cases (where it's not a single binary but still there's some self contained thing we can copy over).

But we can certainly also wait until everything else is done there to experiment more with this, it can be boring fighting container images as it sounds like you already have had some fun with.

implementations/elixir-jsv/lib/bowtie_jsv.ex
Julian: > For technical details, true is an atom, not a variable indeed, so we are not assigning a value to true here (though that sounds like a lot of fun haha), it is meant to crash if the right-hand expression is not equal to true

Aha! Interesting, I was hoping it was more something like that, interesting.

implementations/elixir-jsv/lib/bowtie_jsv.ex
lud: Done!

implementations/elixir-jsv/Dockerfile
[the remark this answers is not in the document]
lud: Yeah let's do it at the end but I'm sure it's going to work fine.

lud:
Alright so I added basic error handling. For some reason the CI is failing on python code...

What should I do next? It there a way to validate formats? For now this reports zero failures:

```
bowtie suite -x -i localhost/jsv-bowtie draft2020-12 | bowtie summary --show failures
bowtie suite -x -i localhost/jsv-bowtie draft7       | bowtie summary --show failures
```

So I guess formats are not validated. Maybe it passes the schema-assertion meta for 2020-12.

implementations/elixir-jsv/lib/bowtie_jsv.ex
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  defp jsv_vsn do
+    {:ok, jsv_vsn} = :application.get_key(:jsv, :vsn)
+    List.to_string(jsv_vsn)
+  end
+
+  defp errored(e, stacktrace) do
+    %{
+      errored: true,
```
Julian, line 106: You're missing passing `seq` here -- error responses still need to indicate which test case they blew up on.

Julian:
> Alright so I added basic error handling

Nice! Left one comment but we're close then.

> For some reason the CI is failing on python code...

Yes looks like certainly unrelated changes upstream -- merge `main` and they should go away.

> What should I do next? It there a way to validate formats? For now this reports zero failures:

We don't run the optional tests quite yet (that's #36 which I didn't get a chance to get to unfortunately). So passing all of them is a good sign!

implementations/elixir-jsv/lib/bowtie_jsv.ex
lud: Fixed!

lud:
Alright so no need for the formats right now.

I migrated to a multistage Dockerfile. I had to use a mix release as you suggested, without any stdio problems.

Julian:
Great! Let's give this a shot, gonna merge and will do a release so you show up on the site. 

Thanks for contributing a harness (and congrats on your implementation it sounds like)!

Stay in touch!

lud:
Alright thank you ! 

As a cleanup task I suggest to remove the `test.sh` file. I can file a PR if you want but I guess this can be done directly by your team if you prefer.

Julian:
This is released though something strange is happening, specifically the container crashes on startup nondeterministically (like 50% of the time) which causes the report generation to fail. I haven't gotten a chance to look into it yet on why.

lud:
Strange,

Tell me if you have any error output.

I can revert to an escript if you want. The image will be slightly heavier but not much, and may be easier to debug.

lud:
@Julian did you figure it out?

Julian:
@lud I've been on vacation actually the last week or two so I didn't yet unfortunately -- the error output is strangely missing which is what I want to have a look at, perhaps Bowtie's harness handling code is missing a case and hiding the error that's happening. At the minute I'm not sure whether reverting to an escript will or won't help, either way it's important to fix the issue causing no output (unless really nothing is being emitted which would be... strange).

So yeah basically it's been on hold until I'm back home -- if you do see anything obvious obviously let me know, otherwise yeah I'll try to get back to it when I'm back home around the end of this week.

lud:
Hey! Hope you had a great time :)

Alright I'll look into it. Could you give the the last input that is sent to the container to help figuring it out?

PR #1805, opened 2025-02-01
fix: Do not stop elixir-jsv container on EOF

lud:
Hey @Julian 

So I'm not sure why you removed one `rescue` block here, because on exception the process that runs the harness will crash, and the supervisor will not restart it (default behaviour for `Task`, and on restart it would lose context anyway).

Tell me if you want that I remove it again. Anyway, since botwie sends valid schemas that should not be a problem.

I also removed the debug lines by default so you can have a cleaner output, and spot actual error messages.

The main fix here is that previously on end of input the harness would terminate. I suspect this could cause startup errors when bowtie is sending input deparated by some time where it would validate the harness output.

Can you run that in the same environment as what the CI does to see if it's working properly?

----
📚 Documentation preview 📚: https://bowtie-json-schema--1805.org.readthedocs.build/en/1805/

I also get a lot of this errors:

> github3.exceptions.ForbiddenError: 403 API rate limit exceeded for xxx.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

I tried to define the `GITHUB_TOKEN` environment variable but it does not seem to have much impact.


Julian:
Thanks for taking another look at this, I'm just back myself today -- it isn't valid to return an errored response to anything but a run request, so the code looked wrong to me before. I have only just glanced at it but this seems wrong in the same way, though perhaps the other changes you made will help diagnose?

lud:
The other place where the `errored` function is called is here:

```elixir
results = Enum.map(tests, fn test -> run_test(test, root) end)
```

```elixir
  defp run_test(test, root) do
    %{"instance" => data} = test

    case JSV.validate(data, root) do
      {:ok, _} -> %{valid: true}
      {:error, _} -> %{valid: false}
    end
  rescue
    e -> errored(e, __STACKTRACE__)
  end
```

So we build a list of `{"valid": true}`, `{"valid": false}` or `{"errored": true, "context" ...}` objects. Not that I do not pass the `{"seq": ...}` addiitonal data.

So it seems valid, right?

Maybe I should also validate the replies :D

Julian:
> So it seems valid, right?

I'm blind perhaps! No idea what I was thinking before but yes now looking again I don't see whatever mistake I was making before -- thanks for fixing my silly error.

Let's merge then and see if things improve.

lud:
Alright!

Julian:
Seems the same unfortunately, still [failing on startup nondeterministically](https://github.com/bowtie-json-schema/bowtie/actions/runs/13089906879/job/36525359663#step:4:35)... The only place that I saw *doesn't* have decent error handling in Bowtie at the minute is if it's sending back an *invalid* request on start, so I'll have to see if that's the case, but the error is not reproducible locally, which makes me think it only happens on Linux AMD64 rather than macOS and/or ARM64...

Julian:
Actually glancing again without looking too closely again yet, the behavior is even stranger -- it's seemingly not necessarily non-deterministic, it seems one job succeeds at a time -- almost as if the container is contending on a shared resource and failing.

I'm confused as to why that's possible, as if it's say, a port, I would have expected some network isolation from the container -- but does that perhaps ring any bells with you? Does Elixir somehow expect to communicate over a default TCP port or something???

lud:
I'm running on Xubuntu x86_64 and I can run multiple versions of the containers in parallel without problem.

Can you show me how the container is started and called?

I think I'll revert to an escript version of the harness, that should be more straightfoward for IO handling, much more closer to how other harnesses work.

lud:
This is the current output on my machine:

```
❯ bowtie suite $(bowtie filter-implementations -l elixir | sed 's/^/-i /') https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft4 
warning: unsupported-dialect

× 'elixir-jsv' does not support Draft 4.

note: its supported dialects are: Draft 2020-12, Draft 7
      
No implementations started successfully!
```

I have changed nothing yet but I hope I have time to setup the escript tonight.

Julian:
Isn't that just because you're running the draft 4 tests? Does it work if you run the draft 7 ones?

Julian:
(The way that CI runs the container is like that, via `bowtie suite` -- unless you're asking more specifically which Docker SDK calls are made?)

lud:
Yeah it's an error with draft 4 but I get the expected output. I started the command on the error in [your link](https://github.com/bowtie-json-schema/bowtie/pull/1805#issuecomment-2628997243). When I saw it was draft 4 I suspected that maybe unsupported drafts were the cause of the error.

PR #1810, opened 2025-02-03
fix: match dialects dropdown position

tthijm:
I think it might look better if the position of the dialects dropdown is consistent between the dialect and benchmarks page.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1810.org.readthedocs.build/en/1810/

### DIFF
--- a/frontend/src/components/NavBar.tsx
+++ b/frontend/src/components/NavBar.tsx
@@ -99,11 +99,6 @@ const NavBar = () => {
               )}
               {isBenchmarksPage && (
                 <>
-                  <li className="nav-item">
-                    <Link className="nav-link" to="/">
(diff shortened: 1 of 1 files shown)


Julian:
Thanks for this -- I'm going to close it though not because I don't appreciate the PR (I do) but because I think the navbar needs even deeper rethinking to really be useful as right now it doesn't seem nice to me that we mix "decide which dialect you want to look at" (which is the dialects dropdown on both pages) with "what are the main pages on the site" (which right now are the dialect compliance reports and the benchmarks reports, but we also have implementation pages that aren't really linked in the navbar at all!), so I'd rather simply think that through than do small patchwork. Your feedback or help is definitely welcome there in having a better overall layout!

PR #1812, opened 2025-02-04
build: Move from dev-master to 6.1 version

DannyvdSluijs:
This is following up on https://github.com/bowtie-json-schema/bowtie/pull/1512 and moves from the `dev-master` version to the version `6.1`


Julian:
Thanks, congrats on the release!

PR #1822, opened 2025-02-09
feat: improve bowtie info pretty output

### LINKED ISSUE #1518: `bowtie info`'s pretty output is pretty ugly
It should make deeper use of `rich`, and in particular not show dialects in array syntax.

tthijm:
Closes #1518.

----
📚 Documentation preview 📚: https://bowtie-json-schema--1822.org.readthedocs.build/en/1822/

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -1660,6 +1660,62 @@ def latest_report(dialect: Dialect):
     asyncio.run(write(dialect.latest_report()))
 
 
+def _info_links_table_for(metadata: dict[str, Any]):
+    table = Table(
+        Column(style="spring_green4"),
+        box=None,
+        padding=(0, 1, 0, 0),
+        show_header=False,
+    )
+
+    table.add_row("homepage", metadata["homepage"])
+    table.add_row("source", metadata["source"])
+    table.add_row("issues", metadata["issues"])
+
+    if "documentation" in metadata:
+        table.add_row("documentation", metadata["documentation"])
+
+    for link in metadata.get("links", []):
+        table.add_row(link["description"], link["url"])
+
+    return table
+
+
+def _info_table_for(metadata: dict[str, Any]):
+    table = Table(
+        Column(style="cyan bold"),
+        box=box.ROUNDED,
+        show_header=False,
+        border_style="bright_black",
+    )
+
+    table.add_row(
+        "implementation",
+        f"{metadata["name"]} [grey58]{metadata.get("version", "")}[/grey58]",
+    )
+    table.add_row(
+        "language",
+        f"{metadata["language"]} [grey58]{metadata.get("language_version", "")}[/grey58]",  # noqa: E501
+    )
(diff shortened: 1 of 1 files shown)


bowtie/tests/test_integration.py
```diff
@@ -2247,28 +2247,20 @@ async def test_info_links():
...
-            "url": "urn:example:bar"
-          }}
-        ]
+        ╭──────────────────┬─────────────────────────────────────────╮
+        │ name             │ links                                   │
+        │ language         │ python                                  │
+        │ version          │ v1.0.0                                  │
+        │ homepage         │ urn:example                             │
+        │ issues           │ urn:example                             │
+        │ language version │ {platform.python_version():39} │
```
Julian, line 2256: This seems very likely to break if we change the terminal size we're simulating. Probably at least you should reference that value, I think we have a constant somewhere here in the test file.

bowtie/tests/test_integration.py
```diff
@@ -2247,28 +2247,20 @@ async def test_info_links():
...
+        │ os               │ {platform.system():39} │
+        │ os version       │ {platform.release():39} │
+        │ source           │ urn:example                             │
+        │ dialects         │ http://json-schema.org/draft-07/schema# │
+        │ links            │ urn:example:foo                         │
```
Julian, line 2261: This seems to have lost the link description, which probably we should include.

bowtie/tests/test_integration.py
```diff
@@ -2006,24 +2006,24 @@ async def test_info_pretty():
...
+        │ name             │ always_invalid                                      │
+        │ language         │ python                                              │
+        │ version          │ v1.0.0                                              │
+        │ homepage         │ https://bowtie.report/                              │
+        │ issues           │ https://github.com/bowtie-json-schema/bowtie/issues │
+        │ language version │ {platform.python_version():51} │
+        │ os               │ {platform.system():51} │
+        │ os version       │ {platform.release():51} │
+        │ source           │ https://github.com/bowtie-json-schema/bowtie        │
+        │ dialects         │ https://json-schema.org/draft/2020-12/schema        │
```
Julian, line 2019: This seems like a slight improvement, so thanks! Though while we're here, I think it'd be nice to even further improve this by introducing some grouping (e.g. os and os version belong together, and language and language version belong together, and the 4 of those likely belong together in one parent group about versions, etc.)

It'd be nice to put some thought into that while you're looking at this (how we can group these fields in such a way that related information is next to each other).

Julian (commented):
Thanks for the PR.

Left some comments.

(Also it's good to say something when submitting a PR :) -- you're trying to tackle an issue which is super appreciated, but it's good to at least say a sentence or two about what your PR does, what you've considered, etc.)

tthijm:
I will fix the integration tests once the design has been approved.

tthijm:
I will fix the integration tests once the design has been approved.

PR #1844, opened 2025-02-19
Feature/dev harrel new drafts support

harrel56:
I think it should pick up new drafts automatically, just bumping versions and removing test case skipping

----
📚 Documentation preview 📚: https://bowtie-json-schema--1844.org.readthedocs.build/en/1844/


Julian:
The audit failure is some ignorable nonsense which is clearly not relevant (and is now fixed on main), so this looks good I think, will merge, congrats!

Issue #1849, opened 2025-02-20
[POC] Extract test harness implementation into a separate repository


OptimumCode:
This POC aims to gather the key steps and changes in bowtie pipelines required to move test harness implementations into dedicated repositories. This is quite a wall of text but please bear with me till the end 🍀 

The POC will include:

- [x] **creating a separate repository** for one of the test harnesses (my choice is [kotlin-kmp-json-schema-validator](https://github.com/bowtie-json-schema/bowtie/tree/main/implementations/kotlin-kmp-json-schema-validator) but it can be any). This step will require a new repository inside `bowtie-json-schema` organization - @Julian, could you please assist with that when we complete the initial discussion?
https://github.com/bowtie-json-schema/kotlin-kmp-json-schema-validator
- [x] **moving the implementation** from the bowtie repository into a new one and configuring the pipeline to **build**, **validate**, and **publish** the test harness. I believe this one will not require any assistance. All required permissions will be configured inside the workflow files. The published image should have the **same location as the original one**.
I also think we can mark the moment when the implementation is moved with a **git tag** which should help link the history of 2 repositories;
https://github.com/bowtie-json-schema/bowtie/pull/1897
- [x] **adjusting the `bowtie filter-implementations` command** to include external implementations. As @Julian [pointed out](https://github.com/bowtie-json-schema/bowtie/issues/1833#issuecomment-2669696024), the information about available implementations is collected at [build time](https://github.com/bowtie-json-schema/bowtie/blob/main/hatch_build.py#L38). This works perfectly right now but will stop working once the test harnesses are moved to separate repositories. The possible solution is to either query GitHub API for [packages](https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#list-packages-for-an-organization) or [repositories](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-organization-repositories). I would vote for using packages API as it also has the option to query [versions](https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#list-package-versions-for-a-package-owned-by-an-organization) for the package. I think it will be very handy for `bowtie-trends`.
During POC and until all the test harnesses are moved to dedicated repositories the implementation collection will exist in a hybrid state using both local information and information from the GitHub API we pick;
https://github.com/bowtie-json-schema/bowtie/pull/1867
[...]

PR #1867, opened 2025-03-04
Use GitHub packages API to collect known implementations

OptimumCode:
A part of POC #1849 is a modification of how bowtie collects the information about known implementation.
This PR adds functionality to collect known implementations using [GitHub Packages API](https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#list-packages-for-an-organization).

I am not a Python developer - sorry in advance if I made some stupid mistakes)

----
📚 Documentation preview 📚: https://bowtie-json-schema--1867.org.readthedocs.build/en/1867/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,7 +156,6 @@ jobs:
       contents: write
       id-token: write
       attestations: write
-      packages: read
 
     steps:
       - uses: actions/checkout@v4
@@ -170,7 +169,7 @@ jobs:
 
       - name: Build our distributions
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.READ_PACKAGES_PAT }}
         run: uv run --frozen --with 'build[uv]' -m build --installer=uv
 
       - name: Generate attestation for wheels
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Run nox
         env:
-          GITHUB_TOKEN: ${{ secrets.READ_PACKAGES_PAT }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uvx nox -s "${{ matrix.noxenv }}"
 
       - name: Upload Playwright Report
@@ -109,7 +109,7 @@ jobs:
 
       - name: Run nox
         env:
-          GITHUB_TOKEN: ${{ secrets.READ_PACKAGES_PAT }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uvx nox -s tests-3.13 -- -m 'not containers'
 
   coverage:
@@ -173,7 +173,7 @@ jobs:
 
       - name: Build our distributions
         env:
-          GITHUB_TOKEN: ${{ secrets.READ_PACKAGES_PAT }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uv run --frozen --with 'build[uv]' -m build --installer=uv
 
       - name: Generate attestation for wheels
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,6 +176,20 @@ jobs:
           BUILD_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uv run --frozen --with 'build[uv]' -m build --installer=uv
 
+        # this is temporal step to make sure we don't lose any of the implementations during migration
+        # remove once all implementations are successfully migrated to separate repos
+      - name: Collect known implementations
+        env:
+          BUILD_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          EXPECTED_IMPL_COUNT: 32
+        run: |
+          impl_count=$(uvx --from . --python 3.13 bowtie filter-implementations --format plain | wc -l)
+          if [ $impl_count -lt $EXPECTED_IMPL_COUNT ]; then
+            echo "bowtie collected less implementations than expected: $impl_count instead of $EXPECTED_IMPL_COUNT" >> $GITHUB_STEP_SUMMARY
+            uvx --from . --python 3.13 bowtie filter-implementations --format plain >> $GITHUB_STEP_SUMMARY
+            exit 1
+          fi
+
       - name: Generate attestation for wheels
         if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
         uses: actions/attest-build-provenance@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,7 @@ jobs:
       contents: write
       id-token: write
       attestations: write
+      packages: read
 
     steps:
       - uses: actions/checkout@v4
@@ -168,6 +169,8 @@ jobs:
           enable-cache: true
 
       - name: Build our distributions
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
         run: uv run --frozen --with 'build[uv]' -m build --installer=uv
 
       - name: Generate attestation for wheels
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
 
       - name: Build our distributions
         env:
-          GITHUB_TOKEN: ${{ github.token }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uv run --frozen --with 'build[uv]' -m build --installer=uv
 
       - name: Generate attestation for wheels
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Run nox
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BUILD_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uvx nox -s "${{ matrix.noxenv }}"
 
       - name: Upload Playwright Report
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -156,6 +156,7 @@ jobs:
       contents: write
       id-token: write
       attestations: write
+      packages: read
```
OptimumCode, line 159: issue: Apparently, the packages API does not support authentication using a GITHUB_TOKEN
> GitHub Packages only supports authentication using a personal access token (classic). For more information, see [Managing your personal access tokens](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token).

https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#about-scopes-and-permissions-for-package-registries

@Julian could you please create a personal access token with `packages: read` and add it to the secrets when you have time? The name of the secret is up to you (I can suggest something like _PACKAGES_TOKEN_ or _READ_PACKAGES_PAT_)

OptimumCode: Sure, but it looks like you need to add me to the organization first (I tried to do the transfer but got an error)
image

Julian:
Invite sent!

OptimumCode:
The repository has been transferred

pyproject.toml
```diff
@@ -12,6 +12,9 @@ exclude = [".github"]
 packages = ["bowtie"]
 
 [tool.hatch.build.targets.wheel.hooks.custom]
+dependencies = [
+  "requests"
```
Julian, line 16: I think GitHub3.py would have possibly still saved you from implementing the pagination manually (it doesn't have a packages API but you can I think use `GitHubSession` to make a request to an arbitrary URL) -- but since you wrote it already I'm also fine with just using an HTTP client. But not requests, can you use `httpx` instead -- I think nothing should probably change, just replacing `requests` with `httpx` I think should more or less work the same.

hatch_build.py
```diff
@@ -53,3 +57,72 @@ def finalize(
...
+        self,
+        known_local: list[str],
+    ) -> list[str]:
+        """
+        Collects available implementation using GitHub packages API.
+        """
+        gh_token = os.getenv("GITHUB_TOKEN")
+
+        if not gh_token:
+            self.app.display_warning(
```
Julian, line 71: (I'm also OK immediately requiring the token to be set, as really the only thing that runs this script is GitHub actions here in the repo, and me when building the package locally I think) But this is of course good as-is.

Julian: Yeah that seems like a nice idea too!

.github/workflows/ci.yml
OptimumCode: All works now! Thank you

OptimumCode:
It seems like we have a problem: the documentation configuration depends on bowtie to [get the links](https://github.com/bowtie-json-schema/bowtie/blob/main/docs/conf.py#L6) (GITHUB, HOMEPAGE, etc).
According to the readthedocs documentation, it is possible to add a custom env variable that will be available during the build
https://docs.readthedocs.com/platform/stable/guides/environment-variables.html#how-to-use-custom-environment-variables

But it also says this:

> If you make your environment variable public, any user that can create a pull request on your repository will be able to see the value of this environment variable. **In other words, do not use this option if your environment variable is a secret.**

So, it does not look like a good solution to the problem...

Also, the documentation configuration iterates over the implementations to collect their count.

@Julian do you think it would be possible to depend on the latest published version of the bowtie instead of building it from the sources? The only potential problem I see is that the implementations count might be wrong on the generated preview doc page

OptimumCode:
It looks like we might want to use **repositories** API instead of **packages** - it does not require a token for access
```bash
curl -L -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/orgs/bowtie-json-schema/repos
```

hatch_build.py
```diff
@@ -53,3 +60,53 @@ def finalize(
...
+        # join with local implementation until fully migrated to packages API
+        # https://github.com/bowtie-json-schema/bowtie/issues/1849
+        all_harnesses = list(local_set.union(harnesses_set))
+        # sorting is done just to simplify debugging
+        # if some harnesses are missed for some reason
+        all_harnesses.sort()
+        return all_harnesses
+
+    @staticmethod
+    def _collect_harnesses(gh_token) -> list[str]:
```
OptimumCode, line 104: note: I have tested this locally - the `kotlin-kmp-json-schema-validator` is returned as a collected test harness

OptimumCode:
Cannot check all the jobs before https://github.com/bowtie-json-schema/bowtie/pull/1879 is merged

.github/workflows/ci.yml
```diff
@@ -81,6 +81,8 @@ jobs:
           enable-cache: true
 
       - name: Run nox
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Julian, line 85: I'd like to try to avoid accidental leakage of the github token into *non*-build steps -- in other words I don't want it to accidentally end up the case that a *test* relies on talking to GitHub (so that that becomes a runtime dependency) rather than just the *build* getting access to the token. I think it needs a few seconds of thought for what the best way to do that is -- one would think it'd be "only pass this envvar to the `build` noxenv, but of course this isn't sufficient, as right now we re-build bowtie in each nox environment I think...

I haven't thought too hard yet about how to do this (though I think it won't be terribly difficult) but does the concern make sense to you and/or does anything immediately occur to you?

OptimumCode: When I tried running [CI with the required GITHUB_TOKEN](https://github.com/bowtie-json-schema/bowtie/actions/runs/13704277457) the following jobs failed because it was not provided:

- ci (build)
- ci (tests)
- ci (typing)
- ci (docs)  - all except ci (docs(style))
- ci (ui(tests))
- conteinerless

Unfortunately, I don't have enough expertise with `python` or `nox` to suggest anything to avoid building bowtie in those jobs...

We can remove the env variable for `Run nox` step (and all others) - it is not required now. However, we might hit the rate limit (it is only [60 requests per hour for unauthenticated users](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users)). This limit does not look enough to me - it is just ~3 runs of CI jobs... Of course, the limit is `associated with the originating IP address` I am not sure how GitHub creates the runners and if GitHub API will see the same IP for different runners... (but we could try, I think, to execute 10-15 workflows in a raw without a token to see if we hit a limit)

Regarding the leakage of the GitHub token:

I get your concern, but I don't think that it is a big problem. The GitHub token leaves [24 hours max](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret). We can also configure it to have only `read` permissions so even if it leaks it should not a big problem

Julian: To be clear I'm not talking about a security issue! What I mean is I don't want a test to only pass because it's set (i.e. a test isolation issue!) But I'll have another think, thanks for your research!

.github/workflows/ci.yml
OptimumCode, line 1: > [!NOTE]
>  TODO: add a step to check the number of implementations does not decrease (until we fully move all harnesses into separate repositories)

.github/workflows/ci.yml
OptimumCode: Ahh, okay, I see now. Thank you for the clarification.

Maybe we could clear GITHUB_TOKEN env variable before the test execution?

But I don't think that such a dependency is possible since the only place where the GITHUB_TOKEN is used is a build hook

OptimumCode: But I am happy to assist with any option you think works better here

OptimumCode:
Hi, @Julian!

There are a few new alerts from ruff:

- TC006 [*] Add quotes to type expression in `typing.cast()`
- C420 [*] Unnecessary dict comprehension for iterable; use `dict.fromkeys` instead

They all have auto-fix as I can see. Would you like me to open a separate PR to fix them? They seem more or less reasonable to me

Julian:
Yes that seems reasonable to me too on the ruff changes!

Julian:
Feel free to merge that new PR immediately even by the way!

OptimumCode:
> Yes that seems reasonable to me too on the ruff changes!

Done in https://github.com/bowtie-json-schema/bowtie/pull/1890

OptimumCode:
Hi, @Julian ! I have added a step to check the number of known implementations to `packaging`.
After updating UI deps to fix the CI it was successfully executed (also I checked the code locally to see if it actually reports error in something was missed). Could you please take a look at PR when you have time?

.github/workflows/ci.yml
```diff
@@ -168,8 +172,24 @@ jobs:
           enable-cache: true
 
       - name: Build our distributions
+        env:
+          BUILD_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: uv run --frozen --with 'build[uv]' -m build --installer=uv
 
+        # this is temporal step to make sure we don't lose any of the implementations during migration
```
Julian, line 179:
```suggestion
        # this is temporary step to make sure we don't lose any of the implementations during migration
```

hatch_build.py
```diff
@@ -53,3 +60,56 @@ def finalize(
...
+        )
+
+        harnesses_set = set(harnesses)
+        local_set = set(known_local)
+        if harnesses_set.issubset(local_set):
+            return known_local
+
+        # join with local implementation until fully migrated to packages API
+        # https://github.com/bowtie-json-schema/bowtie/issues/1849
+        all_harnesses = list(local_set.union(harnesses_set))
```
Julian, line 100:
```suggestion
        all_harnesses = sorted(local_set | harnesses_set)
```

(and then you can remove the `.sort()` below).

Julian (commented):
Minor comments but then this seems great, thanks again for all your hard work!

Feel free to merge.

hatch_build.py
OptimumCode: Thank you for the suggestion! It looks much cleaner

OptimumCode:
> Minor comments but then this seems great

Just one modification: the locally collected harnesses were not sorted so I added a `sort` call there (281a64424fedf3507a98acbbff7d24a364675d96). If you are okay with that @Julian I will merge changes and we can move further I think: publish image from the new harness repo and remove harness directory from bowtie repo

PR #1879, opened 2025-03-09
Update frontend dependencies

OptimumCode:
This PR updates frontend deps to the latest available version. It also resolves failing UI tests - the driver location was changed after the new playwright release

----
📚 Documentation preview 📚: https://bowtie-json-schema--1879.org.readthedocs.build/en/1879/

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -25,7 +25,9 @@
   "packageManager": "pnpm@9.15.0",
   "pnpm": {
     "auditConfig": {
-      "ignoreGhsas": ["GHSA-67mh-4wv8-2f99"]
+      "ignoreGhsas": [
+        "GHSA-67mh-4wv8-2f99"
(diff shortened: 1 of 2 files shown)


OptimumCode:
btw, @Julian, could you please advise if this is expected?
image
It says the test has failed but the CI is green

Julian:
Oh wonderful, no that's definitely not expected :(

That test is a bit horrid in that it takes a long time and I've frequently seen it time out but it definitely should not be failing and still saying the build passed, so if you see why that's happening definitely up for a fix.

Julian:
(happy to merge this PR anyhow until we figure it out, by the way)

OptimumCode:
> so if you see why that's happening definitely up for a fix

Don't know what is the reason for that yet. But it would be good to merge PR to unlock others

PR #1916, opened 2025-03-26
Use GH API to collect versions from remote test harnesses

OptimumCode:
The PR adds a collection of version information from remote test harnesses. It collects all available version tags (like `harness-release-*`), plus it uses the `latest` tag to get the latest available test harness version.

Related to #1849 

----
📚 Documentation preview 📚: https://bowtie-json-schema--1916.org.readthedocs.build/en/1916/

### DIFF
--- a/.github/workflows/versioned-report.yml
+++ b/.github/workflows/versioned-report.yml
@@ -29,7 +29,7 @@ jobs:
       - name: List all Bowtie Supported Implementations having a "matrix-versions.json" file
         id: implementations-matrix
         env:
-          GH_TOKEN: ${{ github.token }} # for GH API access
+          GH_TOKEN: ${{ github.token }} # because we use gh CLI in the script (https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows)
         run: |
           implementation=${{ inputs.implementation }}
           IMPLEMENTATIONS=$(bowtie filter-implementations --format json)
@@ -93,7 +93,7 @@ jobs:
       - name: Generate New Versioned Reports for all Supported Dialects of ${{ matrix.implementation }}
         id: generate-new-versioned-report
         env:
-          GH_TOKEN: ${{ github.token }} # for GH API access
+          GH_TOKEN: ${{ github.token }} # because we use gh CLI in the script (https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows)
         run: |
           impl=${{ matrix.implementation }}
           MATRIX_VERSIONS_FILE="implementations/$impl/matrix-versions.json"


.github/workflows/versioned-report.yml
```diff
@@ -78,9 +92,19 @@ jobs:
...
         run: |
           impl=${{ matrix.implementation }}
           MATRIX_VERSIONS_FILE="implementations/$impl/matrix-versions.json"
+          if [ ! -f $MATRIX_VERSIONS_FILE ]; then
+            echo "Dump versions for remote test harness $impl"
+            mkdir $(dirname "$MATRIX_VERSIONS_FILE")
+            gh api \
+              -H "Accept: application/vnd.github+json" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              /repos/$GH_ORGANIZATION/$impl/git/matching-refs/tags/v | jq -c '[ .[].ref | ltrimstr("refs/tags/v")] + ["latest"]' > $MATRIX_VERSIONS_FILE
```
OptimumCode, line 106: question: the last existing version will be displayed as `latest`. Could it cause any issues with creating a graph, @Julian? Also, does the order matter? Should the latest version be the first on the list of versions?

Julian: I think latest hopefully shouldn't be an issue -- we'll have the latest version tagged also with its version number right? so it appears twice? If not (if it only appears once named latest) then I think we should fix that, we do want the version number in the UI.

Also just to be sure, it seems like maybe there's an issue here if someone adds a tag starting with a `v` -- maybe we should make a note to add validation to harness repos which fails if there's such a tag pushed.

Julian: Also clearly all the shell code would be nicer if it were built-in to bowtie :( rather than here in CI -- I wanted to do that at some point as well

OptimumCode: > If not (if it only appears once named latest) then I think we should fix that, we do want the version number in the UI.

Yes, you are right, it will only appear once as `latest` (because we don't keep a tag with version for the latest one - we would have to move it on every push to the main while the implementation is the same).
What do you think, @Julian, if we extract the version using `bowtie info` command right here in this case?

> maybe we should make a note to add validation to harness repos which fails if there's such a tag pushed

There is one case, when we actually want to push a tag with a version manually - it is when we move out a test harness which already has `matrix-versions.json` file. We would need to create a tag for each version manually (manually here means - not from GH workflow). But we could create a workflow, for example, which leaves an annotation on the commit to which the tag is attached (if this tag was pushed by someone other than GH actor). This should highlight that it might break something.

> Also clearly all the shell code would be nicer if it were built-in to bowtie

I agree. Do you think it should be something similar to `bowtie filter-implementations`? A command to show all available versions for a particular implementation. We could create an issue for that (if it is okay to implement later)

Julian: >  What do you think, @Julian, if we extract the version using bowtie info command right here in this case?

I think yeah that using `bowtie info` sounds like it might solve quite a few problems rather than calculating it externally!

> There is one case, when we actually want to push a tag with a version manually 

The kind of case I was talking about was someone creating a tag called `very-important-thing` -- which starts with a `v` but clearly isn't a version at all, but I think would cause issues on the linked line!


> Do you think it should be something similar to bowtie filter-implementations? [...] We could create an issue for that (if it is okay to implement later)

Yeah that seems reasonable -- an issue for later seems sufficient for now!

OptimumCode: > The kind of case I was talking about was someone creating a tag called very-important-thing

The problem is that we don't have a unified version format across the implementations (e.g. `blaze` impl has a commit hash as a version and `justinrainbow-json-schema` had a `dev-master` as a version until recently). So, the version can be almost anything... If we move the version collection inside the `bowtie` we maybe could try to check if the image with such a version exists at all and not consider a tag as a version if the image for it does not exist

(And I just checked - the Golang versions have `v` at the beginning... We will have tag `vv*` in this case - we probably need to check for that and remove one `v`)

Julian: That sounds perfectly fine with me!

Julian:
@OptimumCode what did we leave off on here? I'm sure it's my fault that this PR stalled (really sorry!) if you have a moment let me know what I needed to do here (was it just it's ready for a final review / merge)?

OptimumCode:
After the following change (https://github.com/bowtie-json-schema/kotlin-kmp-json-schema-validator/pull/20), the tag for the test harness will start with a `harness-release-` suffix. That should reduce the chance of collision with user-defined tags and with implementation versions that start with `v`

OptimumCode:
The PR is ready for review now. Could you please take a look, @Julian, once you have time?

.github/workflows/versioned-report.yml
```diff
@@ -26,6 +28,8 @@ jobs:
 
       - name: List all Bowtie Supported Implementations having a "matrix-versions.json" file
         id: implementations-matrix
+        env:
+          GH_TOKEN: ${{ github.token }} # for GH API access
```
Julian, line 32: Remind me, this is because in the script we have our "own" env var which is used for building the package so that we don't overlap with the `GITHUB_TOKEN` used for `github3.py` right?

Can you add a comment just reminding that here?

.github/workflows/versioned-report.yml
```diff
@@ -47,6 +51,16 @@ jobs:
             for impl in $(echo "$IMPLEMENTATIONS" | jq -r '.[]'); do
               if [ -f "implementations/$impl/matrix-versions.json" ]; then
                 MATRIX=$(echo $MATRIX | jq --arg impl "$impl" '. + [$impl]')
+              elif [ ! -d "implementations/$impl" ]; then
```
Julian, line 54: If you have time could you write up an issue ticket which describes the steps we'd need in order to remove this shell code and replace it with another bowtie subcommand? I really want to kill all these ad hoc scripts, they have already grown unmaintainable...

Julian (commented):
I haven't tried to run this, just glanced at it, but if you've run it and it works this is good to merge with me. Just left 2 minor comments.

OptimumCode:
Thank you, @Julian! I have ran it partially, will try to run the whole workflow and if all is good, I will merge the changes. I will do that closer to the end of the week (and make changes according to your comments)

.github/workflows/versioned-report.yml
OptimumCode: No, this is because the [GH CLI](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows) requires that variable to have access to the github token

OptimumCode:
Hi @Julian, I have tried to execute the whole workflow locally, and it looks like I am missing something because I am getting an error on the following command:

```
mkdir "$impl/v$version" # when dialect.json is created for each version

mkdir: kotlin-kmp-json-schema-validator: Not a directory
```

A few lines above, we copy the matrix version file to the current dir with the name equal to the implementation name

`cp $MATRIX_VERSIONS_FILE $impl`

Now, we have a **file** with the implementation name, and because of that, we cannot create a directory with the same name. But I don't see any errors in the workflow execution on GitHub, meaning it works okay there. Have you happened to know how it works on GitHub runner? Maybe we should correct the script a bit so we can execute the same code locally if needed?

.github/workflows/versioned-report.yml
OptimumCode: We already have #1928 issue to create a command to collect the available versions for the implementation. Would you like to create another issue (linked to the existing one) to replace the shell code with the new command?

Julian:
I saw it but I didn't have any idea either without trying it out myself :) at a first glance you looked right but as you say it works here so clearly we're misreading the shell code and it's doing something slightly different (which is why I always want to get rid of it and exchange it for working and tested Python code)

I don't know when I'll get a chance to run it myself, I have a huge number of things on my plate to get to -- I'd say if it's not obvious to the two of us but works here it's possibly still fine to merge, as we must be missing something simple...

OptimumCode:
Agree, let's wait until weekend - I will merge the changes and try to execute the updated workflow. If something doesn't work - I will prepare a PR with fix

OptimumCode:
I forgot I have a day off today) I have found a small issue and fixed it in #2020. The rest looks pretty good - I can see the trends chart for external test harness (https://bowtie.report/#/implementations/kotlin-kmp-json-schema-validator).

Issue #1915, opened 2025-03-26
New reports aren't being uploaded to the bowtie.report site


DannyvdSluijs:
I was going over the https://bowtie.report site and noticed the reports are from 4 days ago. Checking the GitHub [actions](https://github.com/bowtie-json-schema/bowtie/actions/workflows/report.yml) I can see the job has been failing for 4 days now.

Seems something isn't working but it was left undetected at this point. Still trying to see if more information is available but also wanted to raise awareness. Will continue to see if there is more to discover.

Julian:
It's detected I just haven't had time to spend to fix it :) I don't know yet why it started failing so definitely help welcome, the error message says a file we build has expired (so we presumably need to retrigger building it) but it's been a few months since I had to look at how that part of the CI works.

And yes again help definitely welcome, appreciate the issue!

OptimumCode:
I think the issue is related to the fact that the last versioned test report was collected 3 months ago (I have started a new collection job)

![Image](https://github.com/user-attachments/assets/3cd7625d-57f2-4b91-8481-6c83311cae53)

Based on the error below it looks possible to me

![Image](https://github.com/user-attachments/assets/67193d83-bceb-4113-83d2-375404e22165)

OptimumCode:
It also looks like `java-json-schema` test harness has a `matrix-versions.json` file, but no versioned images were built for that test harness (because they are built only when the workflow is triggered manually for a specific test harness).

@Julian do you mind if I trigger a workflow for `java-json-schema`?

Julian:
Definitely don't mind!

OptimumCode:
The versioned test results are [collected successfully](https://github.com/bowtie-json-schema/bowtie/actions/runs/14108428130). Let's see if the test report workflow passes as well

OptimumCode:
The workflow is green and the report should be updated now. @DannyvdSluijs could you please also confirm that everything looks okay now?

OptimumCode:
> I'm curious to see if it would now update every six hours as defined in de report.yml

All the latest scheduled jobs look okay. And UI shows it was updated 2 hours ago (for your PR to have effect, a new image must be published for the test harness - I have triggered the build, and we should see the correct version after the next report collection).
![Image](https://github.com/user-attachments/assets/1db70f9a-d0e2-4720-9352-74d5084c0254)

> Perhaps we can have at least a schedule defined in versioned-report.yml

[...]

PR #1920, opened 2025-03-28
Add schedule trigger for versioned report workflow

OptimumCode:
PR adds a schedule trigger to the [versioned-report.yml](https://github.com/bowtie-json-schema/bowtie/blob/main/.github/workflows/versioned-report.yml) workflow (as suggested in #1915).

The automatic build is scheduled for Sunday at 23:30 (45 minutes before the report workflow is scheduled) to provide the fresh artifacts with versioned reports (the latest versioned report runs took about 40 minutes, so I have added 5 more minutes just in case).

I don't think there is a point in triggering the workflow more often than once a week - we don't have many test harnesses with different versions' support

----
📚 Documentation preview 📚: https://bowtie-json-schema--1920.org.readthedocs.build/en/1920/

### DIFF
--- a/.github/workflows/versioned-report.yml
+++ b/.github/workflows/versioned-report.yml
@@ -3,6 +3,9 @@
 name: Collect New Versioned Test Results
 
 on:
+  schedule:
+    # every Sunday at 23:30 UTC (45 minutes before a report workflow is scheduled)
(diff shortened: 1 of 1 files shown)


Julian (approved):
Looks perfect thanks!

PR #1922, opened 2025-03-29
Provide GITHUB_TOKEN during test suite execution

OptimumCode:
The PR adds a GITHUB_TOKEN into env variables during test suite execution. There is a chance that without a token we hit a rate limit. Example here:
https://github.com/bowtie-json-schema/bowtie/actions/runs/14143123995/job/39627152056

PR also reduces the scope of write permissions only to the job responsible for site deployment

----
📚 Documentation preview 📚: https://bowtie-json-schema--1922.org.readthedocs.build/en/1922/

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -19,8 +19,6 @@ on:
 
 permissions:
   contents: read
-  pages: write
-  id-token: write
 
 concurrency:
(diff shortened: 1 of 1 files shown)


Julian (approved):
Seems reasonable

PR #1950, opened 2025-04-12
Explicit type parameters in factory function

OptimumCode:
The new version of `pyright` checks that the function return type matches the defined type. For that we need to specify type parameters in the factory function.

We didn't ignore the error because this new check is added to existing one and ignoring the new error would mean we completely disable the type checking for those fields

----
📚 Documentation preview 📚: https://bowtie-json-schema--1950.org.readthedocs.build/en/1950/


Julian:
Thanks!!

PR #2085, opened 2025-07-08
Adds swift-json-schema Implementation

ajevans99:
Add new implementation for [`swift-json-schema`](https://github.com/ajevans99/swift-json-schema).

----
📚 Documentation preview 📚: https://bowtie-json-schema--2085.org.readthedocs.build/en/2085/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -339,6 +339,11 @@ updates:
       interval: daily
 
   - package-ecosystem: "docker"
+    directory: "/implementations/swift-swift-json-schema"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "docker"
     directory: "/bowtie/tests/fauxmplementations/envsonschema"
     schedule:
       interval: "daily"


.pre-commit-config.yaml
```diff
@@ -114,3 +114,12 @@ repos:
             "-d",
             "implementations",
           ]
+  - repo: https://github.com/swiftlang/swift-format
```
ajevans99, line 117: Hmm... looks like this hook keeps timing out. I tried both Apple's `swift-format` and also `[swiftlint`](https://github.com/realm/SwiftLint) but both timeout. Likely due to dependency on `swift-syntax` which is a pretty heavy package. Is there an easy way to increase the timeout?

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/BowtieProcessor.swift
```diff
@@ -0,0 +1,103 @@
...
+    let identifier = run.seq
+    log("Schema: \(run.testCase.schema)")
+
+    let rawTestSchema = run.testCase.schema
+    let schema: Schema
+    do {
+      schema = try Schema(
+        rawSchema: rawTestSchema,
+        context: .init(
+          dialect: .draft2020_12,
```
Julian, line 62: Even though you only (currently) support draft 2020-12 it seems a bit more future-proof to (once you follow the above on dialect requests) use the dialect which was sent previously here rather than hard-coding 2020-12.

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/Models/Response.swift
```diff
@@ -0,0 +1,143 @@
...
+      #if os(Linux)
+        return "Linux"
+      #elseif os(macOS)
+        return "macOS"
+      #else
+        return "Unknown"
+      #endif
+    }()
+    let osVersion = ProcessInfo.processInfo.operatingSystemVersionString
+    let languageVersion: String = {
```
Julian, line 29: Is there a reason to throw away the granularity here on what specific version it is?

Julian (commented):
Thanks for sending this (and congrats on your implementation)!

I didn't actually try to run it yet, but I left a few minor comments. Let me know if any aren't clear, but overall I think this looks great!

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/Models/Response.swift
ajevans99: The compiler directives are just booleans and added for each minor version of Swift. Technically, could add 6.1 here since that is the docker version. Or could do a similar trick to `jsonSchemaVersion` in `Version.swift` and codgen a swift version variable before the docker build.

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/BowtieProcessor.swift
```diff
@@ -0,0 +1,103 @@
...
+      version: start.version,
+      implementation: .init(version: jsonSchemaVersion, dialects: Constants.supportedDialects)
+    )
+    try write(response)
+  }
+
+  mutating func handleDialect(_ dialect: Dialect) throws {
+    assert(didStart)
+    log("Dialect: \(dialect.dialect)")
+    try write(DialectResponse(ok: Constants.supportedDialects.contains(dialect.dialect)))
```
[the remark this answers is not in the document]
ajevans99: Ah, looks like I started this and forgot about it with the unused `private var currentDialect: JSONSchema.Dialect?`. How's this look?

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/Models/Response.swift
Julian: Aha got it, basically it's up to you of course but my recommendation is to use as granular of a version here as possible such that a user seeing some unexpected behavior would never suspect that there's simply some obscure Swift (language) bug in the version they're using, which usually means as much info as possible included, but of course if that's lots of extra work the likelihood it ever matters is small.

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/BowtieProcessor.swift
```diff
@@ -0,0 +1,103 @@
...
+          )
+        )
+      )
+      return
+    }
+
+    let testResults = run.testCase.tests
+      .map { test in
+        let instance = test.instance
+        let result = schema.validate(instance)
```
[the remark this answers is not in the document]
Julian: Got it, awesome!

implementations/swift-swift-json-schema/Sources/BowtieJSONSchema/BowtieProcessor.swift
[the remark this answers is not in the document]
Julian: Awesome that looks perfect thanks!

.pre-commit-config.yaml
ajevans99: Unfortunately couldn't find a good way to pull a `swift-format` binary without some fairly complex bash script branching on OS and architectures. I added the `ci: skip` for now.

I'm able to run pre-commit locally without issue, although the first run did take quite a while.

```bash
➜  swift-swift-json-schema git:(swift-json-schema) pre-commit run swift-format --all-files
swift-format (swift implementations).....................................Passed
```

Julian:
Looks great, let's merge and see how it goes. Thanks again!

PR #2088, opened 2025-07-21
Bump Blaze pinned commit SHA

jviotti:
To fix a little issue discovered by the addition of a new test case in the test suite. See https://github.com/sourcemeta/blaze/pull/479


jviotti:
Oh interesting. Let me know how it goes!

PR #2183, opened 2025-09-12
add ajv-plus implementation

bavulapati:
- [ajv-plus](https://github.com/theflashlabs/ajv-plus)
- Ajv-Plus wants to be actively maintained. Support the upcoming JSON Schema specification version too.
- Ajv-Plus wants to be 100% compliant with the standard JSON Specification.

----
📚 Documentation preview 📚: https://bowtie-json-schema--2183.org.readthedocs.build/en/2183/

### DIFF
--- a/implementations/js-ajv-plus/bowtie_ajv_plus.js
+++ b/implementations/js-ajv-plus/bowtie_ajv_plus.js
@@ -64,12 +64,12 @@ const cmds = {
 
     // For some reason ajv's process for Draft 6 is different, so split.
     if (args.dialect !== "http://json-schema.org/draft-06/schema#") {
-      ajv = new DRAFTS[args.dialect]({ strict: false });
+      ajv = new DRAFTS[args.dialect]();
     } else {
       const Ajv = require("@theflashlabs/ajv-plus");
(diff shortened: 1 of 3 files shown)


implementations/js-ajv-plus/bowtie_ajv_plus.js
```diff
@@ -0,0 +1,128 @@
...
+    console.assert(started, "Not started!");
+
+    // For some reason ajv's process for Draft 6 is different, so split.
+    if (args.dialect !== "http://json-schema.org/draft-06/schema#") {
+      ajv = new DRAFTS[args.dialect]({ strict: false });
+    } else {
+      const Ajv = require("@theflashlabs/ajv-plus");
+      const draft6MetaSchema = require("@theflashlabs/ajv-plus/dist/refs/json-schema-draft-06.json");
+
+      ajv = new Ajv({ strict: false });
```
Julian, line 72: I see you still explicitly passing `strict: false` here. Isn't that the core thing you were saying ajv-plus changes about ajv's defaults?

The harness is meant to check implementations in their default configuration (at least until we ever do #8, in which case we'd also add ajv with strict mode off).

PR #2287, opened 2025-11-18
feat(js-ajv): add standard ajv-formats with ajv

markrzen:
AJV is often paired up with the recommended ajv-formats:
https://ajv.js.org/packages/ajv-formats.html

I noticed AJV has MANY failures, and some of them are because formatting isn't supported out-of-the-box. This brings AJV more in parity with other libraries.

----
📚 Documentation preview 📚: https://bowtie-json-schema--2287.org.readthedocs.build/en/2287/

### DIFF
--- a/implementations/js-ajv/bowtie_ajv.js
+++ b/implementations/js-ajv/bowtie_ajv.js
@@ -2,6 +2,7 @@ const path = require("path");
 const readline = require("readline");
 const os = require("os");
 const process = require("process");
+const addFormats = require("ajv-formats")
 
 const ajv_version = require(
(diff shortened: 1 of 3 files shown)


Julian:
We don't actually yet run any of the optional format tests which would be the ones ajv-format would be relevant for. That's tracked in #36.

More broadly though, we definitely would want to show what compliance looks like with various configuration or additional dependencies but the initial goal was to show what all validators do out of the box with as minimal setup as possible. 

Being able to report on variations is tracked in #8 and definitely would be nice to have, but essentially right now we avoid as much tweaking as possible when setting up a test harness.

PR #2296, opened 2025-11-30
add perl-json-schema-modern implementation

fperrad:
----
📚 Documentation preview 📚: https://bowtie-json-schema--2296.org.readthedocs.build/en/2296/

### DIFF
--- a/implementations/perl-json-schema-modern/bowtie_runner.pl
+++ b/implementations/perl-json-schema-modern/bowtie_runner.pl
@@ -1,0 +1,106 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+use JSON::Schema::Modern;
+use Mojo::JSON qw(decode_json encode_json false true);
+use Try::Tiny;
+
+my %dialect = (
+    'https://json-schema.org/draft/2020-12/schema' => 'draft2020-12',
+    'https://json-schema.org/draft/2019-09/schema' => 'draft2019-09',
+    'http://json-schema.org/draft-07/schema#'      => 'draft7',
+    'http://json-schema.org/draft-06/schema#'      => 'draft6',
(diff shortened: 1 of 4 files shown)


implementations/perl-json-schema-modern/bowtie_runner.pl
```diff
@@ -0,0 +1,86 @@
...
+    },
+    dialect => sub () {
+        my $request = shift;
+        die 'Not started!' unless $started;
+        if ( exists $dialect{ $request->{dialect} } ) {
+            $schema = $dialect{ $request->{dialect} };
+            return { ok => true };
+        }
+        else {
+            return { ok => false };
```
Julian: This code path would be a bug (either in this harness' declared dialects or in Bowtie itself) so crashing here with an error is usually suggested.

implementations/perl-json-schema-modern/bowtie_runner.pl
```diff
@@ -0,0 +1,86 @@
...
+        my $request = shift;
+        die 'Not started!' unless $started;
+        my $js = JSON::Schema::Modern->new( specification_version => $schema );
+        my $case = $request->{case};
+        while ( my ( $url, $content ) = each %{ $case->{registry} } ) {
+            $js->add_schema( $url, $content );
+        }
+        my @results = ();
+        foreach my $test ( @{ $case->{tests} } ) {
+            push @results, $js->evaluate( $test->{instance}, $case->{schema} );
```
Julian: Can you add error handling here to ensure that input which causes the implementation to crash does not crash the whole harness? You should return an error response in that case - https://docs.bowtie.report/en/stable/implementers/#step-5-handling-errors has details albeit not for perl's mechanism.

Julian:
Thanks, looks great, appreciated!

PR #2309, opened 2025-12-11
Bump ruby-json_schemer version

davishmcclurg:
Add 2.4.0 and 2.5.0 to versions matrix and set 2.5.0 as default.


Julian:
Yeah I think you're right there on it being confused by the dockerfile argument. But thanks for the bump.

PR #2357, opened 2026-01-14
fix: handle GitHub API errors gracefully in 'info --versions'

### LINKED ISSUE #1519: `bowtie info --versions` produces an ugly error message when unauathenticated
The fact that it requires authentication at all is a GitHub API bug (which they know about), but we need to cope with that.

AnujAga2005:
Closes #1519

### Problem
Previously, running `bowtie info --versions` would crash with a traceback if the GitHub API request failed due to missing authentication tokens.

### Solution
I've wrapped the GitHub API iteration in `_core.py` with `suppress(GitHubError)`. Now, if the API call fails, Bowtie gracefully skips the remote version tags and displays the rest of the implementation info using the locally available version data.

### Testing
Added `tests/test_info_versions.py` which mocks the GitHub client and forces a `GitHubError` to verify that the function returns the fallback version instead of raising an exception.

### DIFF
--- a/bowtie/_core.py
+++ b/bowtie/_core.py
@@ -549,7 +549,7 @@ class Implementation:
         url = CONTAINER_PACKAGES_API / self.id / "versions"
 
         gh = github()
-        
+
         versions: Set[str] = (
             {self.info.version} if self.info.version else set()
         )
--- a/bowtie/tests/test_info_versions.py
+++ b/bowtie/tests/test_info_versions.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 import sys
 
+
 def test_info_versions_survives_no_credentials():
     """
     Integration test: Ensure 'bowtie info --versions' does not crash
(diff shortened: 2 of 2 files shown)


AnujAga2005:
Understood. Whenever you have a suggestion for the testing pattern, I'm happy to update the code. Thanks!

Julian:
This seems like testing it via an integration test which calls the CLI while *not* preserving the shell environment should be sufficient to do this without a mock, and would be overall a better / more realistic test anyhow.

Julian:
Sorry, I'm honestly not very interested in having a conversation with your LLM. If you're looking to help out here please invest brainpower to take what you're contributing and think carefully about it before sending it at me.

AnujAga2005:
@Julian  if u allow , can i take up other issues and start fresh? i won't be taking help from any LLM for any code or comment whatsoever and do everything by myself.

Julian:
Work which is your own, which you have thought hard about and attempted to make work is welcome.

PR #2377, opened 2026-01-30
Feat: disallow branch names as implementation versions

### LINKED ISSUE #1517: `bowtie smoke` (and the JSON Schema for implementation info) should disallow versions containing `master / main / trunk`
These branch names aren't helpful, because a user seeing them on the site will know that it was a branch, but not *what commit that was* at the time the report was built.

(See #1298 and https://github.com/bowtie-json-schema/bowtie/pull/1512#discussion_r1754866334 for cases this smoke test would have / should have caught).

codefromlani:
closes #1517

Prevent branch names such as main, master, or trunk from being used as implementation versions.

### DIFF
--- a/bowtie/tests/fauxmplementations/lintsonschema/schemas/models/implementation.json
+++ b/bowtie/tests/fauxmplementations/lintsonschema/schemas/models/implementation.json
@@ -32,7 +32,7 @@
       "type": "string",
 
       "not" : {
-        "pattern": "(?i)(^|[^a-z])(master|main|trunk)([^a-z]|$)"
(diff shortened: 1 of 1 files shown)


bowtie/schemas/models/implementation.json
```diff
@@ -29,7 +29,11 @@
 
       "$id": "tag:bowtie.report,2024:models:implementation:version",
 
-      "type": "string"
+      "type": "string",
+
+      "not" : {
+        "pattern": "(?i)(^|[^a-z])(master|main|trunk)([^a-z]|$)"
```
Julian: Just using `enum` seems like it probably should be fine here.

PR #2401, opened 2026-02-12
Add M3 JSON Schema validator (Clojure + ClojureScript)

JulesGosnell:
## Summary

Adds [M3](https://github.com/JulesGosnell/m3), a cross-platform JSON Schema validator written in Clojure/ClojureScript, as two Bowtie implementations:

- **clojure-m3** — JVM (Clojure), built as an uberjar via Leiningen
- **clojurescript-m3** — Node.js (ClojureScript), compiled via shadow-cljs

M3 supports all seven JSON Schema drafts: draft-03, draft-04, draft-06, draft-07, draft-2019-09, draft-2020-12, and draft-next (though draft-next is not declared to Bowtie since it isn't a standard dialect URI).

Both implementations support the full IHOP v1 protocol including `registry` for `$ref` resolution.

### Tested locally

```
$ echo '{"cmd":"start","version":1}
{"cmd":"dialect","dialect":"https://json-schema.org/draft/2020-12/schema"}
{"cmd":"run","seq":1,"case":{"schema":{"type":"string"},"tests":[{"description":"a string","instance":"hello"},{"description":"a number","instance":42}]}}
{"cmd":"run","seq":2,"case":{"schema":{"$ref":"http://example.com/s"},"registry":{"http://example.com/s":{"type":"integer"}},"tests":[{"description":"valid int","instance":7},{"description":"invalid string","instance":"nope"}]}}
{"cmd":"stop"}' | docker run --rm -i bowtie-m3-jvm

{"version":1,"implementation":{"language":"clojure","name":"m3","version":"1.0.0-beta3",...}}
{"ok":true}
{"seq":1,"results":[{"valid":true},{"valid":false}]}
{"seq":2,"results":[{"valid":true},{"valid":false}]}
```

ClojureScript output is identical.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/implementations/clojure-m3/Dockerfile
+++ b/implementations/clojure-m3/Dockerfile
@@ -1,0 +1,12 @@
+FROM alpine AS builder
+RUN apk add --no-cache leiningen
+WORKDIR /usr/src/app
+COPY project.clj /usr/src/app/
+RUN lein deps
+COPY . .
+RUN mv "$(lein uberjar | sed -n 's/^Created \(.*standalone\.jar\)/\1/p')" app-standalone.jar
+
+FROM bellsoft/liberica-openjdk-alpine:25
+WORKDIR /usr/src/app
+COPY --from=builder /usr/src/app/app-standalone.jar .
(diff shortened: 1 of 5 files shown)


Julian:
@JulesGosnell just making sure you're still interested in the above

Issue #703, opened 2023-12-12
Add a way for implementations to request copies of the metaschema(s)


Julian:
Some tests in the test suite rely on the presence of one or more metaschemas (i.e. they include `$ref: someDialectURI`).

Strictly speaking according to the spec that isn't *required* of implementations (to have a copy of the metaschema stored internally such that anyone can use such a schema without otherwise registering the metaschema). But it's strongly encouraged -- c.f. json-schema-org/website#118.

But nearly all implementations *do* include it.

A very small number however do not -- and in their own runnings of the test suite they seem to custom inject those schemas in for some reason.

In order to support such a setup ( :/ ) we could add something like a `include-metaschemas: true` (perhaps in the `dialect` request?) which signals an implementation wants Bowtie to send along the metaschema in registries?

Bit more thinking required, but filing this to track that at least one implementation fails more tests than they in theory could.

PR #2436, opened 2026-02-27
python-jsu: json-schema-utils implementation for python

zx80:
The [JSU compiler](https://github.com/zx80/json-schema-utils) converts a [schema](https://json-schema.org/) to a [model](https://json-model.org/) internally and then uses [jmc](https://json-model.org/#/JMC) as a backend to generate a validator in C, Python, JS, Java, PL/pgSQL or Perl.

This implementation run tests with the generated dynamic Python code.

### DIFF
--- a/implementations/python-jsu/Dockerfile
+++ b/implementations/python-jsu/Dockerfile
@@ -4,27 +4,21 @@ FROM alpine:3.23
 # without recompiling from sources, whereas the base alpine image has it,
 # so start from alpine.
 
-RUN mkdir /app /app/schema-cache-by-hashed-urls
-WORKDIR /app
+RUN mkdir -p /usr/src/myapp
+WORKDIR /usr/src/myapp
 
 # allow to install from package (not set) or build from sources (branch or commit)
 ARG JMC
 ARG JSU
 
-# system site package dependencies
 RUN apk add git py3-pip py3-re2 icu-data-full
 
-# setup Python virtual environment
-ENV PATH=/venv/bin:$PATH
-RUN python -m venv /venv --system-site-packages
-RUN pip install jsonschema-specifications
-
+# force install, otherwise it would require a virtual environment
+RUN pip install --break-system-packages jsonschema-specifications
 RUN if [ "$JMC" ] ; then jmc="git+https://github.com/clairey-zx81/json-model@$JMC" ; fi ; \
-    pip install "${jmc:-json_model_compiler}"
-
+    pip install --break-system-packages "${jmc:-json_model_compiler}"
 RUN if [ "$JSU" ] ; then jsu="git+https://github.com/zx80/json-schema-utils@$JSU" ; fi ; \
-    pip install "${jsu:-json_schema_utils}"
+    pip install --break-system-packages "${jsu:-json_schema_utils}"
 
-# harness script
 COPY bowtie_jsu_python.py .
 CMD ["python3", "./bowtie_jsu_python.py"]
--- a/implementations/python-jsu/bowtie_jsu_python.py
+++ b/implementations/python-jsu/bowtie_jsu_python.py
@@ -11,6 +11,7 @@ from pathlib import Path
 import hashlib
 import json
 import platform
+import shutil
 import sys
 import traceback
 
@@ -37,7 +38,7 @@ VERSIONS: dict[str, int] = {
 }
 
 # cache is used for registry and meta schemas
-CACHE: str = "./schema-cache-by-hashed-urls"
+CACHE: Path = Path(__file__).parent / "schema-cache-by-hashed-urls"
 
 # version for both front-end and back-end
 JSU_VERSION: str = (
@@ -97,12 +98,14 @@ class Runner:
 
         case = req["case"]
         assert isinstance(case, dict), "case is an object"
-        description = case.get("description")
-        assert description is None or isinstance(description, str)
+        jschema = case["schema"]
+        assert isinstance(jschema, (bool, dict)), "boolean or object schema"
         tests = case["tests"]
         assert isinstance(tests, list), "tests is a list of instances"
+        description = case.get("description")
+        assert description is None or isinstance(description, str)
 
-        files: list[str] = []
+        CACHE.mkdir(exist_ok=True)
         results: JsonArray = []
 
         try:
@@ -112,14 +115,12 @@ class Runner:
                     for url, schema in reg.items():
                         # use truncated hashed url as filename
                         uh = hashlib.sha3_256(url.encode()).hexdigest()[:16]
-                        fn = f"{CACHE}/{uh}.json"
-                        files.append(fn)
-                        with Path.open(fn, "w") as fp:
+                        with Path.open(CACHE / f"{uh}.json", "w") as fp:
                             json.dump(schema, fp)
 
             # compile schema to python
             checker = json_schema_to_python_checker(
-                case["schema"],
+                jschema,
                 description,
                 cache=CACHE,
                 version=self.version,
@@ -131,17 +132,15 @@ class Runner:
         except Exception:  # an internal error occurred
             return {
                 "errored": True,
-                "seq": req.get("seq"),
+                "seq": req["seq"],
                 "context": {"traceback": traceback.format_exc()},
             }
 
-        finally:
-            # remove registered files
-            for fn in files:
-                Path(fn).unlink()
+        finally:  # wipe out cache to avoid state leaks
+            shutil.rmtree(CACHE)
 
         return {
-            "seq": req.get("seq"),
+            "seq": req["seq"],
             "results": results,
         }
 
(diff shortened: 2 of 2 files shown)


implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,175 @@
+#! /usr/bin/env python3
+#
+# Implement the bowtie protocol against json-schema-utils schema
+# validator based on the dynamic json-model-compiler python backend.
+#
+
+# stupidly sorted imports
```
Julian: Please remove this comment and the ones below which don't really add anything.

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,175 @@
+#! /usr/bin/env python3
+#
+# Implement the bowtie protocol against json-schema-utils schema
```
Julian:
```suggestion
"""
A Bowtie harness for the json-schema-utils schema validator using
the dynamic json-model-compiler Python backend.
"""
```

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,175 @@
...
+    version: int | None = None
+
+    # count input lines for some error messages
+    line: int = 0
+
+    def cmd_start(self, req: Jsonable) -> Jsonable:
+        assert req.get("version", 1) == 1, "expecting protocol version 1"
+
+        # version for both front-end and back-end
+        jsu_version = (
```
Julian: (This can just be a global.)

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,175 @@
...
+        # put registry in cache
+        registry = case.get("registry", {})
+        if isinstance(registry, dict):  # may be set to null
+            for url, schema in registry.items():
+                uh: str = hashlib.sha3_256(url.encode()).hexdigest()[:16]
+                with Path.open(f"{CACHE}/{uh}.json", "w") as fp:
+                    json.dump(schema, fp)
+
+        # TODO what about skipping unsupported cases?
+        # FIXME there is no clean way to identify them
```
Julian: Can you elaborate on what is missing?

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,175 @@
...
+        """Runner purpose is to run."""
+        # request/response protocol is to receive and send one-line jsons
+        for line in sys.stdin:
+            self.line += 1
+            try:
+                req = json.loads(line)
+                assert isinstance(req, dict), "input must be a json object"
+                res = self.process(req)
+            except Exception as e:
+                res = {
```
Julian: (Same as above, you should just be crashing here.)

implementations/python-jsu/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    python -m venv /venv --system-site-packages && \
+    pip install "$jmc" && \
+    pip install "$jsu" && \
+    apk del git && \
+    apk cache clean
+# no: rm -rf /root/.cache
+
+# cache meta schemas
+RUN apk add curl openssl && \
+    for url in \
```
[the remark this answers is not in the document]
zx80: - Indeed there is no copy of the meta schemas coming with the implementation. I added them because otherwise some tests were failing, probably because the execution environment is prevented from downloading. It is indeed the case described in #703. I'd _really_ prefer not to add copies in the implementation which does not need it in any way, even if  the "JSON Schema team" recommends it. I see good reasons _not_ to add them, eg if they are improved somehow then we get the updated version.
- About the suggested pypi project: it still requires some mapping one way or another, thus implies bowtie-specific integration work, so I'm _really_ fine with the current direct cache injection method which needs nothing specific, unless this is a blocker.

implementations/python-jsu/Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM alpine
```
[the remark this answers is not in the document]
zx80: IMO this is harmless and means that less updates would be needed, especially as python is a base/basic package, but I can do it nevertheless.

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,175 @@
...
+        return {"ok": self.version != 0}
+
+    def cmd_run(self, req: Jsonable) -> Jsonable:
+        """Run one case and its tests."""
+
+        case = req["case"]
+        assert isinstance(case, dict)
+        description = case.get(
+            "description",
+            f"case from input line {self.line}",
```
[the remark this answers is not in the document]
zx80: Ok.

implementations/python-jsu/Dockerfile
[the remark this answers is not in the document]
Julian: Follow what you see please, consistency and reproducibility is more important.

implementations/python-jsu/bowtie_jsu_python.py
Julian: Yes it sounds like you're looking for https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/698

> The fact that non supported features coldly fail seems a reasonable behavior anyway.

I agree. Please remove the comments then.

implementations/python-jsu/bowtie_jsu_python.py
zx80: Ok, I understand that It must voluntary crash on errors, which is kind of an anathema.

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,170 @@
...
+                for url, schema in registry.items():
+                    uh: str = hashlib.sha3_256(url.encode()).hexdigest()[:16]
+                    with Path.open(f"{CACHE}/{uh}.json", "w") as fp:
+                        json.dump(schema, fp)
+
+            # compile schema to python
+            checker = json_schema_to_python_checker(
+                case["schema"],
+                description,
+                cache=CACHE,
```
Julian: Does your implementation modify the cache it's given? If so you are going to leak state across test runs. Do you have a way to not do so in your implementation?

implementations/python-jsu/bowtie_jsu_python.py
```diff
@@ -0,0 +1,170 @@
...
+                version=self.version,
+            )
+
+            # apply to test vector
+            results = [
+                {"valid": checker(test["instance"])} for test in case["tests"]
+            ]
+
+            return {
+                "seq": req.get("seq", f"input line {self.line}"),
```
Julian: Please remove the default here and everywhere, you should rely on the protocol being valid that you're speaking; if it isn't, we do not want to proceed, we want to crash. Errors should not pass silently.

implementations/python-jsu/bowtie_jsu_python.py
zx80: Ok.

implementations/python-jsu/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    else \
+        jmc=json_model_compiler ; \
+    fi && \
+    if [ "$JSU" ] ; then \
+        jsu="git+https://github.com/zx80/json-schema-utils@$JSU" ; \
+    else \
+        jsu=json_schema_utils ; \
+    fi && \
+    apk add git py3-pip py3-re2 icu-data-full && \
+    python -m venv /venv --system-site-packages && \
```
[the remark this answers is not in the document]
zx80: I actually tried that first but because of the dependency on `re2`, which is **not** available without recompiling the packages from source on the python docker images, which required more packages for building the stuff (c++ compiler, headers for various libraries…).

Thus the current situation looks like a reasonable compromise.

If it starts from the python image, I'm afraid you might not like the installation process, but I can do it as I already did it before being unconvinced and simplifying the stuff to the above.

Given the above information about the previous solution, please tell me what you prefer.

implementations/python-jsu/Dockerfile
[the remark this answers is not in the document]
zx80: Ok, I put "schema-cache-by-hashed-urls" to have "cache" in the directory name.

implementations/python-jsu/bowtie_jsu_python.py
Julian: Yeah the concern would be test `A` says that `urn:example:foo` is `<some schema>`, and then later in the run test B is testing that `urn:example:foo` does not exist. The official test suite does not do this but Bowtie is meant to be more general precisely to support richer tests.

implementations/python-jsu/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# allow to install from package (not set) or build from sources (branch or commit)
+ARG JMC
+ARG JSU
+
+# system site package dependencies
+RUN apk add git py3-pip py3-re2 icu-data-full
+
+# setup Python virtual environment
+ENV PATH=/venv/bin:$PATH
+RUN python -m venv /venv --system-site-packages
```
Julian: No other Dockerfile uses a virtualenv, please follow that. It's possible we'd change that at some point, if so we'd do it for all images.

PR #2453, opened 2026-03-06
update jsonschema.net docker image

gregsdennis:
Updates the docker image to run .Net 10.  Not sure if this will help the two failures, but I don't have docker on my machine, and this is the only thing I can't run.  Everything else runs fine and all tests pass.


gregsdennis:
Hmmm.  Looks like it's still trying to run .net 9.  I'll check it out.

PR #2454, opened 2026-03-08
feat(cli): add --failures-only flag to smoke

### LINKED ISSUE #1318: Ensure it's easy to run `smoke` on many implementations and only see output for failing ones
There should be an easy workflow way to do "run smoke across many implementations and only see failures".

For `--format json` this likely just means ensuring there's a decent `jq` query on what we output.

But for "pretty" output we may need a CLI option, or else simply to ensure this is easily doable.

abhi-03-kh:
Adds a --failures-only flag to `bowtie smoke`.

When used with pretty or markdown output, only implementations that
fail smoke tests are displayed. JSON output remains unchanged so
machine consumers can filter results using tools like jq.

Fixes #1318.

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -2422,19 +2422,18 @@ async def smoke(
 
     filtered = (
         [(id, info, r) for id, info, r in results if not r.success]
-        if failures_only and format != "json"
+        if failures_only
         else results
     )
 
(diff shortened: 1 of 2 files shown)


abhi-03-kh:
@Julian,  any feedback on this when you have time?

PR #2479, opened 2026-03-23
Add JSON Schema Utils (JSU) compiler with Java, JS and Perl backends

zx80:
They use the same harness as `python-jsu` and `c-jsu` but distinct containers.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -148,6 +148,9 @@ updates:
   # ignore: python-jsonschema - installs the latest (unpinned) version
   # ignore: python-jsu - installs the latest (unpinned) version
   # ignore: c-jsu - installs the latest (unpinned) version
+  # ignore: java-jsu - installs the latest (unpinned) version
+  # ignore: js-jsu - installs the latest (unpinned) version
+  # ignore: perl-jsu - installs the latest (unpinned) version
 
   - package-ecosystem: "bundler"
     directory: "/implementations/ruby-json_schemer"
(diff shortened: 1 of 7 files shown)


Julian:
@zx80 builds of JSU seem to fail for all backends. Can you have a look at e.g. https://github.com/bowtie-json-schema/bowtie/actions/runs/24770095765

zx80:
I pushed a new version yesterday :cry:

I'm investigating.

It seems that bowtie does not want anything on stderr, is this a change?
I do have warnings generated by the compiler.

Julian:
No that will always have been the case, Bowtie will not be able to tell if the harness has crashed and what's there is a traceback or something. If they're expected just silencing them in the harness should be ok.

zx80:
Work in progress:

- **c-jsu**: works fine.
- **python-jsu**: the github run failed while installing qemu because some package could not be downloaded, this is unrelated to the tool, and it seems to work out of the box.
- **js-jsu**: my bad, added dependency, easy fix that should work for future dependencies as well.
- **perl-jsu**: my bad, added dependencies. I have to install of evolving perl modules with alpine as they do not seem to be available in existing packages. The dependencies are related to value validators.
- **java-jsu**: It works with interactive testing. I still have to understand why it generates a stderr output while it does not on other versions.

I'll submit a patch when the 3 containers are fixed, or I can submit one for the first 2.

Julian:
Thanks, feel free to send a patch when it's all fixed, for now I've temporarily removed them to get things working again. When you send your PR we'll also have to measure how long they take to run -- as-is the versions that we had here were more than 10x slower than any other implementation (including others which compile schemas) which completely blew up the time it takes to collect new test results. Hopefully once things are working correctly that will come down more towards other implementations.

Looking forward to the PR.

zx80:
> Thanks, feel free to send a patch when it's all fixed, for now I've temporarily removed them to get things working again. When you send your PR we'll also have to measure how long they take to run -- as-is the versions that we had here were more than 10x slower than any other implementation (including others which compile schemas) which completely blew up the time it takes to collect new test results. Hopefully once things are working correctly that will come down more towards other implementations.
> 
> Looking forward to the PR.

I'm working on fixing the java docker.

About the slowness, I think that it comes from:

1. The registry cache is fully reset on each test case, involving writing dozens of meta schema files. This can be improved by writting the spec files once instead.
2. It is a compiler: the schema is converted to an intermediate form (one frontend command), then to source code in the target language (one backend command), which may need to be compiled (C, Java) or interpreted (Python, Perl, JS). Each time several processes and files are involved even for the most trivial schema, which is quite different compared to other tools which interpret schemas directly inside the harness process. The benefit of the compilation approach is that the resulting validation code is really fast, but it takes some time to build. Obviously, this process is quite inefficient when running over hundreds of very small schemas.

zx80:
See #2564 PR for reinstating JSU tests.

PR #2519, opened 2026-04-09
introduce: implementation json-schema-library (typescript)

sagold:
Hey there,

this pull request adds **bowtie integration** for [json-schema-library](https://github.com/sagold/json-schema-library).

The library supports draft4 to draft2020-12 validation and utilities and is already registered on JSON Schema Tools:
[https://json-schema.org/tools](https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=toolingTypes&licenses=&languages=&drafts=&toolingTypes=&environments=)

- the implementation is prefixed with `ts` for typescript, but also can be consumed using javascript only
- the implementation is small. Still, there are two files where `bowtie-api.ts` mainly contains types for the bowtie integration
- the installation is small. Installing only the implementation and one dependency to format the source-code as requested
- I am using `node:24-alpine` like other javascript implementations
- I added prettier and an editor-config for autoformatting the source code. prettier can be run with `npm run format`
- When running docker the latest `json-schema-library` version is downloaded (starting with v11.2.0 which supports bowtie). Dependabot entries for docker and npm have been added for _ts-json-schema-library_
Please let me know if something is amiss or requires changing.

Thank you for your efforts - highly appreciated!
Sascha


implementations/ts-json-schema-library/bowtie-jlib.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        return { ok: true };
+    },
+
+    run: async (args) => {
+        console.assert(state === "dialect" || state === "testing");
+        state = "testing";
+        const { tests, registry } = args.case;
+        let { schema } = args.case;
+        if (schema != null && typeof schema === "object") {
```
Julian: In general a harness shouldn't contain logic like this -- we want to be able to see what an implementation does *by default* and with minimal configuration. 

Here it would seem you're adding `$schema` properties to all incoming schemas, which makes asking "does this implementation do the right thing for the schema `{}`" now effectively "lie", since you never run that schema, you silently run a different one.

If you don't support implicit dialect schemas (i.e. if `$schema` is required for you to run) you should be returning `ok: false` in your dialect response and then still not modifying schemas here.

sagold: Hi Julian,

thank you for your review and feedback! I know what you mean with "lie" - that is not the case here though. Maybe you have suggestion to improve my api or the testing-harness:

`json-schema-library` 

- includes all draft-version simultaneously for cross-draft support of remote references
- the corresponding draft-version to use is specified by the required `$schema`-keyword

The `test harness` works differently though:

- it specifies the version once in the dialect command 
- it specifies a custom meta-schema once in one of the tests

So I pass in the required $schema version from the dialect command (which a normal consumer would either do directly on the schema or using the draft-options)

**For me, two cases would solve this consistently**

1. the test-harness always sends a $schema-keyword with the schema-version it references
2. I add only the requested draft-version from dialect-command to the validator (disabling cross-draft support) - which would make adding the schema-version above obsolete.

or **maybe I am missing a point here** _(implicit dialect schemas?)_ - happy to hear your thoughts. 

Cheers,
Sascha

----
References

```ts
// supports draft-4 - draft-2020-12
compileSchema;

// compiles the schema with draft-07
compileSchema({ $schema: "https://json-schema.org/draft-07/schema#" });

// compiles the schema with latest draft (draft-2020-12)
compileSchema({});

// only draft07 is used for all JSON schema
compileSchema({}, { drafts: [draft07] });

// draft04 and draft07 are supported, last draft is default draft
compileSchema({}, { drafts: [draft04, draft07] });
```

- [Docs#compileSchema](https://github.com/sagold/json-schema-library?tab=readme-ov-file#compileschema)

sagold: So my assumption is:

on command "dialect" it is expected to set the draft-version under which the incoming schemas should be validated (implicit)

Julian: (Thanks in particular for what I presume to be you fully writing your own comments, which unfortunately is somewhat rare these days! And for the kind words.)

> the test-harness always sends a $schema-keyword with the schema-version it references

The case that's relevant here (which you referenced in your examples) is that the schema `{}` is a valid draft 7 schema, and we want to be able to answer "what does your implementation do to this schema" -- always adding `$schema` in the harness would prevent being able to answer that question, which is the same reason Bowtie as a harness runner does not do so -- it's valid JSON Schema spec behavior, so we want to know what an implementation does with such a schema, and for better or worse, in earlier drafts it is *not* valid spec-compliant to simply reject that schema and say `$schema` is required even if it's poor schema hygiene (and Bowtie is interested primarily in compliance, not in style or linting).

> I add only the requested draft-version from dialect-command to the validator (disabling cross-draft support) - which would make adding the schema-version above obsolete.

This though works -- it will work for everything other than [tests which require multiple draft support](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/optional/cross-draft.json) but Bowtie itself doesn't yet know how to run those against implementations.

If or when Bowtie does run them, it'll be modifying your harness to include potentially the dialects needed for that specific multi-dialect test.

> on command "dialect" it is expected to set the draft-version under which the incoming schemas should be validated (implicit)

Yes precisely!

So tl;dr It's possible I've missed something, but if I haven't, I think your other idea of just enabling the one dialect passed as `dialect` request seems like the right API to use for your library as it still allows testing schemas without `$schema` keywords!

sagold: Thank you for detailed explanation! 

I now understand the intention and could adjust the library's option to reflect this behaviour according to your description and the existing implementations. The given dialect is now passed in as an option to`compileSchema`, while still allowing a `$schema` id and all drafts being available. That should be helpful behaviour for consumers, too.

I do hope the latest changes resolve this issue. Let me know if something else is still amiss.

> Thanks in particular for what I presume to be you fully writing your own comments, which unfortunately is somewhat rare these days!

I am sorry to hear that! There is lot to say to this... I prefer doing things myself.

Julian:
Thanks! Looks great! Let's see how this goes, appreciate your efforts.

sagold:
@Julian Thank you. And thank you for your time and efforts you put into this project!

PR #2564, opened 2026-04-23
Working JSU implementations

zx80:
- common harness: register meta schemas once instead of on each test case
- c: same as previous setup
- python: same as previous setup
- java: simpler runtime installation
- js: fix dependency installation
- perl: fix dependency installation


zx80:
All five versions seem to work as expected, and the dockerfile build should be future-proof about dependencies.

Next time I'll check that the bowtie validation checks work _before_ releasing a new version.

zx80:
About performance:

1. The updated harness puts meta schemas in cache once instead of on every test case, which should alleviate some of the recurring costs.
2. This is a multi-stage compiler (schema → model → source → binary) , **not** a library: Running it on many small cases takes much more time than interpreting a small schema directly inside the harness as done by most other implementations. A _x6_ does not seem too expensive in this context.
3. Further acceleration may be achieved by launching a validation once per test case instead of for each test, at the price of some refactoring and complexity of the harness. I can try that if you think that the improvement of (1) above is not enough.

zx80:
I did both (1) and (3) outlined above.

For C, a basic round trip for a test case seems to be between 1.5 and 2.0 seconds, which is slow but is probably the worst case, together with Java. Most of the time is spent in the compilation phase, no surprise.

This basic figure was quite unobvious to obtain by the way, I had to hack the bowtie code. How can one get some level of protocol traces when running bowtie? The container cannot communicate with the outside (I tried stderr and network connections), and the JSON interface does not have available fields to provide feedback, eg performance data, from the harness. I have not found obvious options. It would also help it there was a way to run a harness with bowtie directly, without the container layer.

The next possible performance improving step is to internalize both frontend and backend compiler passes in the harness process, as they are all written in Python. That would still let the actual final compilation stage and execution as separate processes. However, I'm relunctant to do that because it departs from the actual use case which is to run the commands, and it could be more fragile: I'd be more at ease if the current performances are deemed good enough. I'll give it a try anyway.

zx80:
The latest version invokes both compilers directly in the harness instead of as subprocesses. It seems to work…

Smoke times are down to 19s (C), 12s (Java), 4-6s (Python, Perl, JavaScript).

zx80:
@Julian I think that you may have a look at this version which improves both stability and performance.

zx80:
@Julian this latest version use a precompiled version for the C runtime, so smoke time for C is down to around 4s only. I think I did everything possible to improve performance for many small tests.

Julian:
How long do they take to run the entire suite?

zx80:
Time to run each suite with `jsu-compile`, for each target language, in rounded seconds:

| Suite   |  C  |  Py |  JS | Perl | Java |
|:--------|----:|----:|----:|-----:|-----:|
| 3       |  15 |  26 |  12 |   12 |   77 |
| 4       |  21 |  39 |  17 |   16 |  118 |
| 6       |  29 |  55 |  22 |   22 |  169 |
| 7       |  32 |  61 |  24 |   24 |  187 |
| 2019    |  43 |  84 |  33 |   33 |  261 |
| 2020    |  44 |  86 |  34 |   34 |  268 |
| ***all*** | _184_ | _351_ | _142_ |  _141_ | _1080_ |

So all the suites running through bowtie take 2-3 minutes for C, JS and Perl, 5 minutes for Python and probably 18 minutes for Java.

Bowtie failed several times on some runs because it reached a GitHub API limit while retrieving repository data: Maybe these data could be cached instead of retrieving them on each run?

As I said in earlier responses, the underlying implementation is a combination of optimizing compilers, not a direct interpreter: it produces very fast code, quite slowly.

Julian:
Is that on your laptop or running here in GitHub actions on runners? Please let me know the latter, as it's what matters. Previously it was taking upwards of an hour.

> As I said in earlier responses, the underlying implementation is a combination of optimizing compilers

I heard you. As I said, even other optimizing compilers like Corvus.JSONSchema here do not take an hour, they take 5 minutes: https://github.com/bowtie-json-schema/bowtie/actions/runs/24932176620/job/73011659303

Having an implementation which takes over 10x more time than any other and which takes an hour when we recalculate every 6 hours isn't feasible the way we have things set up right now, we'd need to rethink how we collect results on an ongoing basis -- which we can do, but I want to be sure I understand what to expect.

zx80:
It's on an (old) server, not my laptop. On my laptop it seems to be about 30-40% faster than these figures… I've _no_ idea of the expected performance on the GitHub CI.

I expect that the effort to put as much of the load as possible out of the case and test loops and avoid launching processes must have impacted the overall performance significantly.

zx80:
> I heard you. As I said, even other optimizing compilers like Corvus.JSONSchema here do not take an hour, they take 5 minutes: https://github.com/bowtie-json-schema/bowtie/actions/runs/24932176620/job/73011659303

I'd guess that it should also be the case for C, Python, JS and Perl. Java seems to be still quite slow, which could warrant further investigations, but I think that the various improvements should have impacted it by at least a factor of 2.

> Having an implementation which takes over 10x more time than any other and which takes an hour when we recalculate every 6 hours isn't feasible the way we have things set up right now, we'd need to rethink how we collect results on an ongoing basis -- which we can do, but I want to be sure I understand what to expect.

Sure. ISTM that reliable information can only be gathered by actually running the implementations on the CI.

Every 6 hours seems a lot. Maybe you could reduce the rate, at least for slow runners, eg once a day or once a week, and reuse cached results in the interval.

zx80:
What more can I do to reinstate the JSU tests?

Julian:
Timing it in CI for a suite run and sharing the results was what I said would be helpful.

zx80:
Ok... ISTM that I do not have access to your CI. Can **you** run the CI once on the PR?!

Julian:
You don't need access to run it in your fork -- it may require you temporarily disabling other implementations or just adding a custom run which just runs your JSU implementations which is the only thing we care about timing.

zx80:
After some painful trial and errors (testing a black box through a black box is so :sob:), the runtime on GitHub CI for the new optimized harness seems to be, in seconds:

| Tests   |    C | Java |  JS | Perl |  Py |
|:--------|-----:|-----:|----:|-----:|----:|
| Smoke   |    4 |   14 |   4 |    4 |   6 |
| Suite 7 |   31 |  238 |  21 |   24 |  71 |

I used my own images for testing because the CI was pulling older images from the previous installation instead of the new ones.

The java implementation is the slowest, 4 minutes for one suite, so it should be under 30 minutes for the 6 suites with Java, which seems acceptable. The other implementations should complete all suites in a few minutes.

@Julian is this ok for you? Or do you need further data?

Julian:
Yes that seems hopefully good enough to try. I'll give this a look hopefully this weekend. Is there a way we can reduce some duplication here, if the files are exactly the same can you just use a single copy and some symlinks? Or is there any actual difference between the harnesses

zx80:
> Is there a way we can reduce some duplication here, if the files are exactly the same can you just use a single copy and some symlinks? Or is there any actual difference between the harnesses

The harnesses are identical.

I tried but did not succeed with links, COPY --link behaviour was not what I was hoping.

From a repository point of view git should figure out that the files are identical.

PR #2565, opened 2026-04-23
Fix Windows test failures

### LINKED ISSUE #2566: Support Windows for the Bowtie build environment
There are a couple of issues that prevent the bowtie build to run succesfully in a Windows environment.

PR #2565 addresses these problems.

mwadams:
Address multiple issues that caused test failures on Windows while preserving existing behavior on Linux:

- test_integration: Build Docker context tar with CRLF-to-LF normalization and executable mode bits on Windows, where git autocrlf introduces CRLF and the filesystem lacks an exec bit. Linux uses tar.add() as before.

- test_integration: Compensate for Rich subtracting 1 from console width on piped stdout (legacy Windows detection) by setting COLUMNS=81 on Windows. Translate Rich ROUNDED-to-SQUARE box-drawing substitution in pretty-output assertions.

- test_integration: Replace pexpect.runu() (Unix-only PTY) with subprocess.run() on Windows for version/help smoke tests.

- test_integration: Decode subprocess output as UTF-8 and normalize CRLF line endings so assertions don't need platform-specific strings.

- _benchmarks: Use delete_on_close=False on NamedTemporaryFile so subprocesses can reopen them on Windows (exclusive file locks).

- _benchmarks: Use --output FILE instead of --pipe on Windows for pyperf, which fails with msvcrt.open_osfhandle() on asyncio pipes. Reserve a unique temp path and delete before passing to pyperf, which refuses to overwrite existing files.

- _suite: Detect Windows drive-letter paths with os.path.splitdrive() before URL.parse(), which misinterprets the drive letter as a scheme.

- test_cli: Set PYTHONUTF8=1 in subprocess env to avoid UnicodeDecodeError from Click/Rich output on Windows.

Fixes #2566


Julian:
Would be great if another Windows user had a look at this to confirm it works well for them, maybe you know another one :D?

bowtie/tests/test_integration.py
```diff
@@ -2330,11 +2390,21 @@ async def test_info_unsuccessful_start(succeed_immediately):
 
 @pytest.mark.asyncio
 async def test_filter_implementations_no_arguments():
-    output, exit_code = pexpect.runu(
-        sys.executable,
-        args=["-m", "bowtie", "filter-implementations"],
-        withexitstatus=True,
-    )
+    if sys.platform == "win32":
```
Julian, line 2393: I suspect in all these subprocess cases we just want to be passing an encoding (which makes Python return a str instead of bytes, and effectively act like universal newlines mode where it will do the line ending conversion for us).

PR #2580, opened 2026-04-26
Migrate to attest action as the old one is a wrapper for it

OptimumCode:
Migrate from https://github.com/actions/attest-build-provenance because it is a wrapper over https://github.com/actions/attest.

Additional permissions added as recommended by the action description.


Julian:
Thanks!

PR #2612, opened 2026-05-08
Bump Blaze SHA

jviotti:
I just realised Bowtie is running an ancient one


Julian:
Now that [dependabot supports vcpkg](https://github.blog/changelog/2025-08-12-dependabot-version-updates-now-support-vcpkg/) I would still encourage you to switch to versions so that dependabot keeps it up to date automatically like other implementations :) but certainly happy to merge.

PR #2617, opened 2026-05-09
Add 5 missing JSU implementations to dependabot configuration

zx80:
Probably this is necessary so that the implementations are run?

Are there other necessary changes to the CI?


Julian:
All this is [already present](https://github.com/bowtie-json-schema/bowtie/blob/aff07f2c712b25e4f46bfc3bd56f39dc9b1ec400/.github/dependabot.yml#L430-L464) (and CI generally should fail if it weren't).

PR #2624, opened 2026-05-13
Add support for annotations in the Java networknt implementation harness

VIDIT45AGARWAL:
As per the discussion on slack, added annotation support for networknt implementation without breaking the existing bowtie commands.
Now when the run command includes ` "collect_annotations": true `, the harness switches from simple validation to annotation collection, while without the flag behavior is identical to the current upstream harness.

**Working**
- A new `collectAnnotations()` method recursively walks the OutputUnit tree returned by networknt and extracts annotations as a nested map: instanceLocation → keyword → schemaLocation → value
- The `TestResult` record now has an optional annotations field, annotated with `@JsonInclude(NON_EMPTY)` so it is completely omitted from the JSON output when not in annotation mode
- The `Test` record's valid field was changed from `boolean` to `JsonNode `to handle annotation test cases that may not have a valid field

**Testing**
``` bash
# Build
docker build -t networknt-test implementations/java-networknt-json-schema-validator/

# Validation mode (no flag — should show only valid field)
printf '{"cmd":"start","version":1}\n{"cmd":"dialect","dialect":"https://json-schema.org/draft/2020-12/schema"}\n{"cmd":"run","seq":1,"case":{"description":"type check","schema":{"type":"string"},"tests":[{"description":"valid","instance":"hello"},{"description":"invalid","instance":42}]}}\n{"cmd":"stop"}\n' | docker run --rm -i networknt-test

# Annotation mode (with flag — should show annotations)
printf '{"cmd":"start","version":1}\n{"cmd":"dialect","dialect":"https://json-schema.org/draft/2020-12/schema"}\n{"cmd":"run","seq":1,"collect_annotations":true,"case":{"description":"title test","schema":{"title":"My Schema","type":"string"},"tests":[{"description":"a string","instance":"hello"}]}}\n{"cmd":"stop"}\n' | docker run --rm -i networknt-test
```

**Output**
``` bash
# Validation mode
{"ok":true}
{"seq":1,"results":[{"valid":true},{"valid":false}]}

# Annotation mode
{"ok":true}
{"seq":1,"results":[{"valid":true,"annotations":{"":{"title":{"#":"My Schema"}}}}]}
```

### DIFF
--- a/implementations/java-networknt-json-schema-validator/BowtieJsonSchemaValidator.java
+++ b/implementations/java-networknt-json-schema-validator/BowtieJsonSchemaValidator.java
@@ -124,14 +124,12 @@ public class BowtieJsonSchemaValidator {
     output.println(objectMapper.writeValueAsString(dialectResponse));
   }
 
-  private void
-  addAnnotation(Map<String, Map<String, Map<String, Object>>> annotations,
-                String instanceLoc, String schemaLoc, String keyword,
-                Object value) {
-    Map<String, Map<String, Object>> byKeyword = annotations.computeIfAbsent(
-        instanceLoc, k -> new LinkedHashMap<>()); // NOPMD
-    Map<String, Object> bySchema =
-        byKeyword.computeIfAbsent(keyword, k -> new LinkedHashMap<>()); // NOPMD
+  @SuppressWarnings("PMD.UseConcurrentHashMap")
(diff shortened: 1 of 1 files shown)


Julian:
We'll come back to this (harnesses) once something like #2642 is merged

PR #2879, opened 2026-06-26
lua-schema: refactor with builder

fperrad:


implementations/lua-schema/Dockerfile
```diff
@@ -1,14 +1,22 @@
...
+WORKDIR /harness
+
+RUN echo "@testing https://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
+RUN apk add --no-cache \
+    lsb-release-minimal \
+    lua5.5 \
+    lua5.5-dkjson@testing \
+    lua5.5-lpeg \
+    lua5.5-rex-pcre2
+COPY --from=builder /usr/local/share/lua/5.4 /usr/local/share/lua/5.5
```
Julian, line 19: Seems strange to build on 5.4 and run on 5.5 here

fperrad: `luarocks-5.5` is not available on Alpine 3.24, so it is not possible to build with Lua 5.5.

as `lua-schema` is a pure Lua module (compatible for Lua 5.1 to 5.5), that works.

note: native extensions (C API) live in `/usr/local/lib/lua/5.x`, Lua files live in `/usr/local/share/lua/5.x`

PR #2880, opened 2026-06-26
fix: Applies fixes for justinrainbow/json-schema for draft 2019-09 once results of 2026.6.1 release where available

DannyvdSluijs:
Seeing the results for draft 2019-09 for justinrainbow/json-schema I suspected something was off. 
This PR resolves the test harness using the correct check mode

### DIFF
--- a/implementations/php-justinrainbow-json-schema/src/TestHarness.php
+++ b/implementations/php-justinrainbow-json-schema/src/TestHarness.php
@@ -159,7 +159,8 @@ class TestHarness
     {
         return match ($this->currentDialect) {
             'http://json-schema.org/draft-06/schema#',
-            'http://json-schema.org/draft-07/schema#' => Constraint::CHECK_MODE_NORMAL | Constraint::CHECK_MODE_STRICT,
+            'http://json-schema.org/draft-07/schema#',
+            'https://json-schema.org/draft/2019-09/schema' => Constraint::CHECK_MODE_NORMAL | Constraint::CHECK_MODE_STRICT,
             default => Constraint::CHECK_MODE_NORMAL,
         };
     }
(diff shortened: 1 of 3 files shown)


implementations/php-justinrainbow-json-schema/src/TestHarness.php
```diff
@@ -159,7 +159,8 @@ private function checkMode(): int
     {
         return match ($this->currentDialect) {
             'http://json-schema.org/draft-06/schema#',
-            'http://json-schema.org/draft-07/schema#' => Constraint::CHECK_MODE_NORMAL | Constraint::CHECK_MODE_STRICT,
+            'http://json-schema.org/draft-07/schema#',
+            'https://json-schema.org/draft/2019-09/schema' => Constraint::CHECK_MODE_NORMAL | Constraint::CHECK_MODE_STRICT,
```
Julian, line 163: I guess I hadn't noticed this earlier but really we ask harnesses to run in their "normal" defaults, at least until we get to #8 and then can run multiple different configurations of an implementation and show each -- does your implementation basically have a non-compliant default mode?

DannyvdSluijs: I understand and in this library it is something else. I've took over maintenance and wanted to move forward. The current setup for draft 3 & 4 is a union of both drafts. 

I didn't want to make it more complex and make new drafts supported in this weird union state. But I could not drop the wide number of users depending on draft 3 & 4. 

Which resulted in this configuration. So draft 3 & 4 are the exception. Hoping you can see it the same and still merge the PR? Ultimately it is up to you if you're okay with it.

Julian: Got it ok it's not new in this PR so I guess I'm happy to merge for now and think about what we can do here.

PR #3028, opened 2026-07-30
feat: update JSON schemas for annotation reporting format

VIDIT45AGARWAL:
As discussed, the PR contains all the agreed schema changes.

### DIFF
--- a/bowtie/schemas/models/test.json
+++ b/bowtie/schemas/models/test.json
@@ -17,11 +17,21 @@
     },
     "instance": {
       "description": "The instance which should be validated against the schema in \"schema\"."
-    },
-    "valid": {
-      "description": "Optionally, whether validation of this instance is expected to be valid or not. If unprovided, implementation results will be reported without comparing against an expected value.",
-      "type": "boolean"
     }
   },
-  "additionalProperties": false
+  "oneOf": [
+    {
+      "$ref": "tag:bowtie.report,2026:output"
+    },
+    {
+      "description": "No assertion — implementation results will be reported without comparing against an expected value.",
+      "not": {
+        "anyOf": [
+          { "required": ["valid"] },
+          { "required": ["annotations"] }
+        ]
+      }
+    }
+  ],
+  "unevaluatedProperties": false
 }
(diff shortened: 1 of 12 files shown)


AgniveshChaubey (approved):
Things look good to me. Let's just check why 2 CI tests are failing. Then we should be good to go.

bowtie/schemas/models/test.json
```diff
@@ -17,11 +17,41 @@
...
     "instance": {
       "description": "The instance which should be validated against the schema in \"schema\"."
-    },
-    "valid": {
-      "description": "Optionally, whether validation of this instance is expected to be valid or not. If unprovided, implementation results will be reported without comparing against an expected value.",
-      "type": "boolean"
     }
   },
-  "additionalProperties": false
+  "oneOf": [
```
AgniveshChaubey, line 22: Isn't the first two schemas here collectively same as `output` schema? Also, I'm not fully sure about 3rd schema here, but if we're going to need it, then I think we can use something like the following, right?:

```json
"oneOf": [
  {
    "$ref": "tag:bowtie.report,2026:output"
  },
  {
    "description": "No assertion...",
    ...
  }
]
```

VIDIT45AGARWAL:
> Let's just check why 2 CI tests are failing.

@AgniveshChaubey they are failing because hypothesis (our testing framework) is generating random mock test cases that include the new "annotations" assertions based on these new schemas. However, because I isolated only the schema changes in this PR, the Python codebase here doesn't have the implementation logic to parse those annotations yet (that logic is sitting in my other branch waiting for this to merge). So the old Python code crashes when it hits the new annotation test cases

AgniveshChaubey:
@Julian, we should be good to go.

Julian:
We can't merge PRs with failing CI! If we need to modify either the hypothesis strategies (temporariliy?) here or another part of the code to get CI to pass we should do that alongside -- or else we can simply go back to merging the schema changes in the other PR along with everything else if that's easier to do!

VIDIT45AGARWAL:
I think it is better to just merge these schema changes along with my older PR. That branch already has the implementation code to handle the schemas, so all the CI tests pass there perfectly.

PR #3067, opened 2026-08-26
Know about the JSON Schema v1 dialect

mertcanaltin:
Refs #40. Backend only, as you suggested, so nothing about the website changes.

`bowtie suite {PATH}/tests/v1` runs against any harness which declares
`https://json-schema.org/v1` in the dialects it returns from `start`.

My issue was wrong about one of the three pieces. The suite loader needs no
change: `_cases_and_dialect` looks a dialect up by the name of the directory
the tests are in, and the suite already lays v1 out as `tests/v1` with its
remotes under `remotes/v1`, so a `shortName` of `v1` is enough. `_remotes_in`
also already picks up the 17 remotes under `v1/` and the 6 shared ones, and
skips every `draft*` one. So this is the `dialects.json` entry, plus what that
entry then breaks.

### The publication date

v1 has not been published, so it has no first publication date, and that field
is now optional and means exactly that. It matters because `Dialect.latest()`
is `max()` over publication date, and `latest` decides the default dialect for
a run, so leaving it alone means every harness is suddenly sent v1 schemas it
never said it supports. An unpublished dialect sorts newest but is kept out of
four places:

* `Dialect.latest()`
* `bowtie filter-dialects --latest`, which is what scripts read
* the website, which has no report to link to for it yet
* the benchmark workflow's dialect matrix. This one caught me while reviewing.
  `filter-benchmarks -D v1` correctly returns nothing, so no benchmark job is
  created, but `merge_benchmarks_into_single_report` matrices over every
  dialect in `dialects.json` and would then call `iterdir()` on a directory the
  download step never created. That fails the merge, which fails the upload job
  that needs it. One `jq` `select` fixes it.

It is still listed by `filter-dialects`, still selectable by name, and still
reported as supported by a harness which declares it, since answering "which
dialects does this implementation support?" with a wrong list would be worse
than the surprise. Filling the date in later makes it appear everywhere with
no further code change.

That is the "gated behind anything" question from the issue, answered the least
surprising way I could find. Happy to reshape it if you would rather it were an
explicit flag than an absent date.

### The specification

`referencing` does not know the v1 identifier, so `Dialect.specification()`
raises `UnknownDialect`, and it is called for every test case in
`TestCase.from_dict`, so nothing loads at all. It now falls back to the
specification whose resource structure v1 shares, meaning the rules for finding
`$id`s, anchors and subschemas. Every structural keyword in `tests/v1` is a
2020-12 keyword, so 2020-12 is that specification for now. The entry should be
deleted once referencing knows v1 itself, and the comment says so.

### Verification

`tests/v1` end to end against a locally built `js-ata` image with the URI added
to its dialects array: 1133 tests ran, 1 failure, 0 errors, 0 skips.

That failure is `remote ref, containing refs itself`, and it is not mine. It
fails identically on draft2020-12, with the same empty registry, because
`cases_from` supplies remotes only for `refRemote`, `dynamicRef` and
`vocabulary` while `ref.json` and `anchor.json` reference `localhost:1234` too.
Left alone here. Say the word and I will open an issue.

The full test suite passes with nothing deselected, containers included, 341
tests. Also ran pyright, ruff, prettier, `tsc --noEmit`, eslint, and the
pre-commit hooks which touch these files.

### Calls I made that you may want to reverse

* The integration fixtures which spell out every dialect gain v1, since the
  miniatures declare whatever Bowtie knows. The alternative was having the
  miniatures declare only published dialects, which felt like bending the fakes
  to avoid updating expectations.
* The date stays a mandatory argument even though it may now be `None`, so that
  forgetting one is a `TypeError` rather than a quiet claim that a dialect is
  unpublished.
* Ordering falls back to the short name, so a second unpublished dialect would
  not make two dialects incomparable, or make `sorted(Dialect.known())` differ
  between runs.
* Only `draft-next` as an alias. `1` reads as draft 1 next to the existing
  `3`/`4`/`6`/`7`, and `next` stops being true the moment a v2 is drafted.
* `prettyName` is just `v1`, which reads a little bare next to "Draft 2020-12".
* One test validates `data/dialects.json` against `models/dialect.json`, since
  nothing did and I had just changed both. Drop it if it is not wanted.

### Not covered

* No implementation in this repo declares v1, so it will read as unsupported
  everywhere until harnesses opt in. That goes to `js-ata` separately.
* `hypothesis.dialects()` still always generates a publication date, so the
  unpublished path is covered by the explicit tests rather than by property
  tests. Teaching the strategy about it also means guaranteeing unique short
  names, which seemed like more than this change should carry.
* Nothing about `tests/v1/proposals/`. `_glob` does not recurse, so it is
  excluded the same way `optional/` is.

### DIFF
--- a/bowtie/schemas/models/dialect.json
+++ b/bowtie/schemas/models/dialect.json
@@ -5,7 +5,7 @@
   "$id": "tag:bowtie.report,2024:models:dialect",
 
   "type": "object",
-  "required": ["firstPublicationDate", "prettyName", "shortName", "uri"],
+  "required": ["prettyName", "shortName", "uri"],
   "properties": {
     "uri": {
       "description": "The URI representing this dialect, also known as its dialect identifier",
@@ -32,7 +32,7 @@
       "pattern": "[a-z]+[a-z0-9-]*"
     },
     "firstPublicationDate": {
-      "description": "The date when the dialect was first published publicly as a document",
+      "description": "The date when the dialect was first published publicly as a document. Absent for a dialect which has not been published yet, which is also how Bowtie knows not to treat it as the latest dialect.",
       "readOnly": true,
 
       "type": "string",
--- a/bowtie/schemas/models/dialect.json
+++ b/bowtie/schemas/models/dialect.json
@@ -6,6 +6,19 @@
 
   "type": "object",
   "required": ["prettyName", "shortName", "uri"],
+
+  "oneOf": [
+    {
+      "description": "A dialect which has been published, and says when it was",
+      "required": ["firstPublicationDate"]
+    },
+    {
+      "description": "A dialect still being written, which has no publication date to give",
+      "required": ["prerelease"],
+      "properties": { "prerelease": { "const": true } }
+    }
+  ],
+
   "properties": {
     "uri": {
       "description": "The URI representing this dialect, also known as its dialect identifier",
@@ -32,11 +45,18 @@
       "pattern": "[a-z]+[a-z0-9-]*"
     },
     "firstPublicationDate": {
-      "description": "The date when the dialect was first published publicly as a document. Absent for a dialect which has not been published yet, which is also how Bowtie knows not to treat it as the latest dialect.",
+      "description": "The date when the dialect was first published publicly as a document",
       "readOnly": true,
 
       "type": "string",
       "format": "date"
+    },
+    "prerelease": {
+      "description": "Is this dialect still being written? Bowtie knows about a prerelease dialect so its tests can be run against harnesses which opt into it, but does not treat it as the latest dialect, benchmark it, or show it on the report.",
+      "readOnly": true,
+
+      "type": "boolean",
+      "default": false
     },
     "hasBooleanSchemas": {
       "description": "Does the dialect have a notion of boolean schemas?",
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -28,10 +28,9 @@ jobs:
       - name: Collect supported dialects
         id: dialects-matrix
         run: |
-          # Published dialects only. An unpublished one has no benchmarks and
-          # no harnesses declaring it, so its merge step would find no
-          # artifacts to combine.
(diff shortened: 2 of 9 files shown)


bowtie/schemas/models/dialect.json
```diff
@@ -5,7 +5,7 @@
   "$id": "tag:bowtie.report,2024:models:dialect",
 
   "type": "object",
-  "required": ["firstPublicationDate", "prettyName", "shortName", "uri"],
```
Julian, line 8: Minor, but I'd certainly feel slightly more comfortable having this be `required` still in a oneOf, like `oneOf: [required: [firstPublicationDate], properties: {prerelease: {const: true}}` to explicitly make this dialect special Without sacrificing any validation. I think that makes the queries in some other places slightly clearer too because you would filter out pre-releases.

mertcanaltin: Done, and it is better. The entry says `prerelease: true`, the schema carries the `oneOf` you sketched, and `Dialect` checks the same rule as a validator so the model and the schema cannot drift.

The `oneOf` buys something I had not thought about: it rejects an entry giving both a publication date and `prerelease: true`, and one giving neither. Both were constructible with the field simply made optional.

The queries did get clearer. The benchmark matrix is now `select(.prerelease != true)`, and the website skips a dialect because it says it is a prerelease rather than because a field is missing.

mertcanaltin:
Thanks for merging #3098 so quickly.

One thing I ran into while wiring v1 into js-ata: harness-ci.yml installs bowtie from PyPI, and 2026.7.4 is from before this PR landed. So a harness that puts https://json-schema.org/v1 in its dialects list dies at startup with `KeyError: <URL https://json-schema.org/v1>`, whatever the workflow pins point at. Nothing on the harness side can work around that.

I have the declaration ready as a draft in bowtie-json-schema/js-ata#30. With bowtie built from main it runs clean on all three dialects, including the v1 remote ref case that #3098 fixed.

Would you cut a release when you get a chance? No rush, just wanted to flag that this is the only thing left in the way.

PR #3088, opened 2026-09-06
Support unified JSON Schema Test Suite validation format

codefromlani:
Update Bowtie to support the unified JSON Schema Test Suite validation format.

- Migrate validation suite loading to `validation/tests/*.json`
- Add support for the unified `compatibility` field
- Add an explicit `--dialect` option to `bowtie suite`
- Update fixtures and tests for the new format
- Remove legacy validation-suite loading code

### DIFF
--- a/bowtie/_cli.py
+++ b/bowtie/_cli.py
@@ -2644,6 +2644,13 @@ def _run_suite(
 @IMPLEMENTATION
 @FILTER
 @fail_fast
+@dialect_option(
+    is_eager=True,
+    help=(
+        "A URI or shortname identifying the dialect to evaluate under. "
+        f"Possible shortnames include: {POSSIBLE_DIALECT_SHORTNAMES}"
+    ),
+)
 @SET_SCHEMA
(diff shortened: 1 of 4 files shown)


jdesrosiers (commented):
I don't know this codebase well, but it surprised me to see the addition of the `--dialect` flag. I'm pretty sure I remember that the CLI already has something like that.

I also noticed that I think the `_is_compatible` function will need to be improved. I don't think it will handle some of the more complex ranges we needed to cover all the different cases. The psudocode in the readme should help to get the right behavior.

https://github.com/bowtie-json-schema/bowtie/blob/672d2109fad58a69b8a7e18b0cdb3cea36d47b26/bowtie/_suite.py#L221-L244

codefromlani:
> I don't know this codebase well, but it surprised me to see the addition of the `--dialect` flag. I'm pretty sure I remember that the CLI already has something like that.
> 

Yeah, I added `--dialect` because unified-format input has no directory to infer dialect from

